Issue #61, opened 2020-04-02
Server is locking and lagging, just restarting to resolve.


ghost:
**Describe the bug**
The server starts to press after a certain time, causing some rollbacks until it is impossible to play.

**Version Encountered**
0.6.2.401

**To Reproduce**
I don't know how to reproduce, just leaving the server on for 24 hours

**Expected behavior**
It seems when the server goes online for a long time

**Logs**
https://ascensiongamedev.com/resources/filehost/7f9ff432827c2569ee8930f7ef74616a.log

**Operating System:**
Windows Server 2016

**Additional context**

jcsnider:
Lots of bugs that could have caused this have been fixed in the past week. Given that the logs here no longer apply and there are no actual repro steps this is being closed.

PR #74, opened 2020-04-03
Fixes collection was modified error when looping through map traps

jcsnider:


jcsnider:
Fixes a collection was modified error seen in the logs of #61

PR #106, opened 2020-04-07
Diagonal movement feature

RedBandana:
NPCs and players can move in 8 directions with those commits.
To attack in diagonal, the player last move must be in diagonal. The same rule apply if the player want to trigger an action event diagonally.

The projectiles cannot be thrown diagonally.

Right now, there is no known bugs.


RedBandana:
Players and NPCs can move and attack in diagonal with those commits.

jcsnider:
Sorry, but this isn't a feature that we'd like in the base engine. It adds complexity in the code, would break existing games big time.

PR #107, opened 2020-04-07
 Unselect a target without using the mouse

RedBandana:
Sometimes you want to unselect a target without using the mouse. With this patch, you can click on Esc to do so.


jcsnider:
While we're open to discuss better ways of handling targeting and targeting based controls (in a new enhancement issue) this is not the way to go about it. We would need separate controls defined in the options window at least instead of piggybacking off of the menu key.

PR #134, opened 2020-04-13
Fixes the logout packet so it doesn't error if received and the clien…

jcsnider:
…t isn't logged in.

### DIFF
--- a/Intersect.Server/Networking/PacketSender.cs
+++ b/Intersect.Server/Networking/PacketSender.cs
@@ -1003,58 +1003,61 @@ namespace Intersect.Server.Networking
         public static void SendPlayerCharacters(Client client)
         {
             var characters = new List<CharacterPacket>();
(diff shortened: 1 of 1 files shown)


Intersect.Server/Networking/PacketSender.cs
```diff
@@ -1003,58 +1003,61 @@ public static void SendCreateCharacter(Client client)
         public static void SendPlayerCharacters(Client client)
         {
             var characters = new List<CharacterPacket>();
-            foreach (var character in client.Characters.OrderByDescending(p => p.LastOnline))
+            if (client.User != null && client.Characters.Count > 0)
```
lodicolo, line 1006: For readability sake, use
```csharp
if (client.User == null || client.Characters.Count < 1)
```
instead of indenting everything it wrap it in an if statement with no else.

jcsnider: I think the commit below is what you're looking for but let me know.

PR #138, opened 2020-04-13
Fixing username lookups not being case insensitive.

### LINKED ISSUE #129: Some players are unable to log in.
**Describe the bug**
After I returned from MYSQL to SQLITE, some players are unable to login, as it gives this error. failed to load the account, please try again later.
When I return to the MYSQL system, that player is able to open his account, but on SQLITE the bug comes back.

**Version Encountered**
0.6.2.0 (LAST DEV PULL)

**To Reproduce**
Migrating from MYSQL to SQLITE, some accounts do not work.

**Expected behavior**
That I didn't have this error when using SQLITE

**Operating System:**
Windows 10

jcsnider:
Resolves #129

### DIFF
--- a/Intersect.Server/Database/DbInterface.cs
+++ b/Intersect.Server/Database/DbInterface.cs
@@ -507,7 +507,7 @@ namespace Intersect.Server.Database
             lock (mPlayerDbLock)
             {
                 // ReSharper disable once SpecifyStringComparison
-                var user = sPlayerDb.Users.FirstOrDefault(p => p.Name.ToLower() == username.Trim().ToLower());
+                var user = User.Find(username);
 
                 if (user != null)
                 {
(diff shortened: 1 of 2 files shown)


Intersect.Server/Database/DbInterface.cs
```diff
@@ -505,10 +506,8 @@ public static UserRights CheckAccess([NotNull] string username)
         {
             lock (mPlayerDbLock)
             {
-                var user = sPlayerDb.Users.Where(
-                        p => string.Equals(p.Name.Trim(), username.Trim(), StringComparison.CurrentCultureIgnoreCase)
-                    )
-                    .FirstOrDefault();
+                // ReSharper disable once SpecifyStringComparison
+                var user = sPlayerDb.Users.FirstOrDefault(p => p.Name.ToLower() == username.Trim().ToLower());
```
lodicolo, line 510: Use `User.Find(username)`.

jcsnider: Good call, updated.

PR #162, opened 2020-04-21
Actually makes party xp range and common event options work

### LINKED ISSUE #158: XP Range for Party not working
**Describe the bug**
I did some tests with players and the distance is not affecting anything.

**Version Encountered**
0.6.2.0 (LAST DEV)

**To Reproduce**
Take a group, and walk as far as you want, kill a monster, XP will share for those in the city

**Operating System:**
Windows Server 2012

jcsnider:
Resolves #158


jcsnider:
Resolves #158

PR #172, opened 2020-04-22
Resolves #171, npcs vanishing when traversing maps

### LINKED ISSUE #171: NPCs disappear when crossing map-boundries
**Describe the bug**
When an aggressive monster (not sure if it would happen with passive) sees you and heads towards you to attack, once it crosses map boundary it disappears. When it finally gets to you it is invisible and unable to be targeted to attack. Logging out and back in reveals monster.

**Version Encountered**
Current Dev Branch w/ all current bug fixes.

**To Reproduce**
Best way to produce is to likely make a map that spans boundaries, and an npc that is set to agro range. Stand outside bounds so it has to travel around the map, through another map boundary and to your character.

**Expected behavior**
No invisible npcs from map travel.

**Screenshots**
https://ascensiongamedev.com/resources/filehost/49fcf858620025bcdcc583587f49bd51.gif
There is a gif (cut out the middle part waiting for monster to arrive.) Showing entire described above.

**Logs**
No log errors.

**Operating System:**
Windows

**Additional context**
None

jcsnider:


rhathaway86:
Added in these fixes and everything works great, thanks for the quick reply on this one glad it wasn't too complicated and you knew where it was!

PR #180, opened 2020-04-26
Added loot ownership

Azurebeats:
* Map items can be spawned with an owner, and only be picked up by the owner and their party. Anyone can still see it, however.
* The time that loot is exclusive to the owner and their party is configurable through the server's config.json under Map Options.
* NPC Drops are owned by the player (and their party) that dealt the most damage to them. (Unless killed by an NPC or something else, then there is no owner)
* Player drops on death are owned by whoever dealt the killing blow and their party.
* Items dropped from your inventory to the map are owned by you and your party until the timer expires.

* Made the packet handler for PickupItempacket slightly less a hassle to read by shortening some checks.

### DIFF
--- a/Intersect (Core)/Config/MapOptions.cs
+++ b/Intersect (Core)/Config/MapOptions.cs
@@ -14,6 +14,8 @@ namespace Intersect.Config
 
         public int ItemDespawnTime = 15000;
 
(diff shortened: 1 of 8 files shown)


Azurebeats:
Seeing some unessecary stuff and a small line I forgot.

PR #181, opened 2020-04-26
Assigning loot ownership + small fixes

### LINKED ISSUE #182: Loot Ownership
Loot dropped by enemies or players can currently be picked up by anyone at any time. The killer (or their party) might not get the loot with another player around.

Loot should be ''owned'' by a player for a short period of time before someone else can pick it up.
Preferably with configuration options for the duration loot should be exclusive, and the ability to hide loot from players that can not pick it up until it's available to them.

**Additional context**
See original request: https://www.ascensiongamedev.com/topic/4861-loot-timer/

### LINKED ISSUE #184: Inactive maps don't fully regenerate NPC health and dismiss their targets after 30 seconds
**Describe the bug**
Once a map is considered inactive for 30 seconds, it should regenerate all Resources and NPCs to full vital stats and reset their targets.

This happens for Resources, but not NPCs.

**Version Encountered**
0.6.1.255

**To Reproduce**
Leave a map inactive with an NPC, the code that fully regenerates their vitals is never triggered.

**Expected behavior**
Vitals should regenerate after the set 30 seconds and targets should be wiped.

Azurebeats:
Resolves #182 
Resolves #184 

Adds the ability to tag MapItems with an owner, and only allow the owner (and their party) to pick the items up for a configurable amount of time.

NPC drops will attempt to use the player that did the most damage to it as the owner, unless killed by something other than a player.
Player drops will attempt to use player that did the killing blow as the owner for the items. Unless killed by something else.

Items dropped from the inventory voluntarily are also subject to ownership.

[Preview](https://s3.us-east-2.amazonaws.com/ascensiongamedev/filehost/6e3028d392bae32b70620ae58e7901fc.gif)

Some small fixes are mixed in to this as well as I run into them.

### DIFF
--- a/Intersect (Core)/Config/LootOptions.cs
+++ b/Intersect (Core)/Config/LootOptions.cs
@@ -1,0 +1,12 @@
+﻿namespace Intersect.Config
+{
+    public class LootOptions
+    {
+
+        // Defines how long (in ms) an item drop will be ''owned'' by a player and their party.
+        public int ItemOwnershipTime = 5000;
+
+        // Defines whether players can see items they do not ''own'' on the map.
+        public bool ShowUnownedItems = false;
+    }
+}
--- a/Intersect (Core)/Config/LootOptions.cs
+++ b/Intersect (Core)/Config/LootOptions.cs
@@ -7,6 +7,11 @@
     {
 
         /// <summary>
+        /// Defines how long (in ms) loot will be available for picking up on the map.
+        /// </summary>
+        public int ItemDespawnTime = 15000;
+
+        /// <summary>
         /// Defines how long (in ms) an item drop will be ''owned'' by a player and their party.
         /// </summary>
         public int ItemOwnershipTime = 5000;
--- a/Intersect (Core)/Config/LootOptions.cs
+++ b/Intersect (Core)/Config/LootOptions.cs
@@ -1,12 +1,19 @@
 ﻿namespace Intersect.Config
 {
+    /// <summary>
+    /// Contains configurable options pertaining to the way loot (item) drops are handled by the engine.
+    /// </summary>
     public class LootOptions
     {
 
-        // Defines how long (in ms) an item drop will be ''owned'' by a player and their party.
+        /// <summary>
+        /// Defines how long (in ms) an item drop will be ''owned'' by a player and their party.
+        /// </summary>
         public int ItemOwnershipTime = 5000;
 
-        // Defines whether players can see items they do not ''own'' on the map.
+        /// <summary>
+        /// Defines whether players can see items they do not ''own'' on the map.
+        /// </summary>
         public bool ShowUnownedItems = false;
     }
 }
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -137,9 +137,7 @@ namespace Intersect
 
         public static int MinChatInterval => Instance.ChatOpts.MinIntervalBetweenChats;
 
-        public static bool ShowUnownedItems => Instance.LootOpts.ShowUnownedItems;
-
-        public static int ItemOwnershipTime => Instance.LootOpts.ItemOwnershipTime;
+        public static LootOptions Loot => Instance.LootOpts;
 
         public static bool UPnP => Instance._upnp;
 
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -119,6 +119,8 @@ namespace Intersect
 
         public static int ItemDespawnTime => Instance.MapOpts.ItemDespawnTime;
 
+        public static int ItemOwnershipTime => Instance.MapOpts.ItemOwnershipTime;
+
         public static bool ZDimensionVisible => Instance.MapOpts.ZDimensionVisible;
 
         public static int MapWidth => Instance?.MapOpts?.Width ?? 32;
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -119,6 +119,8 @@ namespace Intersect
 
         public static int ItemDespawnTime => Instance.MapOpts.ItemDespawnTime;
 
+        public static bool ShowUnownedItems => Instance.MapOpts.ShowUnownedItems;
+
         public static int ItemOwnershipTime => Instance.MapOpts.ItemOwnershipTime;
 
         public static bool ZDimensionVisible => Instance.MapOpts.ZDimensionVisible;
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -117,10 +117,6 @@ namespace Intersect
 
         public static int GameBorderStyle => Instance.MapOpts.GameBorderStyle;
 
-        public static int ItemRepawnTime => Instance.MapOpts.ItemSpawnTime;
-
-        public static int ItemDespawnTime => Instance.MapOpts.ItemDespawnTime;
-
         public static bool ZDimensionVisible => Instance.MapOpts.ZDimensionVisible;
 
         public static int MapWidth => Instance?.MapOpts?.Width ?? 32;
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -54,6 +54,8 @@ namespace Intersect
 
         [JsonProperty("Security")] public SecurityOptions SecurityOpts = new SecurityOptions();
 
+        [JsonProperty("Loot")] public LootOptions LootOpts = new LootOptions();
+
         public SmtpSettings SmtpSettings = new SmtpSettings();
 
         [NotNull]
@@ -119,10 +121,6 @@ namespace Intersect
 
         public static int ItemDespawnTime => Instance.MapOpts.ItemDespawnTime;
 
-        public static bool ShowUnownedItems => Instance.MapOpts.ShowUnownedItems;
-
-        public static int ItemOwnershipTime => Instance.MapOpts.ItemOwnershipTime;
-
         public static bool ZDimensionVisible => Instance.MapOpts.ZDimensionVisible;
 
         public static int MapWidth => Instance?.MapOpts?.Width ?? 32;
@@ -138,6 +136,10 @@ namespace Intersect
         public static int MaxChatLength => Instance.ChatOpts.MaxChatLength;
 
         public static int MinChatInterval => Instance.ChatOpts.MinIntervalBetweenChats;
+
+        public static bool ShowUnownedItems => Instance.LootOpts.ShowUnownedItems;
+
+        public static int ItemOwnershipTime => Instance.LootOpts.ItemOwnershipTime;
 
         public static bool UPnP => Instance._upnp;
 
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -846,7 +846,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me && !Globals.Me.IsInMyParty(this))
+                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
                     {
                         return;
                     }
@@ -1217,7 +1217,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me && !Globals.Me.IsInMyParty(this))
+                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
                     {
                         return;
                     }
@@ -1309,7 +1309,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me && !Globals.Me.IsInMyParty(this))
+                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
                     {
                         return;
                     }
@@ -1430,7 +1430,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me && !Globals.Me.IsInMyParty(this))
+                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
                     {
                         return;
                     }
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -846,7 +846,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
+                    if (this != Globals.Me && !(this is Player player && Globals.Me.IsInMyParty(player)))
                     {
                         return;
                     }
@@ -1217,7 +1217,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
+                    if (this != Globals.Me && !(this is Player player && Globals.Me.IsInMyParty(player)))
                     {
                         return;
                     }
@@ -1309,7 +1309,7 @@ namespace Intersect.Client.Entities
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
+                    if (this != Globals.Me && !(this is Player player && Globals.Me.IsInMyParty(player)))
                     {
                         return;
                     }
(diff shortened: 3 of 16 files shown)


Azurebeats:
Just did some cleaning up on a method I changed, and some repeated code I saw.

Azurebeats:
Did some more work to this, adding the ability to hide loot (configurable, but by default on) from other players if there is currently an ownership claim on it and you're not in the owner's party.

Also made it so a player dying from anything but another player will mark them as the owner of their own drops.

Moved the config options to their own loot section.
Would prefer to move these as well:

ItemDespawnTime
ItemSpawnTime (or rename this to MapItemRespawnTime under Map options for clarity, as it has nothing to do with loot drops.. Just respawning item attributes.)

But this would definitely break existing configs? Opinions?

ThousandX:
This is just perfect, the player has control over who can get their loots, and protects them.

Intersect (Core)/Config/LootOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Intersect.Config
+{
+    public class LootOptions
+    {
+
+        // Defines how long (in ms) an item drop will be ''owned'' by a player and their party.
```
lodicolo, line 6: Proper XML doc for these please.

Also add a brief description to the class itself as a guideline of what sort of variables should or shouldn't be added to this scope in the future.

```
/// <summary>
/// Description here
/// </summary>
```

Intersect (Core)/Config/Options.cs
```diff
@@ -135,6 +137,10 @@ public class Options
 
         public static int MinChatInterval => Instance.ChatOpts.MinIntervalBetweenChats;
 
+        public static bool ShowUnownedItems => Instance.LootOpts.ShowUnownedItems;
```
lodicolo, line 140: Honestly I think it should be `public static LootOpts Loot => Instance.LootOpts` instead of adding more variables at the top level here.

Intersect.Client/Entities/Player.cs
```diff
@@ -130,6 +130,19 @@ public bool IsInMyParty(Entity entity)
             return false;
         }
 
+        public bool IsInMyParty(Guid id)
```
lodicolo, line 133: I know you're a LINQ lover, and this is a case where I would honestly do this instead:

```csharp
public bool IsInMyParty(Guid id) => Party.Any(member => member.Id == id);
```

Intersect.Client/Entities/Player.cs
```diff
@@ -130,6 +130,19 @@ public bool IsInMyParty(Entity entity)
...
+        {
+            foreach (var member in Party)
+            {
+                if (member.Id == Id)
```
lodicolo, line 137: You have `Id` (capital "i", refers to `this.Id` instead of the parameter `id`).

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -651,6 +651,13 @@ public void DrawItemsAndLights()
             //Draw Map Items
             foreach (var item in MapItems)
             {
+                // Are we allowed to see and pick this item up?
+                if (!item.Value.VisibleToAll && !item.Value.Owner.Equals(Globals.Me.Id) && !Globals.Me.IsInMyParty(item.Value.Owner))
```
lodicolo, line 655: Use `==` instead of `.Equals` unless you can't, or it's a string (strings should be compared with `string.Equals(a, b, [comparison type])` except in special compiled query cases).

Since Id/Owner are Guids and are non-nullable `.Equals` can't ever throw an NRE here, but there's no reason to not maintain consistency with the rest of the code base (the built-in `==` can't throw an NRE).

Intersect.Server/Entities/Entity.cs
```diff
@@ -2438,9 +2419,29 @@ public virtual void Die(int dropitems = 0, Entity killer = null)
                         continue;
                     }
 
+                    // Decide if we want to have a loot ownership timer or not.
+                    Guid lootOwner = Guid.Empty;
+                    if (this is Npc)
```
lodicolo, line 2424: `this is Npc npc` and replace the below cast with just `npc` (or if `npc` is already taken use `thisNpc` so it's clear that it's still a reference to `this`).

Intersect.Server/Entities/Entity.cs
```diff
@@ -2438,9 +2419,29 @@ public virtual void Die(int dropitems = 0, Entity killer = null)
...
+                        if (taggedBy != null && taggedBy is Player)
+                        {
+                            // Spawn with ownership!
+                            lootOwner = taggedBy.Id;
+                        }
+                    } 
+                    else
+                    {
+                        // There's no tracking of who damaged what player as of now, so going by last hit.. Or set ownership to the player themselves.
+                        lootOwner = playerKiller != null ? playerKiller.Id : Id;
```
lodicolo, line 2437: `lootOwner = playerKiller?.Id ?? Id;`

Intersect.Server/Entities/Npc.cs
```diff
@@ -32,6 +32,24 @@ public class Npc : Entity
         //Damage Map - Keep track of who is doing the most damage to this npc and focus accordingly
         public ConcurrentDictionary<Entity, long> DamageMap = new ConcurrentDictionary<Entity, long>();
 
+        // Returns the entity that ranks the highest on this NPC's damage map.
```
lodicolo, line 35: Proper XML doc here too please.

Intersect.Server/Entities/Npc.cs
```diff
@@ -32,6 +32,24 @@ public class Npc : Entity
...
+        public Entity DamageMapHighest { 
+            get {
+                var damageMap = DamageMap;
```
lodicolo, line 38: This is just a direct reference to `DamageMap` but it's a concurrent collection anyway, just use it directly in the for-each.

Intersect.Server/Entities/Npc.cs
```diff
@@ -32,6 +32,24 @@ public class Npc : Entity
...
+                long damage = 0;
+                Entity top = null;
+                foreach (var pair in damageMap)
```
lodicolo, line 41: I think this can safely be made into a LINQ selector but I don't remember what the extension function is called.

Intersect.Server/Localization/Strings.cs
```diff
@@ -776,6 +776,9 @@ public sealed class ItemsNamespace : LocaleNamespace
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public readonly LocalizedString stunned = @"You cannot use this item whilst stunned.";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString notyours = @"This item does not belong to you!";
```
lodicolo, line 780: So the reason all these strings were in lowercase is irrelevant now (according to JC), and they're not really legible. I haven't gone through and fixed everything because I'm lazy, but can you change this new one to `NotYours` and also add the `NotNull` attribute to it?

(`[NotNull, JsonProperty...]` is valid syntax and preferable for this situation instead of `[NotNull] [JsonProperty...]`).

Intersect.Server/Maps/MapInstance.cs
```diff
@@ -771,12 +780,25 @@ public void Update(long timeMs)
                 //Process Items
                 lock (MapItems)
                 {
+
                     for (var i = 0; i < MapItems.Count; i++)
                     {
-                        if (MapItems[i] != null && MapItems[i].DespawnTime != -1 && MapItems[i].DespawnTime < timeMs)
+                        if (MapItems[i] != null)
```
lodicolo, line 786: `var mapItem = MapItems[i]` and then reuse it in place of repeated `MapItems[i]`.

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1156,14 +1156,36 @@ public void HandlePacket(Client client, Player player, PickupItemPacket packet)
...
+                    var canTake = false;
+                    // Can we actually take this item?
+                    if (mapItem.Owner == Guid.Empty || Globals.Timing.TimeMs > mapItem.OwnershipTime)
                     {
-                        //Remove Item From Map
-                        MapInstance.Get(player.MapId).RemoveItem(packet.MapItemIndex);
+                        // The ownership time has run out, or there's no owner!
+                        canTake = true;
                     }
+                    else if (mapItem.Owner.Equals(player.Id) || player.Party.Any(p => p.Id.Equals(mapItem.Owner)))
```
lodicolo, line 1170: `==` for `Guid`s here too please.

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1156,14 +1156,36 @@ public void HandlePacket(Client client, Player player, PickupItemPacket packet)
...
+                    {
+                        // The current player is the owner, or one of their party members is.
+                        canTake = true;
+                    } 
+
+                    if (canTake)
```
lodicolo, line 1176: Same effect, less nesting:

```csharp
if (!canTake || !player.TryGiveItem(mapItem)
{
    PacketSender.SendChatMsg(player, Strings.Items.notyours, Color.Red);
}
else
{
    //Remove Item From Map
    MapInstance.Get(player.MapId).RemoveItem(packet.MapItemIndex);
}
```

lodicolo:
> Moved the config options to their own loot section.
> Would prefer to move these as well:
> 
> ItemDespawnTime
> ItemSpawnTime (or rename this to MapItemRespawnTime under Map options for clarity, as it has nothing to do with loot drops.. Just respawning item attributes.)
> 
> But this would definitely break existing configs? Opinions?

@Azurebeats 

I don't know how we're going to accomplish moving ItemDespawnTime into the loot options (it does belong there) at the moment because a simple alias can't be done, but I'll look into this in the future.

I would hold off on the rename until we also figure out a solution for ItemDespawnTime so we can knock out both at once.

You can add a `TODO(panda)` next to the variables with a short blurb about what we want to do, and a link to this comment for context.

Intersect.Server/Entities/Npc.cs
Azurebeats: Would likely have to dive further into that, .Max does not work as it is not IComparable
.Max over the value only returns the value.
OrderByDescending().First() could work, but would likely be pissing slow.

Intersect.Server/Networking/PacketHandler.cs
Azurebeats: This suggestion would send the player a message the item does not belong to them if their inventory is full.

That does not sound like it should be intended behaviour. I did forget to add a message stating their inventory is full, however. Did add that. (yet another inventorynospace string)

Azurebeats:
I've updated the PR with a new commit including some requested changes.

Intersect (Core)/Config/LootOptions.cs
Azurebeats: Added some basic XML documentation to this class and its options.

Intersect.Client/Entities/Player.cs
Azurebeats: Had to make additional changes to make this work for both checks.

Intersect.Client/Entities/Player.cs
Azurebeats: No longer relevant as this was removed.
But yes, that was a typo. :)

jcsnider:
I say we just move the 2 config options. I'll make a note in the Beta 6.2 dev blog. This is the dev branch is breaking that a little really isn't a big deal imo.

Azurebeats:
I mean, luckily they're both not very large involving settings. 

But I'd still say it'd be bad practice to just move them and potentially reset some user settings. What if it involves larger more impactful settings in the future?

jcsnider:
In the case in which it's a bigger deal we'd likely put more effort into retaining the settings while moving them over.

Probably use some sort of JObject parsing in Newtonsoft to read the config before loading it in, checking for old/outdated values, move them, then save the new config on launch. 

Not something we necessarily need here though.

Azurebeats:
I mean, if you feel like we can yolo it and move them then that's fine by me.

Intersect.Server/Entities/Npc.cs
Azurebeats: Looked into this a bit further, managed to do it with .Aggregate, but that is still about 50 times slower than the for loop on average.

OrderByDescending().FirstOrDefault() seems to range from between twice as slow to three times as slow.

The fancy schmancy solution people keep mention is MoreLinq's .MaxBy(), however that is slower even than the aggregate solution and takes an entire new library. So that isn't an option either.

The least horrible ''short'' version I've found so far seems to be OrderByDescending().FirstOrDefault().
Though keeping it as is is still faster.

Azurebeats:
Ran into a little issue while looking at something else pertaining to Map NPC vital regens... That could would never execute for NPCs.

https://github.com/AscensionGameDev/Intersect-Engine/pull/181/commits/d4022267e20387c287494e48849ab42976b06f28

Intersect.Client/Entities/Entity.cs
```diff
@@ -846,7 +846,7 @@ public virtual void Draw()
                 //If unit is stealthed, don't render unless the entity is the player.
                 if (Status[n].Type == StatusTypes.Stealth)
                 {
-                    if (this != Globals.Me && !Globals.Me.IsInMyParty(this))
+                    if (this != Globals.Me || this is Player player && !Globals.Me.IsInMyParty(player))
```
lodicolo, line 849: This was my bad... I think this should have actually been this?

`if (this != Globals.Me && !(this is Player player && Globals.Me.IsInMyParty(player)))`

Intersect.Client/Entities/Entity.cs
```diff
@@ -1217,7 +1217,7 @@ public virtual float GetTopPos(int overrideHeight = 0)
...
```
lodicolo, line 1220: Same as before, my bad.

Intersect.Client/Entities/Entity.cs
```diff
@@ -1309,7 +1309,7 @@ public virtual void DrawName(Color textColor, Color borderColor = null, Color ba
...
```
lodicolo, line 1312: Same as before, my bad.

Intersect.Client/Entities/Entity.cs
```diff
@@ -1430,7 +1430,7 @@ public void DrawHpBar()
...
```
lodicolo, line 1433: Same as before, my bad.

Intersect.Client/Entities/Player.cs
```diff
@@ -114,34 +115,9 @@ public bool IsInParty()
...
-        public bool IsInMyParty(Guid id)
-        {
-            foreach (var member in Party)
-            {
-                if (member.Id == Id)
-                {
-                    return true;
-                }
-            }
+        public bool IsInMyParty(Player player) => Party.Any(member => member.Id == player.Id);
```
lodicolo, line 118: I honestly would have done `=> IsInMyParty(player.Id)` just to limit how many methods are doing essentially the same thing, or have done `Party.Contains(player)` here (first is probably better here).

Intersect.Client/Entities/Player.cs
```diff
@@ -901,7 +877,7 @@ public void AutoTarget()
                     if (en.Value.GetEntityType() == EntityTypes.GlobalEntity ||
                         en.Value.GetEntityType() == EntityTypes.Player)
                     {
-                        if (en.Value != Globals.Me && !Globals.Me.IsInMyParty(en.Value))
+                        if (en.Value != Globals.Me || en.Value is Player player && Globals.Me.IsInMyParty(player))
```
lodicolo, line 880: Also my bad, should be `if (en.Value != Globals.Me && !(en.Value is Player player && Globals.Me.IsInMyParty(player)))`.

Intersect.Client/Entities/Player.cs
```diff
@@ -1150,7 +1126,7 @@ public bool TryTarget()
                                 if (en.Value.CurrentMap == mapId &&
                                     en.Value.X == x &&
                                     en.Value.Y == y &&
-                                    (!en.Value.IsStealthed() || Globals.Me.IsInMyParty(en.Value)))
+                                    (!en.Value.IsStealthed() || en.Value is Player player && Globals.Me.IsInMyParty(player)))
```
lodicolo, line 1129: This was fine, doesn't need changing.

Intersect.Client/Entities/Player.cs
```diff
@@ -1215,7 +1191,7 @@ public bool TryTarget()
                                         en.Value.X == x &&
                                         en.Value.Y == y &&
                                         !((Event) en.Value).DisablePreview &&
-                                        (!en.Value.IsStealthed() || Globals.Me.IsInMyParty(en.Value)))
+                                        !en.Value.IsStealthed() || en.Value is Player player && Globals.Me.IsInMyParty(player))
```
lodicolo, line 1194: Pretty sure that removing the parentheses here also changes the logic, so add the parentheses back around so it's like this:
`(!en.Value.IsStealthed() || en.Value is Player player && Globals.Me.IsInMyParty(player)))`

Azurebeats: Woops, missed that.

Azurebeats: Woops, missed that.

Intersect.Client/Entities/Player.cs
Azurebeats: Definitely does, en is a generic entity here.
IsInMyParty no longer accepts those.

Intersect.Client/Entities/Player.cs
Azurebeats: Good point, overlooked that.

Intersect.Server/Entities/Npc.cs
Azurebeats: Making this an extension doesn't exactly speed it up.

It's about half a milisecond, and doesn't leverage OrderBy. Sure as heck aint pretty though.

`public static KeyValuePair<TKey, TValue> MaxByValue<TKey, TValue>(this IDictionary<TKey, TValue> dic) {
    var max = dic.Values.Max();
    var pair = dic.Where(d => d.Value.Equals(max)).FirstOrDefault();
    return new KeyValuePair<TKey, TValue>(pair.Key, pair.Value);
}`

PR #212, opened 2020-05-13
Updating dev with item editor and packet encryption fixes

jcsnider:


jcsnider:
Closed because Panda made #213

Issue #110, opened 2020-04-07
Log player chat


ghost:
**Is your feature request related to a problem? Please describe.**
Yes, players can speak offense, speak ill of staff or anything when adm is not online

**Describe the solution you'd like**
I would like that one day the server would save chat logs of player conversations, this would help in the future in complaints, things of moderation

lodicolo:
Restoring issue.

PR #248, opened 2020-06-06
Added optional chat logging

marsh0:
fixes https://github.com/AscensionGameDev/Intersect-Engine/issues/110

### DIFF
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -127,6 +127,8 @@ namespace Intersect
 
         public static int MaxChatLength => Instance.ChatOpts.MaxChatLength;
 
+        public static bool LogChatMessages => Instance.ChatOpts.LogChatMessages;
+
         public static int MinChatInterval => Instance.ChatOpts.MinIntervalBetweenChats;
 
         public static bool UPnP => Instance._upnp;
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -439,6 +439,11 @@ namespace Intersect.Server.Networking
                 return;
             }
 
+            if(Options.LogChatMessages)
+            {
+                Log.Pretty.Info("Client " + client.GetIp() + " Player: " + player.Name + " Message: " + packet.Message);
+            }
+
             var msg = packet.Message;
             var channel = packet.Channel;
             if (client?.User.IsMuted ?? false) //Don't let the toungless toxic kids speak.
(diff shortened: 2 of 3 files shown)


Intersect (Core)/Config/Options.cs
```diff
@@ -127,6 +127,8 @@ public class Options
 
         public static int MaxChatLength => Instance.ChatOpts.MaxChatLength;
 
+        public static bool LogChatMessages => Instance.ChatOpts.LogChatMessages;
```
Azurebeats, line 130: I believe we are moving away from top level options and are instead moving them to categories.

Ie reference the chatopts instance , rather than the underlying options.

Azurebeats:
I just noticed, albeit late.. But is this really intended for Master?
It's not a bugfix.

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -439,6 +439,11 @@ public void HandlePacket(Client client, Player player, ChatMsgPacket packet)
                 return;
             }
 
+            if(Options.LogChatMessages)
+            {
+                Log.Pretty.Info("Client " + client.GetIp() + " Player: " + player.Name + " Message: " + packet.Message);
```
lodicolo, line 444: This should use `$"{variable}"` syntax and also simplify the message to save space:

`$"{client.GetIp()} {player.Name}: {packet.Message}"`

lodicolo:
> I just noticed, albeit late.. But is this really intended for Master?
> It's not a bugfix.

This definitely should not be targeting `master` (Beta 6.1).

This _is_ however a simple enough feature, so I think it should go into `prerelease` (Beta 6.2) instead.

I can retarget it myself, but since you made the change against master I would prefer you move the change to be based off of prerelease instead and then update this PR's base branch to be prerelease (I don't know if there will be any merge conflicts, and that's why I am not retargetting it myself).

ThousandX:
will this log be generated separately from the server log?

lodicolo:
> will this log be generated separately from the server log?

@ThousandX Based on the implementation, no.

ThousandX:
It would be good to separate, because it will be very messy for those who are going to inspect the log, when you need to report a bug, everything will be together with the players' chat

lodicolo:
Well there is nothing stopping you from opening your own PR that does that instead.

Your original request did not ask for it to be in a separate file though, and this logs into the info log not the error log, so this is not really true in almost all cases:
> when you need to report a bug, everything will be together with the players' chat

marsh0:
I will fix the issues that you guys mentioned and get it on the right branch when I get some more time.

I can move it to a new file to, not a problem.

Intersect.Server/Networking/PacketHandler.cs
marsh0: Sounds good

PR #251, opened 2020-06-08
consumable code updated

RedBandana:
**Features:**
- Can add negative percentage to the amount value of a consumable item. So you have more possibilities for potions. New range of consumable item percentage amount in Item Editor: -100, 100
- No more double "-" for negative value in action message (like "--4") after using a consumable item.

**Issues solved**
- It was not possible to give a negative value to a consumable item in an efficient way using the percentage, because the percentage value is always positive. So you could end with a positive value when trying to remove health. For example, "-10 + 80 % of 100 HP = 70".
- If the initial value to remove was negative (like the "-10" in the previous example), the final symbol in the action message would be a minus even if the final value (the one taking in consideration the percentage) was positive. The player would still get +70 HP.
- When the final value was negative, the action message had two minus symbols. For example "--6".


RedBandana:
Ok I will do that. Probably next week.
I will also check if it's applicable to other mechanism than consumable.

RedBandana:
@lodicolo I realized there are many places where we can add negative value to an amount, but cannot do so for its percentage field. Like the stat bonuses of a weapon. Should I edit all those places or only the one related to consumable?

lodicolo:
@RedBandana As long as it is documented in an issue exactly which ones you are fixing, you can just change as many places as you feel like adjusting.

You definitely need to provide screenshots though that highlight which fields are updated (circled using MS Paint or something like that is good enough).

PR #269, opened 2020-07-06
242: Add logging to better diagnose the issue with the file

### LINKED ISSUE #242: Se virem para descobrir o BUG :)
Se virem para descobrir o BUG :)

lodicolo:
Resolves #242 

I am rather confident #242 is not really an issue in code (at least not ours or that we can fix) and the real solution for us here is to say as much as we can in the logs about which file is problematic, and the file can be adjusted to not cause further exceptions.


ThousandX:
What has changed compared to the Mod version of the forum? Any improvement?

Issue #302, opened 2020-09-02
Movement Behave and Radius for monster.


Spectreye:
**Hello !**

**As discussed with Joyce, the movement behaving of monster can be quite problematic, i quote here "@Mapyo" from the forum who explain it clearly why it can be a problem :** 

_I am working on guards and other NPC's for questing areas in my game, and I came to realize something... NPC's will follow you to the end of the god damn earth!_

_Seriously though, this can be abused to get new players killed by kiting high level mobs to lower level areas and then dropping your aggro by entering a new area or just logging, now I know this sounds like a simple little flaw with how NPC's currrently work but it can really mess with a game._

_Imagine walking into a dungeon to see that players have already been through and has kited all enemies to the end of the dungeon. Not only are you now screwed over because you have tons of enemies on you (which are supposed to be spread out across the dungeon) but you also don't get the experience the developers intended for you to have. _

_This problem also affects guards who stop hostiles from entering city/town/village gates! Sure you could just use an NPC block but having a guard that will defend you and prevent enemies from entering the kingdom is more immersive and really gives the game a nice touch. The problem you ask? Those guards will run after anything they see and will not return to their post ever!_

 

_Theres a solution here that I think we can all agree this engine needs, and that is more NPC Behaviour settings!_

_Here are a few settings I think should be added!_

https://s3.us-east-2.amazonaws.com/ascensiongamedev/filehost/8c4f4e7791658021bf94784abb89106d.png

 

_Roam - This is currently in the engine already, the option is called "Move Randomly". This mode is for when you want the Original style of NPC movement, no changes to what is currently in the engine._

_Guard - This behaviour setting will have the NPC stand still (or rotate randomly) and will attack (if attacked) enemies it sees withing its sight range (or when it gets hit) and will chase it until it the enemy leaves its guarding radius. If the enemy leaves the guarding radius then the guard will return to the spot it was set to spawn at._

_Radius - This behaviour setting will have the NPC roam around within its "Roam Radius" and will attack (if attacked) enemies it sees withing its sight range (or when it gets hit) if the enemy leaves the "Roam Radius" then it would stop attacking the enemy._

_Original idea belongs to @Aesthetic I just wanted to add to it! Find the original post below:_
https://www.ascensiongamedev.com/bug_tracker/suggestions/npc-quality-of-life-r1829/

Azurebeats:
Part of this was added in 6.2.

[...]

PR #303, opened 2020-09-02
Configurable reset radius and behaviour for NPCs after going into combat.

Azurebeats:
Adds features requested in #302 
Request discussed in shoutbox pertained a reset radius, not a new behaviour type though..

Adds the ability to configure several new global options for NPC behaviour through the server configuration:

* AllowResetRadius - Configures whether or not Npcs are allowed to reset after moving out of a specified radius when starting to fight another entity.
* ResetRadius - Configures the radius in which an NPC is allowed to move after starting to fight another entity.
* AllowNewResetLocationBeforeFinish - Configures whether or not the NPC is allowed to gain a new reset center point while it is still busy moving to its original reset point. NOTE: Can be used to allow the NPCs to be dragged far far away, as it constantly resets the center of its radius!!!
* ResetVitalsAndStatusses - Configures whether or not the NPC should completely restore its vitals and statusses once it resets.

After being forcibly reset, the NPC will attempt to move back to its original location.

Had to adjust some of the nonsensical horrors that was the pathfinding system being COMPLETELY reliant on having a target before allowing it to work. Hasn't broken in my testing yet.

### DIFF
--- a/Intersect (Core)/Config/NpcOptions.cs
+++ b/Intersect (Core)/Config/NpcOptions.cs
@@ -1,0 +1,34 @@
+﻿namespace Intersect.Config
+{
+
+    /// <summary>
+    /// Contains configurable options pertaining to the way Npcs are handled by the engine.
+    /// </summary>
+    public class NpcOptions
+    {
+
+        /// <summary>
+        /// Configures whether or not Npcs are allowed to reset after moving out of a specified radius when starting to fight another entity.
+        /// </summary>
+        public bool AllowResetRadius = false;
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


Azurebeats:
Adding an image of what it looks like in practice..

https://www.ascensiongamedev.com/resources/filehost/88f71a7ef08aba063a448ecf5f758043.gif

jcsnider (approved):

jcsnider:
Looks fine to me. A good future enhancement would be a configurable reset radius on a npc by npc basis via the npc editor.

Will let @lodicolo confirm before merging.

lodicolo (approved):

PR #308, opened 2020-09-03
Asset compression, encryption and packing changes

### LINKED ISSUE #304: Asset pack compression
This enhancement is part of a lot of recent requests for a method to make it harder for people to quickly read or modify game assets. I believe there was a discussion about this at some point to ''solve'' the issue regarding easy access for the majority by simply gzipping the data into a compressed package file.

Gzip the texture packs and load them from said zipped files.

**Additional context**
https://www.ascensiongamedev.com/topic/5425-how-do-i-encrypt-resources/#comment-47782
https://www.ascensiongamedev.com/topic/4765-read-first-standing-requests-rules-please-dont-make-threads-for/

Cheshire92:
Resolves #304 

Will stop people from copy pasting assets randomly. Also shouldn't have a heavy impact on loading performance.

Mostly adds some file compression to the texture packs and methods to use for future asset compression. After the data is compressed it's encrypted with a basic form of encryption.

Adjusted the texture packing sequence to also package up music and sound through a generic class that can be implemented in other areas if desired.

Also by jcs request, removed the texture packing option and implemented it within the update pack procedure. Added configurable options and toggles for this behaviour.


ThousandX:
Great job, however an easy way to circumvent it, is to open the .asset file with winrar, you will have the image file inside, drag it to the desktop and rename it to any .png name, the image is there for free access.

One way to circumvent this would be to put a configurable password, so that it is not possible to open without a password.

For example Packs.assets
inside it will have all the original files, but to open it you will need to know the password.

jcsnider:
No, asset files cannot be opened with WinRar and extracted. A special application would be required to unzip the asset files.

I agree that it wouldn't be hard to remove the gzip and gain access to the files, but this will prevent copy/paste theft and require some technical knowledge to bypass.

Providing a password would NOT be a viable solution. The password could be extracted from the clients configuration or source and therefore isn't a viable.

Even if the password protection worked, modified opengl dlls or graphic card drivers could allow dumping of the textures/graphics, or the graphics could be ripped via screenshots. It doesn't matter. 

This is the best solution that we can feasibly add to an open source engine.

You are banned from AGD and our Github, please refrain from creating new accounts.

Thanks.

Azurebeats:
If opening in winrar is a concern we could flip the data arrays around from the file to be written backwards.

Should stop people from just renaming the file to get to it at least?

placidoandrade:
Thanks for the work, Azure. Very happy about it, but I really agree with the person who had the comment banned. Of course, Jc said you can get the source password and everything or use something like OpenGl, I don't know, but in any case it would make it more difficult for users who understand the middleman. Very grateful for what you did and even more grateful to be able to improve. Thanks!

placidoandrade:
Ps: if you can improve

Azurebeats:
There we go, encrypted the data with a very basic level of encryption.

@jcsnider 
@placidoandrade

placidoandrade:
> Lá vamos nós, criptografamos os dados com um nível muito básico de criptografia.
@Azurebeats 
I leave my sincere thanks for this work, I really needed it a lot and I have no words to thank you, Azure.

Azurebeats:
I recall JC saying he wanted to eventually pack up all assets.. So I did some tinkering and have a class that can pack up assets (in configurable batches of files) and create a meta file for them, then search through the meta file for the asset you want and use the returned data to retrieve the file from the packs.

https://gist.github.com/Azurebeats/8c439556f66074d04902201d63065dfc

It's currently a doodle in LinqPad, though I did test it. 
Packing up all default Intersect sound files is about 0.75 seconds
Finding and then unpacking a single sound file is about 0.012 seconds.. (Including writing a readable file back to the disk for testing purposes)

Is this something we can use?

Azurebeats:
Got bored and used the previous post's idea to improve and change it a bit,
Have a working prototype for packaging up and loading sound assets now: https://github.com/Azurebeats/Intersect-Engine/commit/a05c4d8890e34c385f778c9aeb60cfffdeddc036

Is that roughly what you meant by wanting to package up all assets eventually, @jcsnider ?

Azurebeats:
Note to self make asset batch sizes configurable through the editor.

Azurebeats:
@jcsnider As requested, I've made an options menu that configures whethere packaging an update packages up the assets or not.
As well as options regarding HOW the assets are packed up.

Preview

I have **NOT** changed the way textures are packed up whatsoever, so it creates a meta file for each texture file even if you force it to package up one each by selecting 1.

Intersect.Client/MonoGame/File Management/MonoContentManager.cs
```diff
@@ -261,6 +279,22 @@ public override void LoadMusic()
...
                 mMusicDict.Add(RemoveExtension(filename), new MonoMusicSource(Path.Combine(dir, filename)));
             }
+
+            // If we have a music index file, load from it!
+            if (File.Exists(Path.Combine("resources", "packs", "music.index")))
+            {
+                MusicPacks = new AssetPacker(Path.Combine("resources", "packs", "music.index"), Path.Combine("resources", "packs"));
+                foreach (var item in MusicPacks.FileList)
+                {
+                    if (!mMusicDict.ContainsKey(RemoveExtension(item)))
```
jcsnider, line 289: !mMusicDict.ContainsKey(RemoveExtension(item)).ToLower()

Gotta check for the existing key in lower case since we're adding the keys as lower case.

Intersect.Client/MonoGame/File Management/MonoContentManager.cs
```diff
@@ -244,6 +245,23 @@ public override void LoadSounds()
...
                 );
             }
+
+            // If we have a sound index file, load from it!
+            if (File.Exists(Path.Combine("resources", "packs", "sound.index")))
+            {
+                SoundPacks = new AssetPacker(Path.Combine("resources", "packs", "sound.index"), Path.Combine("resources", "packs"));
+                foreach(var item in SoundPacks.FileList)
+                {
+                    if (!mSoundDict.ContainsKey(RemoveExtension(item)))
```
jcsnider, line 255: Should we be adding to this dictionary as lower case like the music dictionary below?

jcsnider (changes requested):
I'm a Github pleb and can't figure out how to comment on lines that the PR technically doesn't modify.. but in frmMain.cs line 1913 we have the following which determines the ClientIgnore value...
`if (clientExcludeFiles.Contains(relativePath))`

We need to change that to:
`if (clientExcludeFiles.Contains(relativePath.ToLower()))`

So that the client doesn't download files it shouldn't if they have uppercase letters lol.

Intersect.Client/MonoGame/File Management/MonoContentManager.cs
Azurebeats: Hmm, might have to wait and see if that's still the case when Panda is done with his refactor.

Intersect.Client/MonoGame/File Management/MonoContentManager.cs
Azurebeats: Would be more consistent, but if Panda is going to change everything around that might not be the case anymore? I think he wanted to make stuff case insensitive?

Azurebeats:
`We need to change that to:
if (clientExcludeFiles.Contains(relativePath.ToLower()))

So that the client doesn't download files it shouldn't if they have uppercase letters lol.`

If I recall, that was already done in a different PR that came after this one and just isn't on this branch.

lodicolo (approved):

lodicolo (dismissed):

jcsnider (approved):

PR #310, opened 2020-09-04
fix 309: Check if counterparty accepted

### LINKED ISSUE #309: Trading Scamming
**Describe the bug**
When you do a trade and if the other one confirm his trade, you can get your items back in the inventory and still finalise the trade quickly scamming the other player.

**Version Encountered**
0.6.2.429 

**To Reproduce**

1. Go in game and launch a trade with another player
2. Trade one item and when the other one confirm right click to get it back to the inventory and confirm the trade
3. You've scammed somebody (Nothing to be proud of ! )

**Expected behavior**
When a player confirmed the trade it should lock the item in place, the other player can either stop the trade or confirm on his own.
Another possible solution is to put another "confirm?" Button when to player have confirmed their initial trade.

**Operating System:**
Windows, Mac, Or Linux

**Additional context**
Wasn't sure if it was considered as a bug or as an enhancement, but either way it's a dangerous and not fiable way to trade.

alloin:
Resolves #309

### DIFF
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -1643,6 +1643,11 @@ namespace Intersect.Server.Networking
         //RevokeTradeItemPacket
         public void HandlePacket(Client client, Player player, [NotNull] RevokeTradeItemPacket packet)
         {
+            if (player == null || player.Trading.Counterparty == null)
+            {
+                return;
+            }
+
             if (player.Trading.Counterparty.Trading.Accepted)
             {
                 PacketSender.SendChatMsg(player, Strings.Trading.revokenotallowed.ToString(player.Trading.Counterparty.Name), CustomColors.Alerts.Declined);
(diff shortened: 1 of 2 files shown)


Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1643,7 +1643,14 @@ public void HandlePacket(Client client, Player player, [NotNull] OfferTradeItemP
         //RevokeTradeItemPacket
         public void HandlePacket(Client client, Player player, [NotNull] RevokeTradeItemPacket packet)
         {
-            player?.RevokeItem(packet.Slot, packet.Quantity);
+            if (player.Trading.Counterparty.Trading.Accepted)
```
Azurebeats, line 1646: There's a null handler on the original method, and I see you copied that over. But I don't see it in your additions?

alloin: I didn't think there would be a NRE here, assuming the trade is still happening and Accepted is a bool.

Should i add ```
            if (player == null || player.Trading.Counterparty == null)
            {
                return;
            }``` in front of it?

Azurebeats: I'm not sure, in theory you'd assume you never could hit an NRE there.. unless a player somehow manages to not be on the server anymore. WHich is rare, but not impossible.

It was mostly a discrepency I caught reading your changes, mind. But personally, I'd try and keep it consistent. If the old code checks for it, the new should do the same.

PR #319, opened 2020-09-18
Sprite Options (Frames/Timing/Etc)

### LINKED ISSUE #316: Make number of sprite movement frames configurable
Right now the movement frames for sprites and paperdolls are hardcoded as 4. The same goes for animations like attacking, casting, etc.

Those hard coded values should be pulled out into the server configuration at the very least.

**Additional context**
I'm about to tackle this for a project I'm working on, so I might as well share it here.

jcsnider:
Pulls out options for # of sprite frames, time before idling, and more into the server config instead of being hard coded into the client/editor.

Changes a lot of "/4" and "/8" into meaningful and readable values.

Resolves #316

### DIFF
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -56,7 +56,7 @@ namespace Intersect
 
         [JsonProperty("Loot")] public LootOptions LootOpts = new LootOptions();
 
-        [JsonProperty("Sprites")] public SpriteOptions SpriteOpts = new SpriteOptions();
+        public SpriteOptions Sprites = new SpriteOptions();
 
         [JsonProperty("Npc")] public NpcOptions NpcOpts = new NpcOptions();
 
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -56,6 +56,8 @@ namespace Intersect
 
         [JsonProperty("Loot")] public LootOptions LootOpts = new LootOptions();
 
+        [JsonProperty("Sprites")] public SpriteOptions SpriteOpts = new SpriteOptions();
+
         [JsonProperty("Npc")] public NpcOptions NpcOpts = new NpcOptions();
 
         public SmtpSettings SmtpSettings = new SmtpSettings();
--- a/Intersect (Core)/Config/SpriteOptions.cs
+++ b/Intersect (Core)/Config/SpriteOptions.cs
@@ -1,9 +1,4 @@
 ﻿using Newtonsoft.Json;
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Intersect.Config
 { 
--- a/Intersect (Core)/Config/SpriteOptions.cs
+++ b/Intersect (Core)/Config/SpriteOptions.cs
@@ -1,0 +1,75 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Intersect.Config
+{ 
+    /// <summary>
+    /// Contains configurable options pertaining to the way sprites are rendered within the engine
+    /// </summary>
+    public class SpriteOptions
+    {
+        /// <summary>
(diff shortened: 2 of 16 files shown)


Intersect.Client/Entities/Entity.cs
```diff
@@ -1636,6 +1635,27 @@ public void SortStatuses()
             Status = Status.OrderByDescending(x => x.RemainingMs()).ToList();
         }
 
+        private float GetSpriteFrames()
```
lodicolo, line 1638: This should be a property, `SpriteFrames`.

lodicolo: Also why is this a `float`?

Intersect (Core)/Config/Options.cs
```diff
@@ -56,6 +56,8 @@ public class Options
 
         [JsonProperty("Loot")] public LootOptions LootOpts = new LootOptions();
 
+        [JsonProperty("Sprites")] public SpriteOptions SpriteOpts = new SpriteOptions();
```
lodicolo, line 59: There is no `Sprites` field to conflict with this, just call it `Sprites` and drop the `JsonProperty` attribute since it won't be necessary.

lodicolo (changes requested):

lodicolo:
Added some review comments, also added @Azurebeats as a reviewer so we can get some more eyes on this.

Intersect (Core)/Config/SpriteOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Defines how long (in ms) a player must idle before the idling sprite starts to render
+        /// </summary>
+        public int TimeBeforeIdle = 4000;
+
+        /// <summary>
+        /// Defines the number of rows in sprite sheets which will correlate to the number of directions in the game (Intersect is programmed by default with only 4 directions)
+        /// </summary>
+        [JsonIgnore] public int Directions => 4;
```
Azurebeats, line 73: Is there a reason this isn't just Directions = 4; ?

Intersect (Core)/Config/SpriteOptions.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
Azurebeats, line 6: Why all of these unecessary references?

Azurebeats (changes requested):

Intersect (Core)/Config/Options.cs
Azurebeats: Just as an aside, shouldn't changing this for the other options categories become a cleanup issue then?

Intersect (Core)/Config/SpriteOptions.cs
jcsnider: => makes it readonly, it's not something that should ever be changed at runtime, especially since changing that value won't provide 8 directional movement or anything..

It is not stored in the config right now because changing that value won't automatically provide 8 directional movement or anything.

Azurebeats: Ah, fair enough. That reasoning makes sense!

Intersect (Core)/Config/SpriteOptions.cs
jcsnider: Resolved.

Intersect.Client/Entities/Entity.cs
jcsnider: Shouldn't have been a float. Fixed and is now a property.

Intersect (Core)/Config/Options.cs
jcsnider: Maybe? We should decide on this. Changing the name of this variable only would make referencing options pretty inconsistent.

lodicolo: I plan on cleaning up the configuration rather soon anyway, I'm just trying to not add more code that we're just going to redo later on.

jcsnider: Done.

lodicolo (approved):

PR #331, opened 2020-09-25
Add Chatbox tabs and the ability to tag messages with categories.

### LINKED ISSUE #329: Chat channel tabs

There are no chat tabs

Change the chat channel dropbox into tabs on top of / below the chatbox and add a System tab. Have it so that you can set messages as a system message. For example in events, xp drops, level up messages, etc. 

**Additional context**

Above is a basic system but you can expand by adding filters to the tabs so players can choose which messages they see in each tab.

Cheshire92:
Resolves #329 partially, future PR possibilities to make the chatbox tabs user configurable rather than hardcoded and expose an option to mark event chat additions with specific tags.

Went through every single chat message I could find and marked them accordingly.
Documented some chatbox related stuff I came across.

Made it so the generic tabs (All and System) remember the last chat channel you set for them, the others will change the channel to the relevant channel by default.

Adding new tabs is relatively easy, just need to configure what they should display, default the channel to and say in the source code, and add the localization stuff for them.

Image

Client resources: https://www.ascensiongamedev.com/resources/filehost/398ddf1e96ace2414aad42ce02cd444e.zip (Updated)


Azurebeats:
Fixed a couple of small things and changed the UI for it a little to look less.. slapped together.

jcsnider:
Looks good to me.

jcsnider (approved):

jcsnider (approved):

PR #336, opened 2020-09-26
Cleanup client framework

lodicolo:


Azurebeats:
This seems to function, but loading the game after logging in appears to be SIGNIFICANTLY slower.
It freezes the entire client for about 3-4 seconds for me now on a black screen before it gets to the game window.

Azurebeats:
As a matter of fact, loading a new tileset swapping between maps takes 1-2 seconds as well for me.

lodicolo:
I suppose that explains what I was seeing yesterday, I'll have to dig through the loading logic.

lodicolo:
Reopening this on a separate branch

PR #338, opened 2020-09-27
Configurable Map Layers

Azurebeats:
I believe JC mentioned he wanted this in at some point.
Essentially just [this](https://www.ascensiongamedev.com/topic/5266-15-kash-shop-unlimited-additional-map-tile-layers/) mod applied through a PR.

The [manual](https://s3.us-east-2.amazonaws.com/ascensiongamedev/filehost/906423eabd55bc3c82fbeaf6c109e352.pdf) for it makes it sound a heck of a lot more complex than it is though, may need to write up a proper doc for that.

Any changes that need to be made in general? I skimmed through it and made some changes to the codebase for consistency sake. Haven't really got an existing game with maps to test this on myself though.. Doesn't look like it would break existing games however.


jcsnider:
So I am very impressed with how they dealt with this as a patch.. but I think as an enhancement to the base engine we should handle it differently.

Just as a quick thought/overview... we could have the tile layers defined in the server config file like this:
```
  "Map": {
    "GameBorderStyle": 0,
    "Height": 26,
    "ItemAttributeRespawnTime": 15000,
    "TileHeight": 32,
    "TileWidth": 32,
    "Width": 32,
    "ZDimensionVisible": false
	"Layers": {
      "Lower": [
        "Ground",
        "Mask",
        "Mask2",
      ],
      "Middle": [
        "Fringe",
      ],
      "Upper": [
        "Fringe2"
      ],
	}
  },
```

From a design standpoint, we could change the TileLayers array for each map to a Dictionary<string,byte[]> where the key is the layer defined in the config.

There could be a TrimLayers option in the config to remove layers from the database when removed from the config, but otherwise we could technically keep deleted layers (ie if they removed 'Ground') and then realized they wanted to bring it back later.

We could remove layer names from the localization files and just pull them from the server config as well.

Sad part is in the editor we'd have to remove Zeta's icons :(.

In theory the worst part is the DB migration of the TileLayers to a Dictionary, but that isn't too terrible.

I do think this should be easy to configure though just through the server config.

Azurebeats:
That would be preferable yes, hence why this is a draft rather than an actual PR.
Conceptually okay, but not something we should probably do as it's not exactly user friendly.

I'd have to look into how to make migrations like that possible. lol

Azurebeats:
See #379 for better solution

PR #339, opened 2020-09-27
Packet timestamps updated

### LINKED ISSUE #141: Smarter speedhacking prevention
Our current speedhacking prevention isn't favorable to higher ping environments and causes player ruberbanding effects which are super annoying for legit players.

Panda's Plan.

**Additional context**
```
well what I'm suggesting is game time
independent of clock time
server periodically reminds the client how long they have been connected and what the current "server" clock time is
but that's only to prevent client side glitches
the client's "game time" is what it provides to the server
and the server simply doesn't let you take too many actions within a specified amount of time
basically if you receive two move packets within 10ms
but one was actually sent 500ms earlier
the timestamp in the packet will indicate that
and the server will know the client ping
and if the ping * some factor - 500ms > 0 then it allows that "old" movement packet
if it's below 0
that means someone that has a ping of lets say 100 is trying to speed hack
and the server just ignores it
the "factor" should either be something static like 1.1
(not a huge margin of error on ping)
or it should be max(current_ping, average_ping)
average being a rolling average
because people could have ping spikes
but a minute or so long rolling average should account for true shitty ping spikes
and intentional network unplugs
```

Needs to apply to attacking as well.

### LINKED ISSUE #265: Players are able to speed hack with (projectile) attacks
**Describe the bug**
Players are able to use speed hack for attacks.
For movement speed they can't use it, but they can for normal E attacks

**Version Encountered**
0.6.2.0

**To Reproduce**
Just use Cheat Engine and activate Speed Hacking

**Video**
https://youtu.be/IwfGfU23H7E

**Operating System:**
Windows, Mac, Or Linux

lodicolo:
Resolves #265 
Resolves #141


jcsnider:
Resolves #141

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -212,6 +212,109 @@ public bool HandlePacket(IConnection connection, IPacket packet)
...
+                var localOffsetMs = Timing.Global.MillisecondsOffset;
+                var localUtcMs = localOffsetMs + localAdjustedMs;
+
+                var remoteAdjusted = timedPacket.Adjusted;
+                var remoteAdjustedMs = remoteAdjusted / TimeSpan.TicksPerMillisecond;
+                var remoteUtcMs = timedPacket.UTC / TimeSpan.TicksPerMillisecond;
+                var remoteOffsetMs = timedPacket.Offset / TimeSpan.TicksPerMillisecond;
+
+                var deltaAdjusted = localAdjustedMs - remoteAdjustedMs;
+                var deltaWithPing = deltaAdjusted - ping;
```
Azurebeats, line 233: Would this not be deltaWithoutPing ?

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -212,6 +212,109 @@ public bool HandlePacket(IConnection connection, IPacket packet)
...
+
+                var configurableMininumPing = 10;
```
Azurebeats, line 235: Marked as configurable, but hardcoded?
Same with the next few variables.

Azurebeats (commented):
Not sure this is of any help, but just found these little things. :)

Intersect.Server/Networking/PacketHandler.cs
jcsnider: I think it reads deltaWithPing because it is delta taking ping into account. Idk. @lodicolo?

Azurebeats (approved):

Intersect (Core)/Config/PacketOptions.cs
```diff
@@ -6,24 +6,86 @@ namespace Intersect.Config
     public class PacketOptions
```
Azurebeats, line 6: This file is never added to the options file, so is never actually made configurable.

Azurebeats (changes requested):

Intersect (Core)/Config/PacketOptions.cs
jcsnider: PacketOptions are a part of SecurityOptions which are in the Options file

An argument could be made for the new options being in a different subclass than packet flooding, maybe packet timing instead?

Settings

Azurebeats: That's odd, mine refused to generate those options so I figured they weren't linked properly.

I'll have to take another look when I get near a computer again.

jcsnider: Maybe we just rename the PacketFlooding header to PacketSanitization?

jcsnider: Per the convo in Discord with @lodicolo I renamed that section to just be 'Packets"

Azurebeats (approved):
Tested it again, turns out our good ol' pal Git doesn't pull changes for PR's I download locally, and I have to wipe them entirely locally.. and redownload them. Weird.

PR #340, opened 2020-09-28
Allow clipboard access cross platform

### LINKED ISSUE #178: Allow clipboard access in chat
N/A

Allow users to use the clipboard in chat.

Azurebeats:
Resolves #178 

I know JC asked to put the clipboard logic in the client itself, but the handlers are all in the Client Framework and adding the accellerators outside the components to each textbox instance we create does not sound fun.

Unless someone has a better option?

Also warns users they are missing the appropriate Linux libraries upon logging in through the chat box. Wonder if it might be cleaner to throw them a warning when they open the client?

### DIFF
--- a/Intersect.Client.Framework/GenericClasses/Clipboard.cs
+++ b/Intersect.Client.Framework/GenericClasses/Clipboard.cs
@@ -1,31 +1,229 @@
-﻿namespace Intersect.Client.Framework.GenericClasses
-{
+﻿using System;
+using System.Diagnostics;
+using System.Windows.Forms;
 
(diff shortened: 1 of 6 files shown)


jcsnider:
It shouldn't be a problem, since the code will never execute, but we've had issues with referencing System.Windows.Forms on OSX in the past since they never compiled that library with Mono for x64 targets. I will test this on OSX later before approving, then I can make a few small adjustments in a future PR.

jcsnider (approved):

Azurebeats:
I'm aware, worked on my Mojave system though as the code never executes.

PR #343, opened 2020-09-30
Scrollback on event editor, event map previews

### LINKED ISSUE #236: Event Graphic on Map
By the description of Enhancement idea - Suggest a minor enhancement or quality of life change. There is something that could improve the lives of all game creators on the intersect. The events have the option of choosing graphics, sprites or tilesets, so that they can be represented ingame. A good improvement for us creators would be that in the editor, when we place an event on the map and a graphic for it, that the graphic we choose appears in the editor. Kind of like when we put animation on the map, the animation already appears, and there’s no single A. Even when we change properties to tileset or event, the animation is still there. For events it would be great, because as the intersect is being designed to be a mmo maker, and some games tend to be big, so the creators don't look for event by event for the event they want to modify, they could just see the graph of the event right away the map, making life easier for everyone.

Additional - Events that have an animation, the animation for that event could also appear on the map. If it's too much, only the top one would be great.

### LINKED ISSUE #344: Scroll back to where you left off in event editor
Currently upon adding or removing event commands in the editors it does not remember where you left off which becomes a bit of a chore in longer events.

A simple scroll back down to where the last change was made.

Azurebeats:
resolves #236
resolves #344 

This PR accomplishes two little things that might help out our more avid Event users:

- The event command list will attempt to remember where the user last added/removed a command and tries to select the item again to scroll back to it. (Especially helpful in LOOOOOOOOONG event command lists)
- The map editor has received an option to render Event previews in the map editor screen, rather than only the square E symbol on the event tab. The Event graphics are always rendered, unless the option under View is not ticked.

image

### DIFF
--- a/Intersect.Editor/Core/Graphics.cs
+++ b/Intersect.Editor/Core/Graphics.cs
@@ -1593,7 +1593,7 @@ namespace Intersect.Editor.Core
 
                     if (height > Options.TileHeight)
                     {
-                        destinationY -= Options.TileHeight;
+                        destinationY -= (height - Options.TileHeight);
                     }
 
                     if (width > Options.TileWidth)
--- a/Intersect.Editor/Core/Graphics.cs
+++ b/Intersect.Editor/Core/Graphics.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Imaging;
+using System.Linq;
 using System.Windows.Forms;
 
 using Intersect.Editor.Content;
@@ -45,6 +47,9 @@ namespace Intersect.Editor.Core
 
         //Resources
         public static bool HideResources = false;
+
+        // Events
+        public static bool HideEvents = false;
 
         //Advanced Editing Features
         public static bool HideTilePreview = false;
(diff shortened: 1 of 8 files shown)


Intersect.Editor/Core/Graphics.cs
```diff
@@ -1453,6 +1480,131 @@ bool alternate
...
+                            sourceY = (int)tmpGraphic.Y * Options.TileHeight;
+                            width = (tmpGraphic.Width + 1) * Options.TileWidth;
+                            height = (tmpGraphic.Height + 1) * Options.TileHeight;
+
+                            break;
+                    }
+
+                    if (height > Options.TileHeight)
+                    {
+                        destinationY -= Options.TileHeight;
```
jcsnider, line 1596: Change to the following so the rendering matches the client.
`destinationY -= (height - Options.TileHeight);`

jcsnider (changes requested):
Think it's good otherwise.

Intersect.Editor/Core/Graphics.cs
Azurebeats: My bad, I thought I mimicked it but I missed that one. :)

jcsnider (approved):

PR #354, opened 2020-10-01
Log packet error info

jcsnider:


Azurebeats (approved):
Log that sweet delicious data!

PR #363, opened 2020-10-02
Make Conditional Branch else optional

### LINKED ISSUE #360: Optional Else
It is not a problem, it is more of a life improvement idea. I believe the else of any conditional branch in event commands should be optional, something like this below.
image

I cited the image above as a solution. The "solution" is to avoid leaving the giant scripts unnecessarily, through more organized events as it will not always be used. Like the following examples:
image
image

Azurebeats:
resolves #360 

I realised I based this off of pre-release a bit late. Can change if need be?

image1
Image2


jcsnider (approved):

jcsnider:
Looks great to me. I'm fine with it in prerelease.

PR #369, opened 2020-10-03
Add option for map animations to be considered blocks

### LINKED ISSUE #368: Allow map animations to also be blocks
Richy is placing torches and other animations within Ambardia in which the player should not be able to traverse through.

His only option to have the tile blocked while also showing the animation would be to use an event which is bad if used across the entire game world for simply blocking off a tile.

The PR I am about to link to this issue allows map animation attributes to also act as blocks.

**Additional context**

jcsnider:
Resolves #368 

Editor preview
newopt


Azurebeats (approved):
Looks good to me.

PR #373, opened 2020-10-03
Fixes window sizing in Linux while updating.

### LINKED ISSUE #372: Updater does not render in the correct resolution on Linux
Screenshot from 2020-10-03 15-13-28

**Describe the bug**
Updater window size is incorrect on Linux

**Version Encountered**
0.6.2.442

**To Reproduce**
Run the client on linux with the updater configured

**Expected behavior**
An 800x480 window centered in the screen.

**Screenshots**
Attached

**Logs**
N/A

**Operating System:**
Linux

**Additional context**
NA

jcsnider:
Resolves #372


Azurebeats (approved):
Interesting that the order makes a difference.

PR #379, opened 2020-10-06
Dynamic Map Layers & Extendable Attributes

### LINKED ISSUE #378: Dynamic Map Layers
Intersect should not restrict players to 5 map layers.

Server layers should be configurable in the server config. The config should provide the ability to add and remove lower, middle, and upper map layers by name.

Layers that are removed should remain in the db, but be hidden on the client/editor. The server should have an option to prune/destroy any orphaned (removed) layers upon launch if desired to keep db sizes small.

**Additional context**
From a migration perspective, I need to fix attributes so they would be extendable/expandable so adding properties and whatnot would work. It just made sense to fix up the map layers at the same time as well.

jcsnider:
Resolves #378 

I believe I have everything working, but I wouldn't mind an extra set or two of eyes and testing of the migration.

Ceras was giving me a ton of issues trying to compress/decompress the tile arrays and attributes using translated type names. The newer versions of Ceras wouldn't decompress our data at all, and the older/current version would lose guid values. 

In the end, I moved everything that we store in the db (at rest) using Ceras over to compressed json using Newtonsoft. It is much more reliable and consistent with an acceptable increase in storage sizes.

Feel free to check my work on that, I spent countless hours debugging and playing with settings in order to make Ceras work, in the end it'll be best to solely use it for packet serialization.

I really like and wanted to keep our map layer icons when possible, so if the total number of layers is <= 5 then the icons will show up and label themselves correctly. Right clicking the icons (as before) will hide the respective layers:
layericons

If you have 6 or more layers then you will see the nasty combobox as normal:
layers

If we have the combobox then we don't have a way to clickly right click to toggle layer visibility, so I have added that as an option under the View menu:
layervisibility

Oh, and here's a preview of the updated config...
config

If DestroyOrphanedLayers is true then on server launch, all map layers that are not in the config (orphans) are deleted from the Layers dictionary. Also tested and seems to be working well.

I /think/ that about covers it. Probably a few bugs left to work through but overall seems good so far.


Azurebeats:
the labels will make no sense with relation to the player icon though, if a user sets 5 layers up with four rendering above the player. Is that really a good thing to keep just because they look nice? It can be misleading.

jcsnider:
If it means keeping that visual layout I can go in and either remove the player face, or make a function to render the icons using GDI so they make sense. 

I like the idea of keeping the 1-click method of selecting and hiding layers which I think will apply to 95% of users and I'm absolutely willing to do the work and code maintenance to keep it in there.

jcsnider:
Pushed those changes so the layer icons are drawn dynamically based on the config. No longer confusing :)

Intersect (Core)/GameObjects/Maps/MapBase.cs
```diff
@@ -45,30 +52,19 @@ public MapBase(Guid id) : base(id)
...
-                        for (var y = 0; y < Options.MapHeight; y++)
-                        {
-                            Layers[i].Tiles[x, y] = new Tile();
-                        }
-                    }
-                }
-
-                TileData = mCeras.Compress(Layers);
+                Layers = new Dictionary<string,Tile[,]>();
+                //TODO: Add new layers from config? Remove layers if they no longer exist? Idk!
```
Azurebeats, line 58: What is this todo for exactly?
Isn't this exactly what the proposed changes do?

Intersect.Editor/General/Globals.cs
```diff
@@ -46,7 +46,7 @@ public static class Globals
 
         public static int CurrentEditor = -1;
 
-        public static int CurrentLayer = 0;
+        public static string CurrentLayer = "";
```
Azurebeats, line 49: Shouldn't this be string.Empty?

Intersect.Editor/Localization/Strings.cs
```diff
@@ -4301,15 +4301,6 @@ public struct Tiles
...
 
-            public static Dictionary<int, LocalizedString> layers = new Dictionary<int, LocalizedString>
-            {
-                {0, @"Ground"},
-                {1, @"Mask"},
-                {2, @"Mask 2"},
-                {3, @"Fringe"},
-                {4, @"Fringe 2"},
-            };
-
```
Azurebeats, line 4312: I recall Panda wanting me to NOT have unlocalizable strings in the config, so I don't think this should be removed but changed instead?

Intersect.Server/Database/GameData/Migrations/CerasVersionToleranceMigration.cs
```diff
@@ -0,0 +1,332 @@
...
+
+            connection.Close();
+            connection.Open();
+            if (updates.Count > 0)
+            {
+                var trans = connection.BeginTransaction();
+
+                using (var updateCmd = connection.CreateCommand())
+                {
+                    updateCmd.CommandText = "";
```
Azurebeats, line 127: string.Empty?

Azurebeats (changes requested):
I've read through the thing, functions fine but might need some extra testing on some more filled worlds. All my test stuff tends to be flat ground. :P

Intersect (Core)/GameObjects/Maps/MapBase.cs
jcsnider: I actually handled that logic elsewhere.. I just pushed a commit to remove the comment.

Intersect.Editor/Localization/Strings.cs
jcsnider: So is the goal right now to have a Dictionary in localization of Config String -> Displayed String?  I guess I could do that.

Cheshire92 (dismissed):

Cheshire92 (approved):
Don't think there's much of a reason to keep postponing this one.

Intersect (Core)/Compression/LZ4.cs
```diff
@@ -0,0 +1,41 @@
...
+            return LZ4Pickler.Unpickle(data);
+        }
+
+        public static byte[] PickleString(string data)
+        {
+            return LZ4Pickler.Pickle(Encoding.Default.GetBytes(data), LZ4Level.L12_MAX);
+            var memoryStream = new MemoryStream();
+            var s = new StreamWriter(memoryStream);
+            s.Write(data);
+            return LZ4Pickler.Pickle(Encoding.Default.GetBytes(data), LZ4Level.L12_MAX);
```
lodicolo, line 29: Double return statements?

Intersect (Core)/Compression/LZ4.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public static string UnPickleString(byte[] data)
+        {
+            return Encoding.Default.GetString(LZ4Pickler.Unpickle(data));
+            data = LZ4Pickler.Unpickle(data);
+            var memoryStream = new MemoryStream(data);
+            var s = new StreamReader(memoryStream).ReadToEnd();
+            return s;
```
lodicolo, line 38: Double return statements?

lodicolo (changes requested):

Intersect (Core)/Config/LayerOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                //Must have at least 1 map layer!
+                throw new Exception("Config Error: You must have at least 1 map layer configured! Please update your server config.");
+            }
+
+            foreach (var reserved in reservedLayers)
+            {
+                if (All.Contains(reserved))
+                {
+                    throw new Exception("Config Error: Layer '" + reserved + "' is reserved for editor use. Please choose different naming for map layers in your server config.");
```
lodicolo, line 70: Interpolation is better than concatenation

Intersect (Core)/Config/LayerOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Intersect.Config
+{
+    public class LayerOptions
+    {
+        public const string ATTRIBUTES = "Attributes";
```
lodicolo, line 13: Use the strings as the variable names, and change the values to `nameof`.

lodicolo:
Resolve the conflicts in `Ceras.cs` as well. I'm pretty sure they're all related to JB Annotations being removed.

lodicolo (approved):

Arufonsu:
i'm a bit late into this party, i really think this is a **must**, for the ones PRing this, please consider that _animations are currently limited to few layers as well_, it would be nice to have a layer under the "ground" layers in order to place ocean/lava animations!

PR #385, opened 2020-10-08
Map item rework, drawing names on mouse over and a loot window.

### LINKED ISSUE #246: Mouse hover over an item on the map would show quantity and name
It would be nice when you hover over an item in game that it would show the amount of the item and then the item name.  

Add some sort of feature that allows a mouse hover over an item to show the quantity and the name.  Would probably be similar to the same way mouse hover on player, npc, or event will show the brackets around them.  Example of what it would say - 300 Gold - 2 Health Potion - Snake Rattle

**Additional context**
mouse hover item

Cheshire92:
resolves #246 

This PR does a couple of different things:
- Rework the Map Item system to be position based, rather than a single dimension we need to loop through to see if we have items on a tile.
- Adds a basic method to draw map item names when the mouse hovers over an item on the map. (Includes translation and custom colour support for customizing what this looks like)
- Add a small loot window on the top right of the client UI when a player is standing on loot. This lets them pick up a specific item from the pile or loot them all as well as show a description of the items.
- Actually render the map items in the correct order. Before they were rendered the opposite way you'd pick them up in.

image
image2
image

Additional resources: [Resources](https://www.ascensiongamedev.com/resources/filehost/0cf673d0181139696c1321f621e6cd88.zip)


Azurebeats:
Should probably change the way map items are handled rather than making a second dataset for them.

Azurebeats:
Not 100% happy with this yet, but it works.

image

Window can be moved by users, and clicking it makes you attack somehow?.
Needs some changes.

Azurebeats:
Okay, this ended up being a bit bigger than I expected. lol

Cheshire92:
Thanks for the fix. :)

Intersect.Client/Entities/Player.cs
```diff
@@ -1237,30 +1237,58 @@ private void ClearTarget()
...
         {
             var map = MapInstance.Get(CurrentMap);
             if (map == null)
             {
                 return false;
             }
 
-            foreach (var item in map.MapItems)
+            var location = new Point(x, y);
+            if (!map.MapItems.ContainsKey(location) && map.MapItems[location].Count < 1)
```
jcsnider, line 1257: Should be || not &&

Intersect.Client/Interface/Game/MapItem/MapItemWindow.cs
```diff
@@ -0,0 +1,167 @@
...
+                mMapItemWindow.Hide();
+                return;
+            }
+
+            var location = new Point(Globals.Me.X, Globals.Me.Y);
+            var mapItems = Globals.Me.MapInstance.MapItems;
+            if ((!mapItems.ContainsKey(location) || mapItems[location].Count < 1) && !mMapItemWindow.IsHidden)
+            {
+                mMapItemWindow.Hide();
+            }
```
jcsnider, line 69: Needs to account for Globals.Me.MapInstance possibly being null (during warps between areas) and hiding the window in that case.

jcsnider:
```
            var mapItems = Globals.Me.MapInstance?.MapItems;
            if (mapItems == null || (!mapItems.ContainsKey(location) || mapItems[location].Count < 1) && !mMapItemWindow.IsHidden)
            {
                mMapItemWindow.Hide();
            }
            else if ( mapItems != null && mapItems.ContainsKey(location) && mapItems[location].Count > 0 && mMapItemWindow.IsHidden)
            {
                mMapItemWindow.Show();
            }



```

jcsnider (changes requested):

Intersect.Client/Entities/Player.cs
Cheshire92: Is that tested? Maybe my demented mind might be wrong but that sounds like it might cause an NRE.

I'll have to look into that when I get the time.

Cheshire92: I am dumb, and should not say things at 6 in the morning.

Cheshire92:
Loot window apparently crashes when there's more items than slots and nobody told me. Derp.

lodicolo (approved):

PR #389, opened 2020-10-13
Fix entities able to walk on resources after respawning and resources respawning with Npcs and Players on them.

### LINKED ISSUE #388: Walkable Ressources and Monster respawn on ressources
**Describe the bug**
Monster seems to be able to walk on Ressources( after it had respawned one time i think), player can also dash through those or inside but cannot "move" after. The problem is it can happen on the same ressources ( I mean a spot can bug and another of the same type don't). 

Also, monster can respawn on depleted ressources, leading to "trapped" monster inside ressources.

**Version Encountered**
0.6.2.434

**To Reproduce**
I'm not sure about what's affecting the ressources, my guess would be probably about the respawn mechanics,
Because when ressources spawn for the "first" time there's no problem, and after a respawn they become walkable by monster and dashable.

Also  : My ressources are "walkable after ressources removal" not before, and the dash do not ignore any collision.

**Expected behavior**
-Ressources shouldn't be dashable, or walkable by monster if the option hasn't been check

-Monster shouldn't be allowed to respawn on ressources (Depleted or not).

**Screenshots**
Not sure i could help but i tried to take screen of what happen : 
https://ascensiongamedev.com/resources/filehost/7ae706e9138a829a7bd07dc7cb6d805d.png

**Operating System:**
Windows

Cheshire92:
resolves #388

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -424,8 +424,7 @@ namespace Intersect.Server.Entities
                         else if (en is Resource resource)
                         {
                             //If determine if we should walk
-                            if (!resource.IsDead() && !resource.Base.WalkableBefore ||
-                                resource.IsDead() && !resource.Base.WalkableAfter)
+                            if (!resource.IsPassable())
                             {
                                 return (int) EntityTypes.Resource;
                             }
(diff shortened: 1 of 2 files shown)


Cheshire92:
Confirmed working here:  https://github.com/AscensionGameDev/Intersect-Engine/issue_comments/707929431_

jcsnider (approved):

PR #427, opened 2020-11-14
Refactor asset system

lodicolo:


Cheshire92:
Okay, so I can't replicate the insane amount of VRAM the original PR would consume. That's good.

However, this seems to destroy every JSON UI definition and requires me to go into every single file to change the asset names manually. Is that intentional?

lodicolo:
This is no longer going into B7.

PR #428, opened 2020-11-19
feat(logging-db): add logging db with user action history

lodicolo:
Note that this isn't _totally_ complete, as not all of the user actions are represented in the DB, and some things like Quit to Desktop do not get reported.

Will open a separate issue (or issues) to expand upon it, but this was meant more to be a framework base and example for other logging features to get added.


Cheshire92 (approved):
Looks good to me, and I am definitely not being told so.
👀

PR #431, opened 2020-11-28
fix(building): fix build on linux using msbuild/monodevelop

### LINKED ISSUE #432: bug: cross platform build broken on mono (linux/mac)
**Describe the bug**
Building without Windows VS's MSBuild results in an incomplete/broken binary if the dependency file name matches `*.so.*`, rendering any client built using Mono broken on Linux.

**Expected behavior**
Build should include expected embedded resources in the client binary.

**Build System**
Mono

lodicolo:
Fixes #432 

image


Cheshire92 (approved):
Tested this on Windows just to make sure it still builds there as well, and seems to work fine.
Everything shows up in the correct folders and binaries appear to be packed appropriately. (ie, it runs outside of the build folders without any additional binaries copied)

PR #493, opened 2021-01-30
Reduce packet sizes 

### LINKED ISSUE #492: MessagePack Typeless Serializer Producing Large Packets
**Describe the bug**
We need to keep packet sizes down so that in busy games we are not effectively DDOS'ing users with poor connections due to larger than required packet sizes.

MessagePack's typeless serializer appends the full packet type with every transmission such as the following
```"Intersect.Network.Packets.Client.AdminActionPacket, Intersect Core, Version=0.7.0.0, Culture=neutral, PublicKeyToken=null"```

**Version Encountered**
0.7.0.101

**To Reproduce**
Boot up the client and server and monitor packet sizes, compare to the old ceras serialization we were using.

**Expected behavior**
Smaller packets, instead of sending the full type of each packet, send an integer or a key representing that type.

**Operating System:**
All

### PR DESCRIPTION
**Pull request by jcsnider**
By wrapping packets into a class that saves them as an integer.

Intead of prefixing every packet with a full type name such as the one here:
"Intersect.Network.Packets.Client.AdminActionPacket, Intersect Core, Version=0.7.0.0, Culture=neutral, PublicKeyToken=null"

Fixes #492

### DIFF
--- a/Intersect (Core)/Network/PackedIntersectPacket.cs
+++ b/Intersect (Core)/Network/PackedIntersectPacket.cs
@@ -1,0 +1,60 @@
+﻿using MessagePack;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Intersect.Network
+{
+    [MessagePackObject]
+    public class PackedIntersectPacket
+    {
+        private static readonly string[] BuiltInPacketNamespaces = new[]
+        {
+            "Intersect.Network.Packets",
+            "Intersect.Network.Packets.Client",
+            "Intersect.Network.Packets.Editor",
+            "Intersect.Network.Packets.Server",
+            "Intersect.Admin.Actions"
+        };
+
+        public static Dictionary<Type, Int32> KnownTypes { get; set; } = new Dictionary<Type, Int32>();
+
+        public static Dictionary<Int32, Type> KnownKeys { get; set; } = new Dictionary<Int32, Type>();
+
+        private static IEnumerable<Type> FindTypes(IEnumerable<string> nameSpaces) => nameSpaces.SelectMany(FindTypes);
+
+        private static IEnumerable<Type> FindTypes(string nameSpace) =>
+            typeof(Ceras).Assembly.GetTypes().Where(type => type.Namespace == nameSpace);
+
+        static PackedIntersectPacket()
+        {
+            var i = 0;
+            foreach (var type in FindTypes(BuiltInPacketNamespaces))
+            {
+                KnownKeys.Add(i, type);
+                KnownTypes.Add(type, i++);
+            }
+        }
+
+        [Key(0)]
+        public Int32 PacketKey { get; set; } = -1;
+
+        [Key(1)]
+        public byte[] PacketData { get; set; }
+
+        [IgnoreMember]
+        public Type PacketType => KnownKeys[PacketKey];
+
+        public PackedIntersectPacket()
+        {
+
+        }
+
+        public PackedIntersectPacket(IntersectPacket pkt)
+        {
+            PacketKey = KnownTypes[pkt.GetType()];
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Intersect (Core)/Network/PackedIntersectPacket.cs [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+        private static readonly string[] BuiltInPacketNamespaces = new[]
+        {
+            "Intersect.Network.Packets",
+            "Intersect.Network.Packets.Client",
+            "Intersect.Network.Packets.Editor",
+            "Intersect.Network.Packets.Server",
+            "Intersect.Admin.Actions"
+        };
+
+        public static Dictionary<Type, Int32> KnownTypes { get; set; } = new Dictionary<Type, Int32>();
```
lodicolo, line 22: `int`

Intersect (Core)/Network/PackedIntersectPacket.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public static Dictionary<Int32, Type> KnownKeys { get; set; } = new Dictionary<Int32, Type>();
```
lodicolo, line 24: `int`

lodicolo (changes requested):
This doesn't hook into the packet type registration mechanisms for plugin support for networking, this needs to be fixed.

Intersect (Core)/Network/PackedIntersectPacket.cs
jcsnider: Meant to do an Int16 ---- changed to just short. (Keeping packet sizes small, assuming this matters for messagepack)

Intersect (Core)/Network/PackedIntersectPacket.cs
jcsnider: Meant to do an Int16 ---- changed to just short. (Keeping packet sizes small, assuming this matters for messagepack)

jcsnider:
Fixed the packet registration for plugins and confirmed that it's working.

The plugins themselves will need some love at some point, the server.framework nuget package is missing, and the packets in there need MessagePack annotations.

The known packets idea you sent on discord would probably be fine, but I want to push that back to later, the benefit right now is slim to none and adds excess traffic.

Intersect (Core)/Network/PackedIntersectPacket.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        static PackedIntersectPacket()
+        {
+            AddKnownTypes(FindTypes(BuiltInPacketNamespaces).ToList());
+        }
+
+        [Key(0)]
+        public short PacketKey { get; set; } = -1;
```
lodicolo, line 46: This is honestly redundant, change to just `Key`, `Data` and `Type`.

lodicolo (approved):

lodicolo (changes requested):
Oops. Approved before but meant to put request changes.

Intersect (Core)/Network/PackedIntersectPacket.cs
```diff
@@ -0,0 +1,64 @@
...
+            short i = (short)KnownKeys.Count;
+            foreach (var type in types.Where(type => !KnownTypes.ContainsKey(type))) {
+                KnownKeys.Add(i, type);
+                KnownTypes.Add(type, i++);
+            }
+        }
+
+        static PackedIntersectPacket()
+        {
+            AddKnownTypes(FindTypes(BuiltInPacketNamespaces).ToList());
```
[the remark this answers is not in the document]
jcsnider: The editor needs this. In the Ceras class this was done as an assignment to the KnownTypes dictionary but since we are populating multiple dictionaries now for looking up both ways to save time this is needed.

The alternative is making multiple functions public and overall more messy.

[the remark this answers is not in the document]
lodicolo: Replied on Discord re: Network Helper in Editor network class.

lodicolo (approved):

PR #498, opened 2021-02-06
Send vital/status updates each map update instead of on demand

### LINKED ISSUE #497: Vital/Status Updates Consume Too Much Network Traffic
**Describe the bug**
Vital/Status updates are sent way too often, consuming way too much network, and can effectively ddos players with poor internet connections.

**Version Encountered**
0.7.0.101

**To Reproduce**
Create a large raid boss, and then have 100 people attack it.

**Operating System:**
Windows, Mac, and Linux

**Additional context**
Needs an update that will send these updates in batches at an interval in which still feels nice but isn't instant.

jcsnider:
Fixes #497 

This code/logic was written by @Cheshire92 for Leafling. (Credit where credit is due.)

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -99,15 +99,15 @@ namespace Intersect.Server.Entities
         [JsonIgnore, Column("Vitals")]
         public string VitalsJson
         {
-            get => DatabaseUtils.SaveIntArray(_vital, (int) Enums.Vitals.VitalCount);
-            set => _vital = DatabaseUtils.LoadIntArray(value, (int) Enums.Vitals.VitalCount);
+            get => DatabaseUtils.SaveIntArray(mVitals, (int) Enums.Vitals.VitalCount);
+            set => mVitals = DatabaseUtils.LoadIntArray(value, (int) Enums.Vitals.VitalCount);
         }
 
         [JsonProperty("Vitals"), NotMapped]
-        public int[] _vital { get; set; } = new int[(int) Enums.Vitals.VitalCount];
+        private int[] mVitals { get; set; } = new int[(int) Enums.Vitals.VitalCount];
 
         [JsonIgnore, NotMapped]
-        public int[] _oldvital { get; set; } = new int[(int)Enums.Vitals.VitalCount];
+        private int[] mOldVitals { get; set; } = new int[(int)Enums.Vitals.VitalCount];
 
         //Stats based on npc settings, class settings, etc for quick calculations
         [JsonIgnore, Column(nameof(BaseStats))]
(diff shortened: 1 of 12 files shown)


Intersect.Server/Maps/MapInstance.cs
```diff
@@ -945,6 +950,40 @@ public void Update(long timeMs)
...
+                        // Send a party update if we're a player with a party.
+                        if (en.Value is Player player)
+                        {
+                            for (var i = 0; i < player.Party.Count; i++)
+                            {
+                                PacketSender.SendPartyUpdateTo(player.Party[i], player);
+                            }
+                        }
+                        
+                        en.Value._oldvital = en.Value.GetVitals();
```
lodicolo, line 968: Can you make it a method called `StoreVitals()` or something like that? It seems really weird to access a variable with this name.

Intersect.Server/Maps/MapInstance.cs
```diff
@@ -945,6 +950,40 @@ public void Update(long timeMs)
                     }
 
                     en.Value.Update(timeMs);
+
+                    // Check to see if we need to send any entity vital and status updates for this entity.
+                    if (!en.Value.GetVitals().SequenceEqual(en.Value._oldvital))
```
lodicolo, line 955: Can you move the check to the entity, something like `HaveVitalsChanged()`?

Intersect.Server/Entities/Entity.cs
```diff
@@ -106,6 +106,9 @@ public string VitalsJson
         [JsonProperty("Vitals"), NotMapped]
         public int[] _vital { get; set; } = new int[(int) Enums.Vitals.VitalCount];
 
+        [JsonIgnore, NotMapped]
+        public int[] _oldvital { get; set; } = new int[(int)Enums.Vitals.VitalCount];
```
lodicolo, line 110: Rule of thumb, we absolutely should not be naming public variables like this. If it was absolutely necessary it should be `internal`. That said I commented a different solution further along in the review.

Intersect.Server/Maps/MapInstance.cs
```diff
@@ -945,6 +950,40 @@ public void Update(long timeMs)
...
+                            for (var i = 0; i < player.Party.Count; i++)
+                            {
+                                PacketSender.SendPartyUpdateTo(player.Party[i], player);
+                            }
+                        }
+                        
+                        en.Value._oldvital = en.Value.GetVitals();
+                    }
+
+                    if (!en.Value.CachedStatuses.SequenceEqual(en.Value.OldCachedStatuses))
```
lodicolo, line 971: The same comments apply to statuses. This check doesn't really belong in `MapInstance`.

Another idea is

```csharp
statusUpdates.AddRange(en.Value.StatusUpdates)
```
and

```csharp
var entityVitalUpdates= en.Value.VitalUpdates;
if (entityVitalUpdates.Count > 0) {
    // party code
    entityVitalUpdates.AddRange(vitalUpdates);
}
```

lodicolo: (And you would do all the checks and updates in those two properties on the entities)

lodicolo (approved):

lodicolo (changes requested):
Oops. Approved instead of requesting changes again...

lodicolo (approved):

PR #504, opened 2021-02-06
Auto unequip items that are not equipment, or items we no longer meet conditions for

### LINKED ISSUE #501: Auto unequip items when conditions are no longer met
If a player changes class via events, they will continue to wear their equipment -- even if that equipment can only be worn by the previous class.

Add code that will check for equipment that should no longer be wearable and remove it
 - Upon Login
 - Upon Class Change
 - Check during player update at a configurable interval?

**Additional context**

jcsnider:
Resolves #501 -- mostly. Might loop back around and do this check at various intervals but this is a good start.

### DIFF
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -821,6 +821,7 @@ namespace Intersect.Server.Entities
             }
 
             RecalculateStatsAndPoints();
+            UnequipInvalidItems();
             PacketSender.SendEntityDataToProximity(this);
             PacketSender.SendExperience(this);
         }
@@ -874,6 +875,7 @@ namespace Intersect.Server.Entities
             }
 
             RecalculateStatsAndPoints();
+            UnequipInvalidItems();
             PacketSender.SendExperience(this);
             PacketSender.SendPointsTo(this);
             PacketSender.SendEntityDataToProximity(this);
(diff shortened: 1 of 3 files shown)


Intersect.Server/Entities/Player.cs
```diff
@@ -4619,6 +4621,32 @@ public void EquipmentProcessItemLoss(int slot)
...
+        /// Unequips any items that the player is currently wearing in which they no longer meet the conditions for
+        /// Also unequips any items that are not actually equipment anymore
+        /// </summary>
+        public void UnequipInvalidItems()
+        {
+            var updated = false;
+            for (int i = 0; i < Options.EquipmentSlots.Count; i++)
+            {
+                var itemSlot = Equipment[i];
+                if (itemSlot > -1)
```
lodicolo, line 4634: There's a... lot to unpack here.

For starters use
```csharp
if (itemSlot < 0) {
    continue;
}
```

Intersect.Server/Entities/Player.cs
```diff
@@ -4619,6 +4621,32 @@ public void EquipmentProcessItemLoss(int slot)
...
+                {
+                    if (Items[itemSlot] == null || Items[itemSlot].ItemId == Guid.Empty || Items[itemSlot].Descriptor == null || Items[itemSlot].Descriptor.ItemType != ItemTypes.Equipment || !Conditions.MeetsConditionLists(Items[itemSlot].Descriptor.UsageRequirements, this, null)) {
```
lodicolo, line 4636: Please break this check up into parts. This is far too long and unreadable.

I suggest something like:
```csharp
var item = Items[itemSlot];
var descriptor = item?.Descriptor;
if (descriptor != default && item.ItemId != Guid.Empty) {
    continue;
}

if (descriptor.ItemType == ItemTypes.Equipment) {
    continue;
}

if (Conditions.MeetsConditionLists(descriptor.UsageRequirements, this, null)) {
    continue;
}

Equipment[i] = -1;
updated = true;
```

lodicolo (changes requested):

lodicolo (approved):

Issue #434, opened 2020-12-10
Delete Character


fugakuar:
**Describe the bug**
The character when attacked does not disappear if it is deleted from the panel.

**Version Encountered**
0.6.2.0

**To Reproduce**
1. Go to fight with any npc
2. Press ESC and go to character while you're in battle
3. Quickly you give to delete character. You create a new one and enter the game
4. The character does not disappear until the NPC stops attacking him, kills him or someone kills the NPC.

**Expected behavior**
My expected behavior is that the character is automatically deleted.

**Screenshots**
https://www.youtube.com/watch?v=kPDXC_UUWe8

**Logs**
N/A

**Operating System:**
Windows

**Additional context**
The fact that the character does not disappear in battle is fine, but what if this character is a level higher than the npc? The character will never disappear. If an NPC is in the spawn zone and the character dies again and again, he will not be eliminated either.

Arufonsu:
>     1. Go to fight with any npc
> 
>     2. Press ESC and go to character while you're in battle

imo, there should be a check so players are not able to log off / swap into the character selection screen during a battle for a specific period of time, lets say, 1 minute, this way no one could abuse this.


> **Expected behavior**
> My expected behavior is that the character is automatically deleted.

if the check that i'm suggesting its ever implemented, an "already logged in" check should also be implemented along with this, that way, if the account holder tries to delete the character while this character its still in game/battle and unable to log off, the account holder should receive a "you can't delete a character that's already logged in" warning notice instead of simply delete it...

PR #553, opened 2021-02-17
Fix server crashes due to deleting characters that remain online and in combat

### LINKED ISSUE #543: System of keeping character online in combat is broken.
**Describe the bug**
NOTE: The system works normal, I just haven't found a better title, if you want to change it to make it better for you, feel free.

The main bug here is that if you create a character, go into combat mode. Exit the game and delete the created character and log in with another character. The deleted character will still be there! If you kill the monster that is hitting him, after the combat time expires the **server will crash.**

As if that were not enough, after starting the server again, the character that killed the deleted player's monster will always remain buggy. Combat messages of logut in combat will not appear for him if you try to leave the game. If you try to select a character you go to the home screen. If you try to enter the game with the character you will receive the message "Failed to load account. Please try logging in again"

And don't stop there, that same character will have the ability to bug all the npcs that he enters in combat, if he enters combat and exits the game, the npcs will remain aggressive attacking nothing, stopped in the game (until another player attack it).

**Version Encountered**
0.7.0.106

**To Reproduce**
1. Create a character, go into combat mode
2. Exit the game and delete the created character and log in with another character
3. Kill the monster that is hitting him and wait combat time finish.

**Expected behavior**
Preferably do not crash.

**Screenshots**
https://www.ascensiongamedev.com/resources/filehost/db1e7e56fb8b05c351e6848be15f9331.mkv

**Logs**
[Intersect Server-2021_02_13-18_38_32_399.log](https://github.com/AscensionGameDev/Intersect-Engine/files/5976576/Intersect.Server-2021_02_13-18_38_32_399.log)
[Intersect Server-2021_02_13-18_36_25_674.log](https://github.com/AscensionGameDev/Intersect-Engine/files/5976577/Intersect.Server-2021_02_13-18_36_25_674.log)
[errors-Intersect Server.log](https://github.com/AscensionGameDev/Intersect-Engine/files/5976578/errors-Intersect.Server.log)
[playerdbIntersect Server-2021_02_13-18_36_25_674.log](https://github.com/AscensionGameDev/Intersect-Engine/files/5976579/playerdbIntersect.Server-2021_02_13-18_36_25_674.log)

**Operating System:**
Windows 10

**Additional context**
This system is a good addition to the intersect, because in pvp events, the player will not simply leave easily, among many other situations. But this system must somehow have limits.
1 - I don't really know how it should be done (I'm not a programmer and I'm not even close to being one), but there must be a condition for when the character is deleted and he is online in the game. (to avoid the above crash.)
2 - There should be a configuration on the server that would allow us to configure a time for these disconnected users, but that the characters are kept for being in combat. They shouldn't be infinitely online. Look at what I did and what the players could do.

I opened two clients, logged into two different accounts. In account 1 I left the party leader. In account 2 I created two characters in the same account and left these characters online through the combat system (two online characters from the same account is already a mistake I would say). I put them in a party and closed client 2. So thanks to this system, I can up level of several characters from the same account simultaneously because they will never disconnect unless someone shows up and make them go out of combat.

https://www.ascensiongamedev.com/resources/filehost/c90a26d61e99c366c01508616d70f65e.mkv

3 - Other situations that I did not mention above, why I did not explore further.

jcsnider:
Resolves #543 and #434

### DIFF
--- a/Intersect.Server/Networking/Client.cs
+++ b/Intersect.Server/Networking/Client.cs
@@ -212,15 +212,8 @@ namespace Intersect.Server.Networking
         public void Logout(bool force = false)
         {
             var entity = Entity;
-            if (entity != null)
-            {
-                entity.LastOnline = DateTime.Now;
-
-                entity.TryLogout(force);
-
-                entity.Client = null;
-                Entity = null;
-            }
+            entity?.TryLogout();
+            Entity = null;
 
             if (!force)
             {
--- a/Intersect.Server/Networking/Client.cs
+++ b/Intersect.Server/Networking/Client.cs
@@ -127,7 +127,7 @@ namespace Intersect.Server.Networking
         {
             if (user == null && User != null)
             {
-                User.Logout(User);
+                User.TryLogout();
             }
 
             if (user != null && user != User)
--- a/Intersect.Server/Networking/Client.cs
+++ b/Intersect.Server/Networking/Client.cs
@@ -125,9 +125,9 @@ namespace Intersect.Server.Networking
 
         public void SetUser(User user)
         {
-            if (user == null && User != null)
+            if (user == null)
             {
-                User.TryLogout();
+                User?.TryLogout();
             }
 
             if (user != null && user != User)
--- a/Intersect.Client/Interface/Game/EscapeMenu.cs
+++ b/Intersect.Client/Interface/Game/EscapeMenu.cs
@@ -114,6 +114,8 @@ namespace Intersect.Client.Interface.Game
             {
                 BringToFront();
             }
+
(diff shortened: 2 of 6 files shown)


Intersect.Server/Networking/Client.cs
```diff
@@ -127,7 +127,7 @@ public void SetUser(User user)
         {
             if (user == null && User != null)
             {
-                User.Logout(User);
+                User.TryLogout();
```
lodicolo, line 130: Remove `User != null` and do `User?.TryLogout()`

lodicolo (changes requested):

lodicolo (approved):

PR #559, opened 2021-02-18
Fix invulnerability by only blocking damage (instead of damage and healing)

### LINKED ISSUE #433: Invulnerability effect : Cannot Heal or Gain mana from spell
**Describe the bug**
When under the effect of Invulnerability, players cannot regain mana or HP from other friendly spell or self spell but still can gain from consumables.

**Version Encountered**
0.6.2.448

**To Reproduce**
1. Make a self spell wich heal you every 1 sec and make you invulnerable that last 10 sec.
2. Get Hurt
3. Use Spell
5. Make a consumable that give you back Hp
6. Get Hurt
7. Use spell then use Consumables.
8. See Error

**Expected behavior**
Spells that heal and give mana should be effective when used under invulnerability. (Cause Consumables Are)

**Operating System:**
Windows

jcsnider:
Resolves #433

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -1708,23 +1708,7 @@ namespace Intersect.Server.Entities
                 return;
             }
 
-            //Check for enemy statuses
-            foreach (var status in enemy.CachedStatuses)
-            {
-                //Invulnerability ignore
-                if (status.Type == StatusTypes.Invulnerable)
-                {
-                    PacketSender.SendActionMsg(enemy, Strings.Combat.invulnerable, CustomColors.Combat.Invulnerable);
-
-                    // Add a timer before able to make the next move.
-                    if (this is Npc npc)
-                    {
-                        npc.MoveTimer = Globals.Timing.Milliseconds + (long) GetMovementTime();
-                    }
-
-                    return;
-                }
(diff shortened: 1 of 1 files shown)


Intersect.Server/Entities/Entity.cs
```diff
@@ -1708,23 +1708,7 @@ public virtual void TryAttack(Entity target)
                 return;
             }
 
-            //Check for enemy statuses
-            foreach (var status in enemy.CachedStatuses)
-            {
-                //Invulnerability ignore
-                if (status.Type == StatusTypes.Invulnerable)
-                {
-                    PacketSender.SendActionMsg(enemy, Strings.Combat.invulnerable, CustomColors.Combat.Invulnerable);
```
Cheshire92, line 1717: You're losing this message, but not sending it anywhere down the line?
The client will never know the entity they're hitting went invulnerable.

jcsnider: It should show up as an action message when the Invulnerability takes effect/is casted -- and invulnerability would be a relatively short buff due to its nature. If a player/mob/whatever is getting hit by 100 spells/second we don't want to send the invulnerable action message hundreds of times.

lodicolo (approved):

PR #561, opened 2021-02-19
Updating prerelease with fixes from development

jcsnider:


jcsnider:
Don't merge this yet until I investigate issues with the combat relogging that Joyce encountered.

PR #584, opened 2021-02-27
[Tags] +Player Event Support +Npc Editor Support and more.

### LINKED ISSUE #580: List of entities linked to a list of CustomTags.
Seeing the functionality of the new PR https://github.com/AscensionGameDev/Intersect-Engine/pull/577 for the tags there is a small problem of redundancy of files on the client side, where each of the entities that use the same tag would need to have a repeated PNG file, this in medium-large projects is a problem to maintain tags updated in case of changes.

The ideal would be to have a list of names of the customTags (in case you want to use them) that would look like this: 

`customTags: ['GM', 'Donor', 'VIP']`

`customTagsPlayers: [
{player: 'Blinkuz', tagName: 'GM'},
{player:'Arufonsu': tagName: 'VIP'}
 ]`

This list would be for each type of tags, that is, for NPCS a list and another for Players in case you want to have everything organized

Then in the search for tags it would be to look for: GM_tag.png from the list of tags in the resources folder and assign it to the Blinkuz player, this allows the same png file to be reused in multiple players.

**NOTE:** The lists should be as generic as possible due to the variety of uses that can be given to them, I do not recommend making GMCustomTags, or VIPCustomTags lists, with only having 1 list for each type of entity that would be npc and player should be flexible enough to give it different uses, an example of npcs would be labels for BOSS and for ELITES enemies, where the same labels can be reused for multiple enemies.

Arufonsu:
Resolves #580 

### PR - Summary

- Adds the "Tag" column to Players and Npcs database tables. - _Thanks @Blinkuz for helping with this PR, none of this would be possible without his help when it came to migrate the database changes for Npcs and Players._
- UI Changes for the Npc Editor - _Tags can easily be assigned to Npcs now! 
(Properly **Set** and **Get** the information from the Npc database)._ 
This way, there's no more redundancy when it comes to set the same custom tag to a group of Npcs.
Default Npc Tags are still available and you may disable them if you wish with the ShowCustomTagsOnly option.
2021-02-26 18_12_44-Npc Editor 
_Also, it renders the selected tag sprite._

- New Event Editor Command - Player Tags are now powered by events!
Sprite-0002

- Set, Remove (Set to None), Change, do as you please ;3
tagevent

- Worried about Labels? don't worry mate!
labels

### Extras

- Added a "None" dropdown option for Change Sprite and Change Face forms in the event editor. Selecting this will "remove" (set _null_) the player's sprite and face without any issue, this feature may be handy for developers at some point.
Sprite-0003 Sprite-0004


Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
```diff
@@ -12,27 +12,27 @@ public EventEntityPacket()
         }
 
 
-        [Key(24)]
+        [Key(25)]
```
Cheshire92, line 15: Why is this file touched?

Intersect (Core)/Network/Packets/Server/NpcEntityPacket.cs
```diff
@@ -11,7 +11,7 @@ public NpcEntityPacket()
         }
 
 
-        [Key(24)]
+        [Key(25)]
```
Cheshire92, line 14: Same with this file..

Intersect (Core)/Network/Packets/Server/PlayerEntityPacket.cs
```diff
@@ -14,23 +13,23 @@ public PlayerEntityPacket()
         }
 
 
-        [Key(24)]
+        [Key(25)]
```
Cheshire92, line 16: And this one

Intersect (Core)/Network/Packets/Server/ProjectileEntityPacket.cs
```diff
@@ -11,19 +11,19 @@ public ProjectileEntityPacket()
         {
         }
 
-        [Key(24)]
+        [Key(25)]
```
Cheshire92, line 14: And this one

Intersect (Core)/Network/Packets/Server/ResourceEntityPacket.cs
```diff
@@ -12,11 +12,11 @@ public ResourceEntityPacket()
         }
 
 
-        [Key(24)]
+        [Key(25)]
```
Cheshire92, line 15: And this one

README.md
```diff
@@ -58,14 +58,14 @@ If you'd like to help the project by supporting us financially we accept and gre
 
 ## Source Code
 
-Getting started with Intersect couldn't be easier. Make sure you have Visual Studio 2017 Community (or later) installed. 
+Getting started with Intersect couldn't be easier. Make sure you have Visual Studio 2019 Community installed.
```
Cheshire92, line 61: Should this really be in this pr?

Cheshire92 (commented):
Not looked at the full thing and functionality but just some quick stuff I saw scrolling through.

Also is this tested to work with asset packing?

Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
blinkuz: We had problems keeping the same key that was in the entityPacket, and as we had to add one more property to the EntityPacket.cs it showed us an error in the other files that were being made to join, what we saw more logical was to readjust the keys to match the order and it worked.

Intersect (Core)/Network/Packets/Server/NpcEntityPacket.cs
blinkuz: > Not looked at the full thing and functionality but just some quick stuff I saw scrolling through.
> 
> Also is this tested to work with asset packing?

We did not test it, I will do a test and respond again if it works or not with the packer.

Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
Arufonsu: Since we needed an extra key for tags in both ends (Players and Npcs), we decided to add it as the 24th key on EntityPacket.cs. Is because of this that we had to change the keys of the Message Pack Objects in order to avoid a duplicated 24th key in the Entity packet structure.

Intersect (Core)/Network/Packets/Server/NpcEntityPacket.cs
Arufonsu: Since we needed an extra key for tags in both ends (Players and Npcs), we decided to add it as the 24th key on EntityPacket.cs. Is because of this that we had to change the keys of the Message Pack Objects in order to avoid a duplicated 24th key in the Entity packet structure.

Intersect (Core)/Network/Packets/Server/PlayerEntityPacket.cs
Arufonsu: Since we needed an extra key for tags in both ends (Players and Npcs), we decided to add it as the 24th key on EntityPacket.cs. Is because of this that we had to change the keys of the Message Pack Objects in order to avoid a duplicated 24th key in the Entity packet structure.

Intersect (Core)/Network/Packets/Server/ProjectileEntityPacket.cs
Arufonsu: Since we needed an extra key for tags in both ends (Players and Npcs), we decided to add it as the 24th key on EntityPacket.cs. Is because of this that we had to change the keys of the Message Pack Objects in order to avoid a duplicated 24th key in the Entity packet structure.

Intersect (Core)/Network/Packets/Server/ResourceEntityPacket.cs
Arufonsu: Since we needed an extra key for tags in both ends (Players and Npcs), we decided to add it as the 24th key on EntityPacket.cs. Is because of this that we had to change the keys of the Message Pack Objects in order to avoid a duplicated 24th key in the Entity packet structure.

README.md
Arufonsu: This was previously approved and merged into the main branch by @lodicolo . I don't see why this shouldn't be in the Dev branch for now either, plus to this, the authors have been updated as well, since @blinkuz helped along with thir PR.

Intersect (Core)/Network/Packets/Server/NpcEntityPacket.cs
Arufonsu: > 
> Also is this tested to work with asset packing?

Tested: this implementation has no issues with asset packing.

Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
Cheshire92: I don't consider this a valid solution as it's random and unclear in the future.

README.md
Cheshire92: Should be its own pr though,. entirely unrelated to this feature addition.

Arufonsu: 👍🏽 Reversed this commit from this PR, 
and opened a new PR with this change, as suggested by the contributor @Cheshire92   - >  #585

Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
Arufonsu: > 
> 
> I don't consider this a valid solution as it's random and unclear in the future.
> 
> (Besides it'll be undone if anyone ever auto generates the keys again)

Hello! [this](https://docs.freemmorpgmaker.com/en-US/developer/advanced/packets.html#sending-packets) is the only documentation i could find regardless packets on this source. If you don't consider this to be a valid solution, perhaps a constructive comment as to how this packet key implementation should be done would be quite helpful for this PR, fellow contributor 😄.

If you can help on this, whichever change is required will be commited then.

Also, i can't find any documentation regardless the auto generation of these keys in the source, it would be nice to add such important information to the [developer docs](https://docs.freemmorpgmaker.com/en-US/developer/) as well, since i'm probably not the first one that seeks to learn from this and contribute at some point as well.

Cheshire92: I believe I mentioned it somewhere although that may have been somewhere else that I'd personally put it on the entity class so it would be shared between anything inheriting from that class as well, solving this problem.

Would also let you deduplicate the draw tags method from the NPC and player classes and simply have a single shared one on entities.

Arufonsu: The draw tags method its present for every entity in the client's Entities/Entity.cs, currently coded thinking on Npcs, and there's an override for Players, as suggested by @lodicolo in the client's Entities/Player.cs. Having an override is specially useful in this case in order to avoid conflicts beetwen Npcs and Players. This is one of the reasons for adding the 24th key in the EntityPacket.cs. Now if there's a better way to grab the information from the Tags columns from both, Npcs and Players database, we would love to see an example so we can work on the next commit in order to satisfy what you are requesting in this review.

Cheshire92: If the data is already there then why not just send it if the packets expect it to be there as well? Makes it easier to expand on later and doesn't leave random gaps that are confusing and can cause future issues if anyone ever hits the auto generate button again.

Arufonsu: Hello :3 i *think* we got what you meant, recently commited few changes, now the 24th key is being referenced from entities only, at both, client and server side, i simply followed as example how the character sprites and face sprites are handled. If you could enlight us with the knowledge of this auto generate button in order to test if everything works as intended without any conflict when auto generating the keys, that would be awesome, but sadly and like i mentioned 2 comments ago, there's no such documentation on the topic.

Arufonsu:
> 
> Also is this tested to work with asset packing?

Tested: this implementation has no issues with asset packing.

Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
Cheshire92: It's an external library, not sure why we would write up documentation for something that already has it of its own. https://github.com/neuecc/MessagePack-CSharp

Arufonsu: > It's an external library, not sure why we would write up documentation for something that already has it of its own.
> 
Did i ever used the words "write up", oh yeah, i didn't :) every kind proyect should always have this kind of documentation at least referenced within its own, specially if we are talking about external libraries, it's not like we had a way to guess this if is not explicit within this source documentation, do we? thanks for the documentation link anyway, we should think about adding these to an external libraries documentation list in the source docs in order to help futher developers and collaborators that will eventually join this open source project.

> And reading your commit you're still skipping the number in the packet definition.
After we read the proper documentation, we are probably going to find out what you are trying to say here.

PS: We are still waiting to see a clear example of what a "valid solution" should look like from the reviewer, so far, only questions and criticism with a spice of negativity towards us for reasons that i still don't understand. hopefully, after reading this documentation you just sent (which is probably the only helpful hint in this review so far) we will make it to fullfit your expectations as a fellow collaborator that decided to review this.

Have a nice day, will take a look on this after work.

blinkuz: Finally something that we can use to better understand the issue of packages, we had no idea that an external library was used, as I have read in the documentation they mention the following:

https://github.com/neuecc/MessagePack-CSharp#union

> Please be mindful that you cannot reuse the same keys in derived types that are already present in the parent type, as internally a single flat array or map will be used and thus cannot have duplicate indexes/keys.

So I ask you, what is the correct way to have modified the main keys of the packages that derive from EntityPacket if it marks the error mentioned in the official documentation that the same key index cannot be repeated in the inheriting classes? I have investigated on the autogeneration of keys and I have not found documentation on the matter so if you can be so kind to indicate where we can carry out the tests of "autogenerate" the keys we would be very grateful.

Cheshire92: I believe there's some hostility being added in translation here which is unintended from my side. But if that's how we're going to double team someone I'm done here.

Nevertheless, I've given you the solution before:

Put the tags in the entity class so it's shared across everything and used in all the packets. That way there are no skipped numbers. They can be used for just about every entity derivative in the future should the desire be there anyway so why not.

Arufonsu: > I believe there's some hostility being added in translation here which is unintended from my side.
> 

My bad if it was unintended, but the words used felt pretty hostile towards us, plus to that the lack of a clear example of what a valid solution should be like, we are supposed to be on this together as collaborators, i'm not trying to test how skilled you are, neither should you, the link to the documentation would been an expected feedback few days ago, but instead we simply got "this makes no sense", "this is not valid", etc. if you are going to point out something as "wrong", fine, then show us the proper way to do it, we want to **learn about this**, and everyone reading this would be happy to learn for sure too.


> Nevertheless, I've given you the solution before:
> 
> Put the tags in the entity class so it's shared across everything and used in all the packets. That way there are no skipped numbers. They can be used for just about every entity derivative in the future should the desire be there anyway so why not.

That's already been done that way, the tags packets are only referenced from the entity class only (client and server), so they are available for every entity, what we have been trying to understand is this issue you came up with, related with the auto generator of the packet keys, and we just got the documentation for it **today**, we will try (despite of this situation that's probably nothing but a misunderstood due to some cultural differences to what "hostility" looks like in text), to solve this.

Have a nice day, if your hostility was really unintended, it would be nice to keep doing things together in the future, because here, we really want to **learn** and **help**, and surely more people will show up around this git with the very same desire to collaborate and learn at the same time.

Cheshire92: I'm writing these messages on my phone from memory, so they might not be the most detailed. A lot of the messagepack stuff is also relatively new to the engine as not too long ago it was all handled by Ceras. So aside from my experience with a live game working with it I'm not too deeply familiar with the library myself yet.

As for an example, I think you're misunderstanding which entity classes I'm talking about. On the server all entities are based off of a singular class, Entity.cs.

If tags are stored there they can be used by and sent for every entity type that exists. Meaning we no longer keep random holes in our packet property numbering in MessagePack.

This way the system could also easily be expanded to resources and events in the future since the fields and data sent to the client already exist.

Arufonsu: > 
> If tags are stored there they can be used by and sent for every entity type that exists. Meaning we no longer keep random holes in our packet property numbering in MessagePack. Right now you're saving them on player and NPC's only if I'm correct?
> 
> This way the system could also easily be expanded to resources and events in the future since the fields and data sent to the client already exist.


Thanks! will make sure to do this and test it asap once i get home today.

blinkuz: > If tags are stored there they can be used by and sent for every entity type that exists. Meaning we no longer keep random holes in our packet property numbering in MessagePack. Right now you're saving them on player and NPC's only if I'm correct?
> 
> This way the system could also easily be expanded to resources and events in the future since the fields and data sent to the client already exist.

I'm working on this and I just saw your answer, about that we have it only in player and npc the answer is no, currently we have it as a property in the entity.cs of the server which is where you are suggesting it, that's why It has been made easier for us to integrate the event system and reuse the same property in both cases for both npc and player.

Server: https://github.com/AscensionGameDev/Intersect-Engine/blob/4bba6695324d079c038b57cd7f173d67fe2a15db/Intersect.Server/Entities/Entity.cs#L80

Client: https://github.com/AscensionGameDev/Intersect-Engine/blob/4bba6695324d079c038b57cd7f173d67fe2a15db/Intersect.Client/Entities/Entity.cs#L85

With this I think your question would be answered? or is there some other point that we should take into account?

Arufonsu: https://github.com/AscensionGameDev/Intersect-Engine/blob/4bba6695324d079c038b57cd7f173d67fe2a15db/Intersect.Server/Entities/Entity.cs#L2667

https://github.com/AscensionGameDev/Intersect-Engine/blob/4bba6695324d079c038b57cd7f173d67fe2a15db/Intersect.Client/Entities/Entity.cs#L306

These Lines should be considered too, Npc's tag is referenced from the main Entity class, while the same happens for Player's override. Thing is, how can we do this without a key in the entity packet MessagePack chain? Not sure if i understand correctly, but this seems what is Ches has been asking us to do in the review. Also, i did took a look over the documentation for MessagePack, and couldn't find the auto generate topic on it sadly u_u' no sure what i'm missing here >.< !

jcsnider:
I want to keep you in the loop while working out stuff on our end. I don't think the original Tags PR should have been accepted in that state, and this PR is a much better WIP.

While waiting on a resolution I've done a whole lot of development on prerelease which we need to also bring over into development and then return to the development > prerelease > main update chain.

For the time being I think we're going to force push prerelease into development which will temporarily wipe out the original tags, and then hopefully continue working with you all (maybe in a clean PR -- not sure if that will be necessary) to bring these back into into the fold.

Arufonsu:
> 
> 
> I want to keep you in the loop while working out stuff on our end. I don't think the original Tags PR should have been accepted in that state, and this PR is a much better WIP.
> 
indeed, i do agree on this one.

> While waiting on a resolution I've done a whole lot of development on prerelease which we need to also bring over into development and then return to the development > prerelease > main update chain.
> 
> For the time being I think we're going to force push prerelease into development which will temporarily wipe out the original tags, and then hopefully continue working with you all (maybe in a clean PR -- not sure if that will be necessary) to bring these back into into the fold.
Sure! also, starting a new PR for TAGS from scratch won't be a problem (plus, i knew that bug fixing was more likely the priority over features, hence why i decided to leave it as draft for now). At first, this was intended to work as "monster labels", and somehow it went down to the road of the graphic .png tags. Yes, this WIP is way more complete than the first attempt of implementation, plus im not an expert in C# and i did this in order to learn and hopefully help with the project. hopefully this WIP helps and can be improved as others collaborate, just let me know when, what's needed, what could be changed or done differently, and i will do my best!

Have a nice day JC ~ !

jcsnider:
Thanks for your understanding. I just updated the dev branch. This PR will need to be recreated unfortunately. So I will close this one for now.

Issue #589, opened 2021-03-03
Global EXP Modifier


Bioblaze:
**Is your feature request related to a problem? Please describe.**
Global EXP Modifier for a Server

**Describe the solution you'd like**
Add Supported Functionality for a Global EXP Modifier
[https://www.ascensiongamedev.com/topic/5808-beta-6-global-exp-modifierw-console-support/](https://www.ascensiongamedev.com/topic/5808-beta-6-global-exp-modifierw-console-support/)

**Additional context**

blinkuz:
It is a very good utility, and it can be used for many types of games.

Bioblaze:
I made a PR about this Topic: https://github.com/AscensionGameDev/Intersect-Engine/pull/592

PR #591, opened 2021-03-03
Globalexpmodifier

Bioblaze:
Most MMO Servers have a Global EXP Modifier, so when they make a new Server they can simply just change the modifier. I provide that here.

https://github.com/AscensionGameDev/Intersect-Engine/issues/589

https://www.ascensiongamedev.com/topic/5808-beta-6-global-exp-modifierw-console-support/

### DIFF
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -31,6 +31,9 @@ namespace Intersect
         [JsonProperty("OpenPortChecker", Order = 0)]
         protected bool _portChecker = true;
 
+        [JsonProperty("GlobalEXPModifier", Order = -4)]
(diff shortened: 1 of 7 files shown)


Khaikaa:
I dont know if that would be the correct way of doing that. Adding it to the server configuration means that you have to restart the server each time you want to change the exp modificator. It is actually possible to make a dynamic exp system by giving exp through onkill events + a global variable with the modification value you want, and it will work without restarting the server. If intersect would give the possibility to add an exp modifier to work directly with base exp system that may be stored in another way, such as in the database, instead of the config file. That's just my opinion, not saying I'm right, but I think adding features that need to shut down the server to be changed has to be avoid as much as possible

Issue #600, opened 2021-03-15
Bag does'nt work after a reconnect


lucas-tardivo:
**Describe the bug**
The bags can opened but any interaction does not work after a reconnect, Basically the HasBag function cannot find the respective bag because the Bag attribute is being loaded as null, so the function returns any interaction.

**Version Encountered**
v0.7.0.0

**To Reproduce**
1. Get a bag item
2. Open it
3. See it works well
4. Close the client and log again
5. Open the bag, any interaction does not work

**Expected behavior**
The interactions works well after reconnect

**Operating System:**
Windows

Arufonsu:
>     5. Open the bag, any interaction does not work
> 
> 
> **Expected behavior**
> The interactions works well after reconnect

I couldn't reproduce this. besides, it seems like by default you can't walk nor interact with anything while a bag its open.
You can walk and interact around if you close the bag, reconnecting isn't necessary.

lucas-tardivo:
Sorry for did'nt specify well @Arufonsu , the interactions i mean is to put/get items on the respective bag

Magehernan:
hi, i found a bug with this, i don't know if is like you describe about after reconnect, to me happens always, i already have the bag in the player inventory.

I debug it and i found that when checking comparing to check if the player HasBag

https://github.com/AscensionGameDev/Intersect-Engine/blob/9ac502785d1584917ab584ddb365d4f87727458e/Intersect.Server/Entities/Player.cs#L3164-L3175

the bag inside the items is null because when you open the bag... 

https://github.com/AscensionGameDev/Intersect-Engine/blob/9ac502785d1584917ab584ddb365d4f87727458e/Intersect.Server/Entities/Player.cs#L3131-L3163

In the line 3154 it assign the new bag to the item only if the bag is new... so because this the bag in the item is null.

I look back and found this commit 1a8edc5 - fix 191: Bag access is done via cached value (#219)
and in here the code is this
```csharp
 //Bag
        public bool OpenBag(Item bagItem, ItemBase itemBase)
        {
            if (IsBusy())
            {
                return false;
            }

            //Bags will never, ever, be stackable. Going to use the value property for the bag id in the Database.
            if (bagItem.Bag == null)
            {
                bagItem.Bag = DbInterface.GetBag(bagItem);
                if (bagItem.Bag == null) //Bag doesnt exist, create it!
                {
                    //Create the Bag
                    var slotCount = itemBase.SlotCount;
[...]

PR #608, opened 2021-03-23
Fix bug for using bag, trying to store or retrive items form a bag don't work

Magehernan:
it is related but not sure if fix all the problems  for this #600


jcsnider (approved):

jcsnider:
Most of the bag logic should be refactored, but this is a good fix for now. Thanks!

PR #611, opened 2021-03-25
messagepack AOT compatibility for tool to generate code

Magehernan:
Hi, i am working on using the AOT feature for my client in unity and to make it work there was needed to change the virtual property Action in AdminAction because it generate 2 properties with the same name on the resolvers for AOT so i change it to not use override and pass it the value in the base constructor and it works.

Other problem was that IntersectPacket needs to have the Union Attribute with all child classes so i add it.

Hope it helps

Please let me know if i can make something different to my pull request, i am really new on open souce collab.


blinkuz:
The key in the package in theory if it should be done manually, I was investigating about it for another PR that I was collaborating with and we did not find any way to autogenerate them so I do not think that I have done that part wrong.

Magehernan:
@blinkuz Hi, to auto genereate the code for AOT MessagePack has a proyect named MessagePack.Generator, here is the documentation to use it https://github.com/neuecc/MessagePack-CSharp#aot-code-generation-support-for-unityxamarin
When i use it to generate the code for Intersect (Core) project folder, you received this message
```
Project Compilation Start:CodeGenTemp
Project Compilation Complete:00:00:02.3629061
Method Collect Start
Fail in console app running on MessagepackCompiler.RunAsync
MessagePackCompiler.CodeAnalysis.MessagePackGeneratorResolveFailedException: Serialization Type must mark UnionAttribute. type: global::Intersect.Network.IntersectPacket
```
So for the IntersectPacket was necessary for mpc to explisit the Union Attribute for all the child of it.

Then i make all the changes to generete de file for all resolvers and formatters for the packages, the problem with the overrided property Action was it generate 2 properties with the same name on the formatters and resolvers and it fails so to not fix the generated code i make that change so we don't have override on AdminAction and it generate the code without errors...

now with this changes you can use mpc to generate the code and it works fine, i compile with unity and il2cpp and it works.

Sorry if my english is bad, i am from argentina and i'm not really good at writing english.

Arufonsu:
hey, this is amazing, so if what you are explaining here is right, then we will be able to use MPC auto generator for packet's keys  without any issue right? hopefully this can can be reviewed by other devs, since this would be helpful for a PR that i've been working on along with @blinkuz

PR #644, opened 2021-04-24
Add Discord Rich Integration

Unitedbr:
Adds Discord rich integration to update the user's discord status with details about the game and character.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -13,4 +13,6 @@
 
 [Dog from the Kash Shop](#)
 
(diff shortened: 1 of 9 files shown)


jcsnider:
While I love the idea I'm not sure if we want to add specific integrations like this directly into the engine. Seems like something that would be more appropriate for the plugin system.. which needs some updating of the examples but could probably handle this already.

Unitedbr:
Where can I find more information about plugins?

Issue #464, opened 2021-01-23
Change items enhancement : Drop Item


Spectreye:
**Is your feature request related to a problem? Please describe.**
Actually when people die in pvp or in some situation we can only have people drop random items including costy gears and such.
Also it would add a lot of possibilities to let people drop stuff instead of just "taking" it.

**Describe the solution you'd like**
Add the action "Drop", to the Event "Change Player Items"
It would make the chosen items drop to the ground.

**Additional context**
I think it would benefit a lot of games types that include a lot of PvP, but not only.

Issue #243, opened 2020-05-30
 rare item for exchange.


Moscano:
**Is your feature request related to a problem? Please describe.**

https://prnt.sc/sqnggo   
There could be another checkbox, a "rare" item that you can exchange with other players.
but the item could not fall on the floor.
Like an item purchased with real money, being able to trade with players.

**Describe the solution you'd like**

**Additional context**

WeylonSantana:
I think the term you meant is "tradable" when "bound" is checked. That would be a good idea, yes. It would have items linked to the player and items linked, but that could be traded.

Azurebeats:
Maybe this would work if we tore apart the bound option into several smaller settings. For example a droppable and tradable checkbox.

Maybe add another that marks an item as unique so a user can never have more than one of it.

PR #720, opened 2021-05-12
Make item binding options more granular

### LINKED ISSUE #670: Something wrong with the chance to drop items when you die
**Describe the bug**
I put 1 in the drop configuration when dying, and it seems that the percentage is very high, the whole inventory almost drops all when dying, if you put 0, nothing drops, if you put 1, half of the inventory drops.

**Version Encountered**
0.7.0.115

**To Reproduce**
1 - Set the drop in the server settings, to number 1,
2 - fill your inventory, and see almost all of it dropping.

**Expected behavior**
should 1 really be 1% or 10%? I think 1%, I tried to put a value of 0.1 to 1% but it seems that this is not how it works ... or it should work

**Operating System:**
Windows Server 2016

### LINKED ISSUE #684: Non-dropable item property
Rather than having a global bool in the server's config regarding dropping items upon dying, this could be an item editor configuration.

Have area to enter the percentage of drop the item upon dying.

0% do not drop upon dying.

**Additional context**
The intersect would probably overcome some challenges if that were to happen :3

When we set up an item drop chance when dying, it’s global for all items and that’s annoying, I don't want all items to be dropable, so I'm forced to mark bound checkbox, but I don't want them to be bound, I just want it not to drop, so I'm left with no choice.

Cheshire92:
Resolves #684 
Resolves #670

Added the following options to replace the singular Bound property on individual tems:
- CanDrop
- CanTrade
- CanSell
- CanBank
- CanBag
- Drop Chance on Death (%)

If an Item was bound prior to this update then a custom migration will make sure that **ALL** these options are disabled.
The global option for item Drop Chance on player death from the server configuration is no longer used globally, but rather as a default configuration value for new items (and existing non-bound items upon running the migration)

Also fixed a minor issue with a typo regarding item drop rates from players.

Image1

### DIFF
--- a/Intersect.Server/Entities/BankInterface.cs
+++ b/Intersect.Server/Entities/BankInterface.cs
@@ -107,14 +107,14 @@ namespace Intersect.Server.Entities
             {
                 if (mPlayer.Items[slot].ItemId != Guid.Empty)
                 {
+                    if (!itemBase.CanBank)
+                    {
+                        PacketSender.SendChatMsg(mPlayer, Strings.Items.nobank, ChatMessageType.Bank, CustomColors.Items.Bound);
+                        return false;
+                    }
+
                     lock (mLock)
                     {
-                        if (!itemBase.CanBank)
-                        {
-                            PacketSender.SendChatMsg(mPlayer, Strings.Items.nobank, ChatMessageType.Bank, CustomColors.Items.Bound);
-                            return false;
-                        }
-
                         if (itemBase.IsStackable)
                         {
                             if (amount >= mPlayer.Items[slot].Quantity)
--- a/Intersect.Server/Entities/BankInterface.cs
+++ b/Intersect.Server/Entities/BankInterface.cs
@@ -109,6 +109,12 @@ namespace Intersect.Server.Entities
                 {
                     lock (mLock)
                     {
+                        if (!itemBase.CanBank)
+                        {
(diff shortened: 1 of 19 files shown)


WeylonSantana:
@jcsnider @Cheshire92 I believe that this PR could also close these 2 feature requests
https://github.com/AscensionGameDev/Intersect-Engine/issues/464
https://github.com/AscensionGameDev/Intersect-Engine/issues/243

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -109,6 +109,12 @@ public bool TryDepositItem(int slot, int amount, bool sendUpdate = true)
                 {
                     lock (mLock)
                     {
+                        if (!itemBase.CanBank)
+                        {
+                            PacketSender.SendChatMsg(mPlayer, Strings.Items.nobank, ChatMessageType.Bank, CustomColors.Items.Bound);
+                            return false;
+                        }
+
```
jcsnider, line 117: Can we move this outside of the lock above? No point in taking the time to acquire a lock if we are just going to return out of the function.

jcsnider (changes requested):

Intersect.Server/Entities/BankInterface.cs
Cheshire92: Fair, just changed it.

jcsnider (approved):

PR #738, opened 2021-05-19
Hotkey for Guild Window

### LINKED ISSUE #724: Guild Shortcut
You forgot to add that. G to Guild.

Add guild shortcut and in the settings in options as well.

**Additional context**

AisenGaming:
Should resolve #724 
Not 100% sure on the JSON generation part but tested to work with a hand edited OptionsWindow.json

### DIFF
--- a/Intersect.Client/Core/Controls/ControlEnum.cs
+++ b/Intersect.Client/Core/Controls/ControlEnum.cs
@@ -56,6 +56,8 @@
 
         OpenSpells,
 
+        OpenGuild,
+
         OpenFriends,
 
         OpenSettings,
@@ -63,8 +65,6 @@
         OpenDebugger,
 
         OpenAdminPanel,
-
-        OpenGuild,
 
         ToggleGui
 
--- a/Intersect.Client/Core/Controls/ControlEnum.cs
+++ b/Intersect.Client/Core/Controls/ControlEnum.cs
@@ -64,6 +64,8 @@
 
         OpenAdminPanel,
 
+        OpenGuild,
+
         ToggleGui
 
     }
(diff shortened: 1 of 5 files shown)


Intersect.Client/Core/Controls/ControlEnum.cs
```diff
@@ -64,6 +64,8 @@ public enum Control
 
         OpenAdminPanel,
 
+        OpenGuild,
```
Cheshire92, line 67: Any particular reason this isn't at the end of the enum?

AisenGaming: No reason at all

jcsnider: Should we move it above OpenSettings so it is with the other opening of game windows (Inventory/Spells/Quests/etc)?

Intersect.Client/Core/Controls/Controls.cs
```diff
@@ -85,6 +85,7 @@ public void ResetDefaults()
             CreateControlMap(Control.OpenDebugger, Keys.F2, Keys.None);
             CreateControlMap(Control.OpenAdminPanel, Keys.Insert, Keys.None);
             CreateControlMap(Control.ToggleGui, Keys.F11, Keys.None);
+            CreateControlMap(Control.OpenGuild, Keys.G, Keys.None);
```
jcsnider, line 88: I'm not sure if this determines the order of the controls or not but we might want this above the OpenSettings line as well?

Kamusvalenzuela (dismissed):

Intersect.Client/Core/Controls/Controls.cs
Cheshire92: I'm not sure either but for the sake of clarity in the source files we might as well.

Intersect.Client/Core/Controls/ControlEnum.cs
Cheshire92: Probably yeah. I don't think the config uses the enum numbering but uses the names instead so it shouldn't make a difference where it's located.

Cheshire92 (dismissed):

jcsnider (approved):

Issue #722, opened 2021-05-12
Option to lose XP when dying


dhmhc:
**Is your feature request related to a problem? Please describe.**
I saw that the engine is getting quite modular and for all audiences, games, and it is always possible to do more with this engine due to its ability to serve various types of projects.

**Describe the solution you'd like**
And here is my suggestion, what if it was possible to set a percentage of XP loss on the server when dying?

Loses XP when dying: true / false
Percentage of loss on death: int%
It can regress level when dying when there is no more XP: true / false

Some important points if you return to the level when you have no more XP to lose.

1 - Remove the point that was used in the attribute. Or remove the attribute that was upgraded when gaining the distribution point
2 - Check equipment when losing the level, in this case, if the equipment is not at the same level as the character, he will automatically unequip.
I think these are the most important points so that there is no glitch when dying, as a character dying and losing xp on purpose to gain infinite points of attributes.
**Additional context**
I think it will make some projects more difficult, and many people may like this option in their game

da-rulez:
We made a patch for a previous Intersect version:
https://kasplant.itch.io/kash-shop-exp-loss-on-death

This could serve as a base for a PR. I imagine it's not too hard to make. 

This does not include leveling down. Which strikes me as way too game specific and bad game design / asking for trouble in 99% of cases.

Cheshire92:
Levelling down would require the game to start tracking what points were spent on for each level which it simply doesn't right now.. And wouldn't work when someone decides to save all their points for five levels and then spend them.

Losing experience isn't a hard ask, neither is levelling down technically.. But removing the points the user spent themselves last level might be tricky. I think the current system for it simply tears points away one by one until it reaches a valid threshold again.

dhmhc:
> Levelling down would require the game to start tracking what points were spent on for each level which it simply doesn't right now.. And wouldn't work when someone decides to save all their points for five levels and then spend them.
> 
> Losing experience isn't a hard ask, neither is levelling down technically.. But removing the points the user spent themselves last level might be tricky. I think the current system for it simply tears points away one by one until it reaches a valid threshold again.

In fact, this is almost impossible to do and it would not be worth the effort ...

WeylonSantana:
@dhmhc There is a "rudimentary" way to make the player lose XP when dying, create a variable and assign a negative value to that variable, then give the player exp and give exp by choosing this variable with a negative value, it will work.

[...]

PR #746, opened 2021-05-23
Option to lose XP when dying

dhmhc:
I don't know if I'm doing it right, I'm sorry if the way I'm doing it is invalid

It resolves: #722


dhmhc:
@jcsnider Sorry I never made a pull request like this, I sent the change via GitHub Desktop to my repository, right after, I tried to make a pull request to your source

blinkuz:
It seems to me that you have not uploaded the PR correctly, apart from that the Main branch from where you are trying to send is not the same as the prerelease, main = 6.2 so it is very likely that you will have to delete it.

jcsnider:
We do want all PRs to target the development branch and we based on that so we aren't committing or recommitting hundreds of commits.

dhmhc:
right, I opened another pull with the correct commit, I had requested a pull from main to prerelease. you can close this pull, the most recent I sent this correct.

PR #747, opened 2021-05-23
Option to lose XP when dying

dhmhc:
I think now I did it correctly, I noticed an error in the repositories when sending the pull

It resolves: #722


jcsnider:
Need one that PRs from your development into our development instead of into the prerelease branches.

jcsnider:
As an aside.. I dont know why this is messing with the .net framework version as well.. it shouldn't need to.

dhmhc:
> Precisa de um PRs de seu desenvolvimento para o nosso desenvolvimento, em vez de nos ramos de pré-lançamento.



> Need one that PRs from your development into our development instead of into the prerelease branches.

sorry i don't know how to do this.

dhmhc:
> As an aside.. I dont know why this is messing with the .net framework version as well.. it shouldn't need to.

I do not understand too. but always when low it tries to convert to a previous version

PR #748, opened 2021-05-24
Option to Loss Exp When Die

dhmhc:
Resolves: #722

### DIFF
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -59,6 +59,12 @@
         /// </summary>
         public bool ShowLevelByName = false;
 
+        /// <summary>
+        /// First thanks for letting me share an idea, of losing XP when dying <3
+        /// </summary>
+        public bool XPLossOnDeath = false;
+        public int XPLossPercent = 0;
+
     }
 
 }
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -62,7 +62,6 @@
         /// <summary>
         /// First thanks for letting me share an idea, of losing XP when dying <3
         /// </summary>
-        public bool XPLossOnDeath = false;
         public int XPLossPercent = 0;
 
     }
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -63,6 +63,7 @@
         /// First thanks for letting me share an idea, of losing XP when dying <3
         /// </summary>
         public int ExpLossOnDeathPercent = 0;
+        
         /// <summary>
         /// If true, it will remove the associated exp, otherwise you will lose the exp based on the exp required to level up.
         /// </summary>
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -63,6 +63,10 @@
         /// First thanks for letting me share an idea, of losing XP when dying <3
         /// </summary>
         public int ExpLossOnDeathPercent = 0;
+        /// <summary>
+        /// If true, it will remove the associated exp, otherwise you will lose the exp based on the exp required to level up.
+        /// </summary>
+        public bool ExpLossFromCurrentExp = true
 
     }
 
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -62,7 +62,7 @@
         /// <summary>
         /// First thanks for letting me share an idea, of losing XP when dying <3
         /// </summary>
-        public int XPLossPercent = 0;
+        public int ExpLossOnDeathPercent = 0;
 
     }
 
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -815,10 +815,17 @@ namespace Intersect.Server.Entities
 
             if (Options.Instance.Player.ExpLossOnDeathPercent > 0)
             {
+                if (Options.Player.ExpLossFromCurrentExp.Equals(true))
+                {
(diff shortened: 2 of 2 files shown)


Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -59,6 +59,12 @@ public class PlayerOptions
         /// </summary>
         public bool ShowLevelByName = false;
 
+        /// <summary>
+        /// First thanks for letting me share an idea, of losing XP when dying <3
+        /// </summary>
+        public bool XPLossOnDeath = false;
```
jcsnider, line 65: We don't need a boolean here.

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -59,6 +59,12 @@ public class PlayerOptions
...
+        public int XPLossPercent = 0;
```
jcsnider, line 66: Rename this to ExpLossOnDeathPercent

Intersect.Server/Entities/Player.cs
```diff
@@ -812,7 +812,13 @@ public override void Die(bool dropItems = true, Entity killer = null)
             {
                 base.Die(dropItems, killer);
             }
-            
+
+            if (Options.Player.XPLossOnDeath.Equals(true))
```
jcsnider, line 816: Then do

if (Options.Instance.Player.ExpLossOnDeathPercent > 0)

Intersect.Server/Entities/Player.cs
```diff
@@ -812,7 +812,13 @@ public override void Die(bool dropItems = true, Entity killer = null)
...
+            {
+                var ExpLoss = (GetExperienceToNextLevel(this.Level) * (Options.Player.XPLossPercent / 100.0));
```
jcsnider, line 818: This isn't calculating based on the current xp...

instead you want:
var ExpLoss = (this.Exp * (Options.Player.XPLossPercent / 100.0));

jcsnider (changes requested):

Intersect.Server/Entities/Player.cs
dhmhc: But the XP loss is calculated based on the total required to level and not the xp that the player currently has, that was purposeful, I believe that in most games the XP loss calculation is the percentage over what he needs to upar and not what he already has, no?

jcsnider: I might be wrong I but I don't think that's how everyone will want it.... how about one more boolean in the options?

public bool ExpLossFromCurrentExp {get; set;} = true

Then you can do an if/then to calculate exp loss based on both?

dhmhc: Great idea I will change the files here, and let the staff choose the mode

dhmhc: I think its all ok now 👯

dhmhc:
Any further changes needed? 🍰

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -59,6 +59,15 @@ public class PlayerOptions
         /// </summary>
         public bool ShowLevelByName = false;
 
+        /// <summary>
+        /// First thanks for letting me share an idea, of losing XP when dying <3
+        /// </summary>
+        public int ExpLossOnDeathPercent = 0;
+        /// <summary>
```
jcsnider, line 66: Please add a a line break before this summary so it matches the rest of the file aesthetically.

Intersect.Server/Entities/Player.cs
```diff
@@ -812,7 +812,20 @@ public override void Die(bool dropItems = true, Entity killer = null)
             {
                 base.Die(dropItems, killer);
             }
-            
+
+            if (Options.Instance.Player.ExpLossOnDeathPercent > 0)
+            {
+                if (Options.Player.ExpLossFromCurrentExp.Equals(true))
+                {
+                var ExpLoss = (this.Exp * (Options.Player.ExpLossOnDeathPercent / 100.0));
```
jcsnider, line 820: Please indent (add a tab) on these 2 lines.

Intersect.Server/Entities/Player.cs
```diff
@@ -812,7 +812,20 @@ public override void Die(bool dropItems = true, Entity killer = null)
...
+                TakeExperience((long)ExpLoss);
+                }
+                else
+                {
+                var ExpLoss = (GetExperienceToNextLevel(this.Level) * (Options.Player.ExpLossOnDeathPercent / 100.0));
```
jcsnider, line 825: Please indent (add a tab) on these 2 lines.

Intersect.Server/Entities/Player.cs
```diff
@@ -1010,6 +1023,16 @@ public void GiveExperience(long amount)
...
+        public void TakeExperience(long amount)
+        {
+            Exp -= amount;
+            if (Exp < 0)
+            {
+                Exp = 0;
+            }
+
+            PacketSender.SendExperience(this);
+        }
```
jcsnider, line 1035: Please add a line break after this function so there is space before the next.

jcsnider (changes requested):
Added a few more comments to fix up some spacing/aesthetic issues. Once fixed I think this will be good to merge.

Intersect.Server/Entities/Player.cs
dhmhc: Done

Intersect.Server/Entities/Player.cs
dhmhc: Done

Intersect (Core)/Config/PlayerOptions.cs
dhmhc: Done

jcsnider (approved):

Arufonsu:
@jcsnider i think that this should be moved to **Done**

Issue #753, opened 2021-05-31
'Guild System' bug #1 - (v0.7.0.120)


RandemGamor:
**Describe the bug**
A player can be invited to a guild if they are already in a guild. It should only let a player invite another player if they are without a guild.
As this will also eliminate another issue where a Guild Owner can join another guild which then removes them from their own guild. If they are the only person in the guild then the guild remains but can't be accessed anymore.

**Version Encountered**

 (v0.7.0.120)

**To Reproduce**
Inside the client, invite a player that is already in a guild.

**Expected behavior**
See above, under "describe the bug"

**Screenshots**
n/a

**Logs**
n/a

**Operating System:**
n/a

**Additional context**
n/a

PR #762, opened 2021-06-02
Guild invite fixes

AisenGaming:
This will resolve issue #753 
Error message from a failed invite isn't showing however.


Cheshire92:
> Error message from a failed invite isn't showing however.

I'm not sure I understand this line, does that mean the fix is incomplete?

AisenGaming:
> > Error message from a failed invite isn't showing however.
> 
> I'm not sure I understand this line, does that mean the fix is incomplete?

The fix works, but the message that's supposed to show in the event a failed invite happens doesn't show up in the chat. But it does what is inteded apart from that. I'm still working on the fix for the error message

jcsnider (approved):

Issue #749, opened 2021-05-24
Bonify Group Experience


helenatodev:
**Is your feature request related to a problem? Please describe.**
What do you think of a bonus in XP gain for each member of the group?

**Describe the solution you'd like**
Currently a monster's XP is divided ALSO for each member of the group. But for some members that would be a little unfair.
What is the benefit of going with 4 friends to kill 1 monster, where your EXP will be divided equally for everyone?

I believe that this would better influence the groups, make them have an advantage.

But of course, this would be optional in the server settings, and also the percentage of the gain to be increased ...

**Additional context**
And I come here to ask everyone, what do you think of that?
If they approve, I can send a PR with the changes.

WeylonSantana:
How would it work? The xp would no longer be divided, but increased by a percentage? Would we have the option to disable the xp division and exchange it for a bonus increase for each member? Would the xp bonus be made on top of the individual xp already divided by each member of the group? Could there be different values for each member or would it be a global value?

dhmhc:
I will answer in an example that will be easier for everyone to understand how the system will work.
Percentage configured on the gain server.
Ex: 10 = 10%
Monster gives 1000 XP.
1 player, alone, without a party, will win 1000 by killing him.
2 players in a party, will kill this monster.
The Exp that will be divided will be +20%
Then, the monster will now divide 1200 XP, totaling a division of 600 XP for each member.
In short, for each additional member in a group you will gain + 10% xp on top of the xp to be divided for EVERYONE.
The particular EXP I cannot change because there is already a configuration in the items that can add XP individual bonus for each one.

da-rulez:
Seems a little bit OP to me. You already have the advantage that you can kill the NPC a lot faster, so extra XP on top of that would be a double bonus. 

I personally wouldn't use this functionality but it can't hurt if you make it an option in the NPC editor I suppose?

dhmhc:
> Seems a little bit OP to me. You already have the advantage that you can kill the NPC a lot faster, so extra XP on top of that would be a double bonus.
> 
> I personally wouldn't use this functionality but it can't hurt if you make it an option in the NPC editor I suppose?

Making it available in the NPC editor would make it more complicated when configuring, and this is currently a global configuration.
You can use it for example in a BOSS, this option of you gaining more XP if you are in a group can gather players to do it.

[...]

PR #773, opened 2021-06-07
Bonify XP Party for Each Member

helenatodev:
Its Resolves: #749

Percentage configured on the gain server.
Ex: 10 = 10%
Monster gives 1000 XP.
1 player, alone, without a party, will win 1000 by killing him.
2 players in a party, will kill this monster.
The Exp that will be divided will be +20%
Then, the monster will now divide 1200 XP, totaling a division of 600 XP for each member.
In short, for each additional member in a group you will gain + 10% xp on top of the xp to be divided for EVERYONE.
The particular EXP I cannot change because there is already a configuration in the items that can add XP individual bonus for each one.

### DIFF
--- a/Intersect (Core)/Config/PartyOptions.cs
+++ b/Intersect (Core)/Config/PartyOptions.cs
@@ -14,5 +14,11 @@
 
         public int NpcDeathCommonEventStartRange = 0;
 
+        /// <summary>
+        /// It will determine if you will have an XP bonus for each member, Eg 10, will give 10% more to the monster's total XP, then split equally to all members.
+        /// A monster that would give 100 XP, now gives 120, getting 60 for each member, when 2 in the party.
+        /// </summary>
+        public float BonifyExpPercentPerMember = 0;
+
     }
 }
--- a/Intersect (Core)/Config/PartyOptions.cs
+++ b/Intersect (Core)/Config/PartyOptions.cs
@@ -18,7 +18,7 @@
         /// It will determine if you will have an XP bonus for each member, Eg 10, will give 10% more to the monster's total XP, then split equally to all members.
         /// A monster that would give 100 XP, now gives 120, getting 60 for each member, when 2 in the party.
         /// </summary>
-        public float BonifyExpPercentPerMember = 0;
+        public float BonusExperiencePercentPerMember = 0;
 
     }
 }
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -1069,8 +1069,8 @@ namespace Intersect.Server.Entities
                         if (Party != null && Party.Count > 0)
                         {
                             var partyMembersInXpRange = Party.Where(partyMember => partyMember.InRangeOf(this, Options.Party.SharedXpRange));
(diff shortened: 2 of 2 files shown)


Intersect (Core)/Config/PartyOptions.cs
```diff
@@ -14,5 +14,11 @@ public class PartyOptions
 
         public int NpcDeathCommonEventStartRange = 0;
 
+        /// <summary>
+        /// It will determine if you will have an XP bonus for each member, Eg 10, will give 10% more to the monster's total XP, then split equally to all members.
+        /// A monster that would give 100 XP, now gives 120, getting 60 for each member, when 2 in the party.
+        /// </summary>
+        public float BonifyExpPercentPerMember = 0;
```
Cheshire92, line 21: Shouldn't this be called something like "BonusExperiencePercentPerMember” for clarity?

Intersect.Server/Entities/Player.cs
```diff
@@ -1069,7 +1069,9 @@ public override void KilledEntity(Entity entity)
                         if (Party != null && Party.Count > 0)
                         {
                             var partyMembersInXpRange = Party.Where(partyMember => partyMember.InRangeOf(this, Options.Party.SharedXpRange));
```
Cheshire92, line 1071: Add .ToArray at the end here.

Intersect.Server/Entities/Player.cs
```diff
@@ -1069,7 +1069,9 @@ public override void KilledEntity(Entity entity)
...
-                            var partyExperience = descriptor.Experience / partyMembersInXpRange.Count();
+                            float bonifyXp = Options.Party.BonifyExpPercentPerMember / 100;
+                            var multiplier = 1.0f + (partyMembersInXpRange.Count() * bonifyXp);
```
Cheshire92, line 1073: Don't use .Count(), use .Length or .Count from the array we created before.

Intersect.Server/Entities/Player.cs
```diff
@@ -1069,7 +1069,9 @@ public override void KilledEntity(Entity entity)
                         if (Party != null && Party.Count > 0)
                         {
                             var partyMembersInXpRange = Party.Where(partyMember => partyMember.InRangeOf(this, Options.Party.SharedXpRange));
-                            var partyExperience = descriptor.Experience / partyMembersInXpRange.Count();
+                            float bonifyXp = Options.Party.BonifyExpPercentPerMember / 100;
```
Cheshire92, line 1072: bonusExp

Intersect.Server/Entities/Player.cs
```diff
@@ -1069,7 +1069,9 @@ public override void KilledEntity(Entity entity)
...
+                            var multiplier = 1.0f + (partyMembersInXpRange.Count() * bonifyXp);
+                            var partyExperience = (int)(descriptor.Experience * multiplier) / partyMembersInXpRange.Count();
```
Cheshire92, line 1074: Don't use .Count()

Cheshire92 (changes requested):

jcsnider:
Resolved the remaining requests.

Cheshire92 (approved):

PR #785, opened 2021-06-19
FIXES AscensionGameDev/Intersect-Engine#798: Healthbar Scaling Fix

AlexVild:
When tile sizes were changed in `server/config.json`, health bars did not render correctly - they never exceeded half the width of the healthbarbackground texture.

Now, they appropriately display according to the ratio of health remaining, even at larger tile sizes.


AlexVild:
image

An  example can be shown here - note the players' health in relation to the distance the healthbar is displaying.

Cheshire92:
According to our guidelines, could you write an issue for this and link it?

AlexVild:
> 
> 
> According to our guidelines, could you write an issue for this and link it?

Sure thing - done!

Cheshire92:
Cool, now from my rough remembering and understanding of the code your fix always sets it to render as if it's full. Could you test if it works with lower health percentages as well?

I might be wrong but never hurts to see.

AlexVild:
> 
> 
> Cool, now from my rough remembering and understanding of the code your fix always sets it to render as if it's full. Could you test if it works with lower health percentages as well?
> 
> I might be wrong but never hurts to see.

Yeah, the bar now displays exactly correctly. I did not expect my change to fix the problem and was kind of surprised when it did, lol. But ultimately I think we just weren't using the right maximum width.

Here you can see two entities, one with low health and one with near full. Both bars are displaying correctly.
image

I've also tested this change with the default tile sizes and it still worked.

Cheshire92:
Even with shield spells? Just to make sure. :) Those are finnicky things.

AlexVild:
> 
> 
> Even with shield spells? Just to make sure. :) Those are finnicky things.

Sure - I'll do more testing and get back to you :)

Cheshire92:
Thanks! Posting from my phone and no access to my computer right now.

AlexVild:
@Cheshire92 Finally got around to testing shield spells - they don't show up in the entity's health bar, but they do show in the top left menu. Is this intended, or no (should I see them on the entity bar?)

jcsnider:
They should show up on the entities health bar on the map as well.

AlexVild:
@jcsnider Mkay good to know - I'll have to do more digging. No i'm even more surprised that this fixed the health bar, lol.

AlexVild:
@jcsnider @Cheshire92 I have fixed this for shields as well; you can see the new changes now.

Tested with 32x32 and 64x64 tile sizes. Shield and health bar behavior was consistent across both tile sizes. The appropriate ratios were displayed, even at max and minimum health values.

One thing I _did_ notice in all this, however, is that we display shields differently in the Entity's health bar as compared to the target/player's health bar (in the top left of the UI). On the entity, the shield appears as a ratio of the player's _new, theoretical_ maximum HP (with the shield), while in the target/player UI widget-thing, it is displayed as it relates to the player's _true_ maximum HP.

Example: Consider a shield spell of power 10, and a player at 5/20 health. On cast, the player's "theoretical" HP is now 15. The health bar, in the top left of the screen, will display the player's HP as "15/20", and the shield bar _and_ health bar will display that ratio.

However, the _entity's_ health bar will display the total ratios between the shield and HP as if the player had "15/30" health.

Ultimately, this is a separate problem, and some sort of coherent design would need to be settled upon for implementing the solution. Not sure what the process is for that! But I'd happily open a ticket.

Cheshire92:
Correct, shield bars are inconsistent and this should be rectified in a separate issue.

jcsnider:
@AlexVild would you mind making a separate issue for the inconsistency. This looks good to me so I'll go ahead and get it merged.

jcsnider (approved):

PR #795, opened 2021-06-22
Resolves 794 by allowing items to be picked up from the map even if you cannot hold the entire amount

### LINKED ISSUE #794: Cannot pickup loot unless you have room for entire stack
**Describe the bug**
If you have 5 unstackable items on the same map tile, but only 2 open inventory slots, you cannot pickup any of those items, where in reality you should be able to grab 2 of them.

**Version Encountered**
0.7.0.121

**To Reproduce**
Spawn a large stack of unstackable items on a map.
This can be done through the editor or by dropping 1 item at a time in the same spot.
Then fill up your inventory with the exception of 1 open slot.
Try to grab an item from that stack.

**Expected behavior**
You should be able to pick up items until your inventory is full.

**Screenshots**
NA

**Logs**
NA

**Operating System:**
Windows, Mac, Linux

**Additional context**

jcsnider:
Resolves #794

### DIFF
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -1623,8 +1623,30 @@ namespace Intersect.Server.Networking
                         }
                         else
                         {
-                            // We couldn't give the player their item, notify them.
-                            PacketSender.SendChatMsg(player, Strings.Items.InventoryNoSpace, ChatMessageType.Inventory, CustomColors.Alerts.Error);
+                            var quantity = mapItem.Quantity;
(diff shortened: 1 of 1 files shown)


Cheshire92 (commented):
Don't we already have a method to give up to X items or overflow the rest back onto the map? Might be able to use those instead of recreating the logic for it.

jcsnider:
@Cheshire92 I seriously hate how easy that solution was -_-.... fixed.

Cheshire92:
Haha, i changed the inventory logic quite a bit a while ago and that just happens to be part of it.

Cheshire92 (approved):

PR #805, opened 2021-06-28
Hide quantity on single-stackable items

### LINKED ISSUE #804: Server configuration to not show a stackable item's quantity if there is only 1 of that item.
Following a "Runescape" design, some users may want to only stack most items when they're placed in a bank. The stackable implementation as-is is useful for implementing this, however, it looks a bit ugly having a (1) next to all of these items that aren't _truly_ stackable when not in a bank.

It would be nice to just have a server config value that can hide the display of an item's quantity if the item's quantity is <= 1.

AlexVild:
Resolves #804

### DIFF
--- a/Intersect (Core)/Config/Options.cs
+++ b/Intersect (Core)/Config/Options.cs
@@ -102,8 +102,6 @@ namespace Intersect
 
         public static int MaxInvItems => Instance.PlayerOpts.MaxInventory;
 
-        public static bool NoSingleQuantityDisplay => Instance.PlayerOpts.DoNotDisplayQuantityIfSingleItem;
-
         public static int MaxPlayerSkills => Instance.PlayerOpts.MaxSpells;
 
         public static int MaxBankSlots => Instance.PlayerOpts.MaxBank;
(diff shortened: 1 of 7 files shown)


jcsnider:
While I appreciate the PR I don't think this needs to be an option. Just hiding the quantity if < 1 is probably fine.

@Cheshire92 thoughts?

Cheshire92:
I'd also say it doesn't need to be an option. 
I doubt anyone is really attached to seeing they've got one item in their inventory.

AlexVild:
Lol, that's fair. I'll get to updating that and ping y'all when it's gone.

AlexVild:
@jcsnider @Cheshire92 It is done

Cheshire92 (dismissed):

jcsnider:
Should these changes apply to Bags as well?

Also can we do <= 1 instead of == 1. I have a bad feeling that there might be some older/legacy cases where you can have a non stackable item and quantity equals 0.

AlexVild:
Lemme check on bags. Otherwise, lol, sure, I'll make that change :)

AlexVild:
@jcsnider Made the logical change you requested; otherwise you were right about bags. Additionally, I added the logic to the trade window, and to the loot window.

jcsnider (approved):

Issue #810, opened 2021-06-29
feature: ability to toggle fading effects when warping


AlexVild:
**Is your feature request related to a problem? Please describe.**
The base-game "warp" functionality can be a bit jarring. Additionally, if you do a work around by making your own fade out system with the event editor (something like displaying increasingly opaque images), you then have to deal with fading back _in_ on any map that the user could have came from via fade out. It's not impossible, but kind of a pain.

**Describe the solution you'd like**
The engine already has a built-in fade system, but the editor is not aware of it. It would be nice to have a toggle option for "fade between warp" or something, like what RPG Maker has, **that we could check for both map warp tiles and the event editor "warp player" option.**

WeylonSantana:
I believe fade should be a command from the special effects area like play sound, show image, it should fade in and fade out, that way it would encompass much more than just teleports like cutscenes and so on.

WeylonSantana:
Closing this out, we have the common fade event, so we can fade, then warp, and then fade again.

PR #811, opened 2021-06-29
Fix AscensionGameDev/Intersect-Engine#810 - Introduce option to fade between warps

AlexVild:


Cheshire92:
Your retargeting created files that don't exist in prerelease anymore. Could you look through those and remove them?

Intersect.Editor/Localization/Strings.cs
```diff
@@ -1783,7 +1783,7 @@ public struct EventCommandList
 
             public static LocalizedString waitforroute = @"Wait for Move Route Completion of {00}";
 
-            public static LocalizedString warp = @"Warp Player [Map: {00} X: {01} Y: {02} Dir: {03}]";
+            public static LocalizedString warp = @"Warp Player [Map: {00} X: {01} Y: {02} Dir: {03} Fade: {04}]";
```
AlexVild, line 1786: I could not get this to work - any thoughts would be appreciated. It just does not show the new addition at all.

AlexVild (commented):
> 
> 
> Your retargeting created files that don't exist in prerelease anymore. Could you look through those and remove them?

Is that still the case? I originally accidentally went to merge with master, lol

Intersect.Editor/Localization/Strings.cs
Cheshire92: Delete your editor strings file and have it regenerate or edit the changes in manually. 

Unfortunately changed lines don't regenerate. Only new lines do.

AlexVild: Ah, right - makes sense, thanks!

AlexVild: (For what it's worth - does in fact work)

Arufonsu:
> Is that still the case? I originally accidentally went to merge with master, lol
>
You merged a  main branch into development, in order to avoid any mismatch with the latest dev branch, you may want to work with your PR from dev to dev branch instead.

As for this feature itself, yes! it would be nice to have a fadeout effect whenever you warp into another map! Regards 👍🏽 !

AlexVild:
> > Is that still the case? I originally accidentally went to merge with master, lol
> 
> You merged a main branch into development, in order to avoid any mismatch with the latest dev branch, you may want to work with your PR from dev to dev branch instead.
> 
> As for this feature itself, yes! it would be nice to have a fadeout effect whenever you warp into another map! Regards 👍🏽 !

Apologies if I'm being dense - but I'm not seeing any files in this PR that I don't expect to see?

Arufonsu:
> 
> Apologies if I'm being dense - but I'm not seeing any files in this PR that I don't expect to see?
>

oh is not that, is just that you are working with prerelease:

imagen
and for new features and testing in general, you might to do such PRs over the development branch >_<

AlexVild:
> > Apologies if I'm being dense - but I'm not seeing any files in this PR that I don't expect to see?
> 
> oh is not that, is just that you are working with prerelease:
> 
> imagen
> and for new features and testing in general, you might to do such PRs over the development branch >_<

Oooh :P I see.

Cheshire92:
The resx files are not supposed to be there. There was some work done to make them obsolete and save some merge conflicts resulting from them.

If you could remove those that would help.

AlexVild:
> 
> 
> The resx files are not supposed to be there. There was some work done to make them obsolete and save some merge conflicts resulting from them.
> 
> If you could remove those that would help.

If I remove them, I get build errors relating to their lack of existence :\. I wish I knew why they were created in the first place.

jcsnider:
Visual studio creates them to sometimes hold metadata for forms and controls, but in 90% of cases they are not useful or required.

You can delete the resx files, and then remove the respective lines from the Intersect Editor.csproj file [here](https://github.com/AscensionGameDev/Intersect-Engine/pull/811/files#diff-f958e7193ee8004c986b8adcfa6a308f01ba714ea01edb3072add2dc61f9dcbcR664-R666) and [here](https://github.com/AscensionGameDev/Intersect-Engine/pull/811/files#diff-f958e7193ee8004c986b8adcfa6a308f01ba714ea01edb3072add2dc61f9dcbcR670-R672).

AlexVild:
Thanks so much! I got rid of the _new_ resx files - though now it thinks I've modified another. Which is rad. Lemme know if _that_ is also a problem and I'll try to just tell git to not bother with that file. What an annoying thing lol

AlexVild:
Just pushed up a _preeeetty_ big bug fix. I wasn't saving the future coordinates anywhere so it was just always warping to spawn - I didn't realize because I tested in a dumb way and the warp location was the same as the spawn :|. I also wasn't keeping direction around at all.

This has been fixed, and is in the newest commit. Apologies for the sloppy testing.

jcsnider:
I don't like all the required back and forth communication in order for this warp to work....

When we depend on the client's input to complete the warp what happens when the client lags, dcs, or maybe is getting attacked while fading?

The warp won't complete upon logging back in, the client could die if being attacked. If the warp was done via an event then upon logging in that event might trigger again breaking games that way (handing out infinite rewards and such).

If this is going to be a provided option I'd much rather see the warp occur immediately on the server side, and then the server can send a packet telling the client to fade but continue rendering the old maps until the fade is done. (In which case the client would then allow it's entity to warp)

AlexVild:
> I don't like all the required back and forth communication in order for this warp to work....
> 
> When we depend on the client's input to complete the warp what happens when the client lags, dcs, or maybe is getting attacked while fading?
> 
> The warp won't complete upon logging back in, the client could die if being attacked. If the warp was done via an event then upon logging in that event might trigger again breaking games that way (handing out infinite rewards and such).
> 
> If this is going to be a provided option I'd much rather see the warp occur immediately on the server side, and then the server can send a packet telling the client to fade but continue rendering the old maps until the fade is done. (In which case the client would then allow it's entity to warp)

That's a good idea. I agree that there's a lot of chatter - but my decision at the time was that ultimately, the server is still just the source of truth of "where am I". I agree about the combat implications, but I felt that that was a designer's choice more than anything - if you're going to place a warp by a combat section, either don't use a transition on the warp, or place NPC avoids around it and make it an effective safe space.

One question I have - if I moved forward with your suggestion, wouldn't that pose similar gameplay problems? Consider warping to a new map with hostile NPC's by the warp point - my understanding is that the player would then be getting attacked, but on their client they wouldn't know about it until the warp was complete.

lodicolo:
> One question I have - if I moved forward with your suggestion, wouldn't that pose similar gameplay problems? Consider warping to a new map with hostile NPC's by the warp point - my understanding is that the player would then be getting attacked, but on their client they wouldn't know about it until the warp was complete.

This requires adding a new state to the entity on the server, but basically the player should be marked as a shadow until the client confirms they finished the fade-and-warp _or_ they disconnect (this would be a transient property that is not persisted to the database).

Shadow entities would not be visible (not sent) to other clients, and would not be valid targets for any entities (and thus would ignore damage, collision, etc).

AlexVild:
For the record, I have this on hold at the moment. I've been using it in my game with no problems - but agree that the current implementation is not ideal for something as big as putting this into prerelease or anything.

Currently i'm keeping it up because it feels _close_. I feel like what Panda suggested would resolve most of JC's concerns? But I'm worried about management of that state - having some sort of shadow property would be potentially disastrous if it got mis-managed and a player was in some sort of state where they could attack things that couldn't attack them, or vice versa.

Cheshire92:
No problem at all!
Just trying to clear out some old PRs :)

Issue #763, opened 2021-06-04
Additional Crafting Features - '+ .value to p.var' for example.


RandemGamor:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
So, I want to be able to give 'crafting' experience to a player when they craft specific items.
Now I know I can create a player variable like "smithingEXP".
But there's no way to + to the player variable value when they craft that I can see?
Like we need options to give even player exp, and value to a variable upon successful completion of the craft.

This also brings me to another feature that will be a real nice addition;
Crafting Requirement. This could be set like player level, or a variable requirement/condition.
like "if smithingLVL => 1" for example.

If they are able to craft it the item picture would be full colour.
If it wasn't able to be crafted, it would be faded out or black and white or have a little image overlaid to show it's not yet craftable and they don't meet the requirements.
We should have a way to state to the player what the requirements are.
So maybe a little 'optional description' box. we could write in say;
"Smithing Level 1 required" as an example.

**Additional context**
n/a

WeylonSantana:
A "run common event" after each crafting performed would solve this.

AisenGaming:
This is something I'm looking into already, will let you know when I have something

AisenGaming:
Adding to this real quick:
I have finished the code to trigger common events after a successful craft, I am currently in the middle of working out the database migration for it to be done. Once that's done I will be making a PR for it :)

worldofjimmy:
Could currently be made by:
Craft a "-Sword".
Common event: If player has a "-Sword" give X points to Variable X and give a "Sword" to the player.

But would love to have an option to run common event after a craft has been made.

PR #813, opened 2021-06-30
Add Common Event Triggers to Crafting Items

### LINKED ISSUE #812: Ability to add a crafting sound to a crafting table
Crafting feels a little strange without auditory feedback.

Being able to play a sound for each item that's crafted would be cool. I imagine a simple first-step implementation is just to allow the user to set a sound on a crafting table - this sound would play for the client every time it crafts an item while within that crafting table.

AisenGaming:
Resolves #812 & #763

### DIFF
--- a/Intersect.Editor/Forms/Editors/frmCrafts.cs
+++ b/Intersect.Editor/Forms/Editors/frmCrafts.cs
@@ -431,6 +431,7 @@ namespace Intersect.Editor.Forms.Editors
             grpIngredients.Text = Strings.CraftsEditor.ingredients;
             lblIngredient.Text = Strings.CraftsEditor.ingredientitem;
             lblQuantity.Text = Strings.CraftsEditor.ingredientquantity;
+            lblCommonEvent.Text = Strings.CraftsEditor.commonevent;
             btnAdd.Text = Strings.CraftsEditor.newingredient;
             btnRemove.Text = Strings.CraftsEditor.deleteingredient;
             btnDupIngredient.Text = Strings.CraftsEditor.duplicateingredient;
--- a/Intersect.Editor/Forms/Editors/frmCrafts.cs
+++ b/Intersect.Editor/Forms/Editors/frmCrafts.cs
@@ -451,6 +451,10 @@ namespace Intersect.Editor.Forms.Editors
             nudCraftQuantity.Value = Math.Max(1, nudCraftQuantity.Value);
             mEditorItem.Quantity = (int) nudCraftQuantity.Value;
         }
+        private void cmbEvent_SelectedIndexChanged(object sender, EventArgs e)
+        {
+            mEditorItem.Event = EventBase.Get(EventBase.IdFromList(cmbEvent.SelectedIndex - 1));
+        }
 
         #region "Item List - Folders, Searching, Sorting, Etc"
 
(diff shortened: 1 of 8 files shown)


Intersect.Editor/Forms/Editors/frmCrafts.cs
```diff
@@ -41,6 +42,9 @@ public FrmCrafts()
             cmbIngredient.Items.Clear();
             cmbIngredient.Items.Add(Strings.General.none);
             cmbIngredient.Items.AddRange(ItemBase.Names);
+            cmbEvent.Items.Clear();
```
Cheshire92, line 45: Localisation please.

Cheshire92 (changes requested):

Intersect.Editor/Forms/Editors/frmCrafts.cs
AisenGaming: Done

Intersect.Editor/Localization/Strings.cs
```diff
@@ -1141,6 +1141,8 @@ public struct CraftsEditor
 
             public static LocalizedString undotitle = @"Undo Changes";
 
+            public static LocalizedString commonevent = @"Common Event:";
```
Cheshire92, line 1144: Casing and JSON properties are required for new additions, see other recently added ones (ie Not null and null value handling)

Something we often forget but probably shouldn't.

jcsnider: I don't think we use NotNull anymore. Think Panda removed all those when we removes the Jetbrains libraries?

Cheshire92:
I'm also debating on whether this should be triggerable here or as an actual event trigger you can get through the autorun triggers.. maybe both?

@jcsnider thoughts?

Intersect.Editor/Localization/Strings.cs
Cheshire92: Is it gone from the others? 
Might be misremembering that.

jcsnider:
I like this. More inline with how others work, and using common event triggers would be a larger performance hit due to and looping/iteration to see what events to trigger.

Both would be fine, but this is a good start. If we do the common event triggers as well it can be a separate PR.

Cheshire92:
Fair. This is more versatile and less of a performance hit.
Just making sure.

Intersect.Editor/Localization/Strings.cs
AisenGaming: Not entirely sure I follow here on what you want me to change/add. The casing is the same as the other things around it.

jcsnider:
Fixed the merge conflict and added the null value handler.

Cheshire92 (approved):

Intersect.Editor/Forms/Editors/frmCrafts.cs
```diff
@@ -549,6 +554,10 @@ private void txtSearch_Click(object sender, EventArgs e)
 
         #endregion
 
+        private void darkComboBox1_SelectedIndexChanged(object sender, EventArgs e)
```
jcsnider, line 557: Rename this function to cmbEvent_SelectedIndexChanged and move to line 454 above
```
#region "Item List - Folders, Searching, Sorting, Etc"
```

jcsnider (changes requested):

Intersect.Editor/Forms/Editors/frmCrafts.cs
jcsnider: And then make sure it still compiles/runs and it should be good to go.

AisenGaming: Done

jcsnider (approved):

PR #848, opened 2021-07-10
Added OnUpdate callback for plugins

celojevic:


jcsnider:
This should be targeting the development branch instead of prerelease. I can try to change the target but it might bring over more than just your single commit, in which case we will need to have this PR remade.

lodicolo:
> This should be targeting the development branch instead of prerelease. I can try to change the target but it might bring over more than just your single commit, in which case we will need to have this PR remade.

Yes, based on the commit history this should be remade. Only 02ad0ef should be in this PR.

lodicolo (changes requested):
This should have been done as part of `ILifecycleHelper`.

Doing it this way mixes up the independence of services from the "main thread" that runs the tight-loop.

Services should boostrap, start, and stop. They are in charge of deciding how and when to "update".

In this case it is more of an event and should follow the same pattern as the `LifecycleChangeState` event in `IClientLifecycleHelper`/`ClientLifecycleHelper`.

celojevic:
Ok, I'll remake it to use a Lifecycle event and have it target the development branch.

PR #858, opened 2021-07-12
Fix map editor import export

Irokaiser:


jcsnider:
This needs to be able to handle events, which is why I haven't dealt with it yet myself. 

Problem being is that all the event ids need to be regenerated on import else they would conflict with event ids on other maps and overwrite those in the database. 

When generating the new event ids all the commands that reference those events within other events on the map such as setting move routes need to also be fixed to reference the new ids. 

Not sure how we have copying and pasting working across maps but that logic might serve as a good reference.

Cheshire92:
@Irokaiser Apologies for the tag, I know you're busy. :)
Are you planning to make the requested changes or shall we close this for now to clean up the PR list?

PR #862, opened 2021-07-13
Generates a plugin JSON config file immediately if none exist

celojevic:
At the moment, if the JSON config file doesn't exist in a given plugin subdirectory, a config file will be automatically generated as the base PluginConfiguration class. I.E. it would be missing the derived PluginConfiguration class's properties. Otherwise, you'd have to run the app twice to generate the proper config file or manually type it in, which obviously isn't ideal.

If you generate the JSON config file immediately as done in this commit, it will cause a backup to also be created immediately and the regular config file to be updated so that the derived class's properties are included.


lodicolo:
This suppresses an intentional error message without fixing the underlying issue (the initializer).

I opened a PR with the correct solution #867

Issue #506, opened 2021-02-06
Option : Regeneration while in Battle.


Spectreye:
**Is your feature request related to a problem? Please describe.**
Now due to how the reset radius of monster work, it's impossible to have Regen while In Combat, it would be very nice if the option to toggle the regen in fight was available.

**Describe the solution you'd like**

Add a new option line in server config here : 
  "Combat": {
    "BlockingSlow": 30,
    "CombatTime": 10000,
    "MaxAttackRate": 200,
    "MaxDashSpeed": 200,
    "MinAttackRate": 500,
    "RegenTime": 3000,
    **_"CombatRegen": True/False_**

It will allow game creator to choose if they want the regen or not.

**Additional context**
Jc provided me the part of the code where it's handled : 

//Regen Timers
                    if (timeMs > CombatTimer && timeMs > RegenTimer)
                    {
                        ProcessRegen();
                        RegenTimer = timeMs + Options.RegenTime;
                    
Removing  **timeMs > CombatTimer &&**  
would do the tricks, but it would be far better foreveryone to make it an option.

Thank you by advance to whoever can do this :-)

blinkuz:
If I take a little time tomorrow I can make the change and send the PR it is something simple to make it optional and more than providing the condition that handles that part.

jcsnider:
@blinkuz get to work :P

blinkuz:
Hahaha and here we see my bad memory working, I completely forgot about this change, lately I am working on new programming in the engine for my project so most likely TODAY YES, I will fulfill the promise of doing this.

PR #875, opened 2021-07-16
Add option in server to enable regeneration vitals in combat

blinkuz:
It is a very basic PR but it is linked to this issue https://github.com/AscensionGameDev/Intersect-Engine/issues/506, I have used the suggestions that they had written but without eliminating the default behavior.

### DIFF
--- a/Intersect (Core)/Config/CombatOptions.cs
+++ b/Intersect (Core)/Config/CombatOptions.cs
@@ -17,6 +17,8 @@
         //Combat
         public int RegenTime = 3000; //3 seconds
 
+        public bool CombatRegen = false; // Enable or disable regeneration in combat.
+
         public bool EnableCombatChatMessages = false; // Enables or disables combat chat messages.
 
         //Spells
--- a/Intersect (Core)/Config/CombatOptions.cs
+++ b/Intersect (Core)/Config/CombatOptions.cs
@@ -16,9 +16,7 @@
 
         //Combat
         public int RegenTime = 3000; //3 seconds
-
-        public bool CombatRegen = false; // Enable or disable regeneration in combat.
-
+        
         public bool EnableCombatChatMessages = false; // Enables or disables combat chat messages.
 
         //Spells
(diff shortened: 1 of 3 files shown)


Intersect (Core)/Config/Options.cs
```diff
@@ -128,6 +128,8 @@ public class Options
 
         public static int RegenTime => Instance.CombatOpts.RegenTime;
 
+        public static bool CombatRegen => Instance.CombatOpts.CombatRegen;
```
Cheshire92, line 131: We're not doing new top level options like this anymore.
Please refer to the option through the category it is in.

Intersect (Core)/Config/CombatOptions.cs
```diff
@@ -17,6 +17,8 @@ public class CombatOptions
         //Combat
         public int RegenTime = 3000; //3 seconds
 
+        public bool CombatRegen = false; // Enable or disable regeneration in combat.
```
Cheshire92, line 20: Instead of making this a comment could you please make it actual XML documentation?

Cheshire92 (changes requested):

Intersect (Core)/Config/CombatOptions.cs
blinkuz: Last night I considered doing it but it was 1AM, today I will adjust the comment correctly.

Intersect (Core)/Config/Options.cs
blinkuz: So would it be to remove the top-level option and then just call an:

` Options.Instance.CombatOpts.CombatRegen`

?

Cheshire92: Yeah, might be worth naming it something descriptive as well that tells the user what it does.

Cheshire92 (changes requested):
Ah sorry missed this one between work. 

If you'd remove the two extra newlines you've added this will pretty much be good to go.

blinkuz:
> Ah sorry missed this one between work.
> 
> If you'd remove the two extra newlines you've added this will pretty much be good to go.

But those lines already existed originally, I had previously deleted them and then added them to leave everything as it was or am I wrong?

Cheshire92 (approved):

PR #903, opened 2021-08-10
Resolves #902: Add ability to set a separate animation for DoT/HoT spells.

### LINKED ISSUE #902: Allow a secondary animation selection for Heal/Damage over Time spells
The HoT/DoT feature on spells works great - but it would be nice to have a separate animation option for whenever HoT/DoT is supplied, vs when the spell first "hits" its enemy. This would allow for an easy implementation of consistent-looking status effects - i.e, a poison slash that deals damage, then subsequently deals poison damage over the next few seconds. It looks kind of weird to have a slash animation everytime you're damaged - it would be prefered to just show a separate "poisoned" animation on the subsequent casts while DoT is in effect.

Another dropdown on the spell creation window, near the "Heal/Damage over Time?" Checkbox, that, if that checkbox is checked, allows you to select an optional, second animation for the overtime casts.

AlexVild:
Resolves #902 

This change adds a new ability to the spell editor that allows the user to select a separate animation for DoT/HoT, effectively allowing a way to delineate between a spell being cast, and its ongoing effect happening. This will be useful for things such as a poison/bleed effect, or a stun spell.

Looks like this in-editor:
image


Cheshire92:
Could this be targeted into development please?
Prerelease should in theory not receive new features or bugfixes unless they come down from development or are super critical.

AlexVild:
> 
> 
> Could this be targeted into development please?
> Prerelease should in theory not receive new features or bugfixes unless they come down from development or are super critical.

My bad - yes, I can make that change.

Intersect (Core)/GameObjects/SpellBase.cs
```diff
@@ -59,6 +60,16 @@ public AnimationBase HitAnimation
             set => HitAnimationId = value?.Id ?? Guid.Empty;
         }
 
+        [Column("OverTimeAnimation")]
+        public Guid OverTimeAnimationId { get; set; }
+
+        [JsonIgnore]
```
lodicolo, line 66: Needs to be both `JsonIgnore` and `NotMapped`

Intersect.Editor/Forms/Editors/frmSpell.cs
```diff
@@ -213,6 +216,7 @@ private void InitLocalization()
             grpHotDot.Text = Strings.SpellEditor.hotdot;
             chkHOTDOT.Text = Strings.SpellEditor.ishotdot;
             lblTick.Text = Strings.SpellEditor.hotdottick;
+            lblOTanimationDisclaimer.Text = "(Set to " + Strings.General.none + " to keep same animation)";
```
lodicolo, line 219: This entire thing needs to be translated

Intersect.Server/Entities/Combat/DoT.cs
```diff
@@ -96,11 +96,19 @@ public void Tick()
...
             var deadAnimations = new List<KeyValuePair<Guid, sbyte>>();
             var aliveAnimations = new List<KeyValuePair<Guid, sbyte>>();
-            if (SpellBase.HitAnimationId != Guid.Empty)
+            if (SpellBase.OverTimeAnimationId != Guid.Empty)
             {
-                deadAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.HitAnimationId, (sbyte) Directions.Up));
-                aliveAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.HitAnimationId, (sbyte) Directions.Up));
+                deadAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.OverTimeAnimationId, (sbyte)Directions.Up));
+                aliveAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.OverTimeAnimationId, (sbyte)Directions.Up));
+            } else
```
lodicolo, line 103: `else if`

Intersect.Editor/Forms/Editors/frmSpell.cs
```diff
@@ -1067,6 +1072,11 @@ private void txtSearch_Click(object sender, EventArgs e)
         }
 
         #endregion
+
+        private void cmbOverTimeAnimation_SelectedIndexChanged(object sender, EventArgs e)
+        {
+            mEditorItem.OverTimeAnimation = AnimationBase.Get(AnimationBase.IdFromList(cmbOverTimeAnimation.SelectedIndex - 1));
```
lodicolo, line 1078: Put the animation id into a variable before calling Get, makes for easier debugging.

lodicolo (changes requested):

lodicolo:
Need to include "Resolves #" so that there's a link, and then link the issue (right side)

Intersect.Editor/Forms/Editors/frmSpell.cs
AlexVild: I think I did this in a really dumb way - lemme know, but otherwise done

Intersect.Server/Entities/Combat/DoT.cs
AlexVild: Done

Intersect.Editor/Forms/Editors/frmSpell.cs
AlexVild: Done, good call

Intersect (Core)/GameObjects/SpellBase.cs
AlexVild: Mhmm, done

AlexVild:
> 
> 
> Need to include "Resolves #" so that there's a link, and then link the issue (right side)

I don't know if I'm just dumb, but I've _never_ gotten this to work. I just see this, no matter which git prefix I assign to the title:

image

Intersect.Editor/Forms/Editors/frmSpell.cs
```diff
@@ -216,7 +216,7 @@ private void InitLocalization()
             grpHotDot.Text = Strings.SpellEditor.hotdot;
             chkHOTDOT.Text = Strings.SpellEditor.ishotdot;
             lblTick.Text = Strings.SpellEditor.hotdottick;
-            lblOTanimationDisclaimer.Text = "(Set to " + Strings.General.none + " to keep same animation)";
+            lblOTanimationDisclaimer.Text = Strings.SpellEditor.overTimeDisclaimer1 + Strings.General.none + Strings.SpellEditor.overTimeDisclaimer2;
```
lodicolo, line 219: Yup, you did. Look at other strings, there's a parameterization feature.

lodicolo (changes requested):

lodicolo:
> > Need to include "Resolves #" so that there's a link, and then link the issue (right side)
> 
> I don't know if I'm just dumb, but I've _never_ gotten this to work. I just see this, no matter which git prefix I assign to the title:
> 
> image

It's not a title prefix, you're supposed to put it in the body of the post, the title should just be something like `fix: what changed`.

That said, because we aren't opening a PR against the main branch the auto-linker doesn't work (or didn't used to) so that might still be manual.

Intersect.Server/Entities/Combat/DoT.cs
```diff
@@ -96,11 +96,16 @@ public void Tick()
 
             var deadAnimations = new List<KeyValuePair<Guid, sbyte>>();
             var aliveAnimations = new List<KeyValuePair<Guid, sbyte>>();
-            if (SpellBase.HitAnimationId != Guid.Empty)
+            if (SpellBase.OverTimeAnimationId != Guid.Empty)
             {
-                deadAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.HitAnimationId, (sbyte) Directions.Up));
-                aliveAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.HitAnimationId, (sbyte) Directions.Up));
+                deadAnimations.Add(new KeyValuePair<Guid, sbyte>(SpellBase.OverTimeAnimationId, (sbyte)Directions.Up));
```
lodicolo, line 101: So I just realized that a KVP doesn't need to be created twice (which introduces an extra line to include a bug) so the KVP should be stored in a variable and then reused for both lists, but...

What is the difference between the over-time animation and the hit animation in this case?

If we use an if statement like this, for DoTs I think they won't have a hit animation anymore which doesn't seem intentional?

lodicolo (changes requested):

Intersect.Server/Entities/Combat/DoT.cs
AlexVild: Hm, sorry if I'm not following - the DoT has either the spell's set DoT animation or, if not that, uses the default hit animation of the spell. If neither, then no animation is added.

The difference between the two is just convenience - otherwise, if we _force_ the user to select an OT animation (and don't assume the default spell one if none selected), the migration will have to be slightly more involved.

Intersect.Editor/Forms/Editors/frmSpell.cs
AlexVild: Good stuff, thank you. I like that

Intersect.Server/Entities/Combat/DoT.cs
lodicolo: Let me rephrase, does the on-hit animation still display when the spell hits?

lodicolo:
The disclaimer probably should be hover-text for animation label, not its own label.

I'm not sure how much I like "OT Animation" as label text. I don't really have a better suggestion, but it feels rather cryptic and not self explanatory. The fact that you needed to add the disclaimer text in the first place is kind of evidence towards that point.

Intersect.Server/Entities/Combat/DoT.cs
AlexVild: Sorry - yes. When the spell first hits, the on-hit animation will play. Subsequent "over time" animations will use whatever is set.

AlexVild:
> 
> 
> The disclaimer probably should be hover-text for animation label, not its own label.
> 
> I'm not sure how much I like "OT Animation" as label text. I don't really have a better suggestion, but it feels rather cryptic and not self explanatory. The fact that you needed to add the disclaimer text in the first place is kind of evidence towards that point.

I think that's fair. I could just leave it as "Animation"? It's in the "DoT/HoT" area of the editor

Cheshire92:
How about "Tick Animation"?

AlexVild:
> 
> 
> How about "Tick Animation"?

I like that! If @lodicolo is okay with it I'll make the change

lodicolo:
> > How about "Tick Animation"?
> 
> I like that! If @lodicolo is okay with it I'll make the change

Sounds fine to me. Remove the disclaimer text entirely with this change. Also make sure the "Tick Animation" label is translated.

AlexVild:
> 
> 
> > > How about "Tick Animation"?
> > 
> > 
> > I like that! If @lodicolo is okay with it I'll make the change
> 
> Sounds fine to me. Remove the disclaimer text entirely with this change. Also make sure the "Tick Animation" label is translated.

@lodicolo @Cheshire92 Rename complete

lodicolo (commented):
Can you provide a new screenshot to show what the interface looks like now, including a translated string (can be faked just to show that the translation is working)? @AlexVild

Arufonsu:
Hello, just saw the PR's description, i think that the "None" option should do what is intended to do (set no secondary animation in this case). Perhaps, adding an extra "Spell Base Animation" option to keep repeating the spell's base animation into the combobox might be better, then, you wouldn't need the disclaimer regardless this under the combobox. 

Regards and nice idea u guys got here!

Cheshire92:
@AlexVild Any progress on the requested screenshot?

AlexVild:
@Cheshire92 Completely forgot about this. Hmm
.. do you you actually use the translated string system? Is there a language config somewhere

Cheshire92:
They're in this file: https://github.com/AscensionGameDev/Intersect-Engine/blob/development/Intersect.Editor/Localization/Strings.cs

And you might need to merge the current dev branch in to fix some conflicts.

Arufonsu:
> no matter which git prefix I assign to the title

imagen

just fixed this for you, apparently git's web view differs from the closing/link keywords 😕 this has been going on for a while now, along with the closing/link keywords, you *have* to link issues/PRs though that little "Development" menu in the right.

lodicolo (approved):

PR #904, opened 2021-08-10
FIX 864: Crafts Editor "Common Event" combobox now displays correct value when flipping between crafts in the editor

### LINKED ISSUE #864: Crafts common event UI slightly broken
Using the most recent development branch as of this moment. When changing common events of crafts, they work properly, but it's impossible to tell which item has which common event. When you change a craft you're looking at, the "Common Event" does not change. It's kind of confusing, because for ex. when you're trying to make the same common event for a few crafts you don't get any feedback if the editor "caught" your change.

Simply swap the displayed common event in the dropdown menu to the craft's chosen common event when changing the viewed crafts, instead of keeping the last chosen common event.

AlexVild:
Resolves #864 

UpdateEditor() was not updated with the change that introduced the craft common event.

### DIFF
--- a/Intersect.Editor/Forms/Editors/frmCrafts.cs
+++ b/Intersect.Editor/Forms/Editors/frmCrafts.cs
@@ -122,6 +122,8 @@ namespace Intersect.Editor.Forms.Editors
                     mChanged.Add(mEditorItem);
                     mEditorItem.MakeBackup();
                 }
+
+                cmbEvent.SelectedIndex = EventBase.ListIndex(mEditorItem.EventId) + 1;
             }
             else
             {


Cheshire92:
Could this be targeted into development please?
Would save some porting into development and back to prerelease later.
I don't think it's particularly critical enough that this goes live right now. Can probably wait another little while.

AlexVild:
> 
> 
> Could this be targeted into development please?
> Would save some porting into development and back to prerelease later.
> I don't think it's particularly critical enough that this goes live right now. Can probably wait another little while.

Done for this and FIX-910

lodicolo (approved):

PR #906, opened 2021-08-12
Showing EXP after killing a NPC

pedroolyntho:
Hello! I thought would be nice to have a way to show in the chat or set in a variable how much EXP you have earned after defeated a NPC. 
Wouldn't it? Thanks!


Cheshire92:
This should be an issue, not a pull request.

PR #910, opened 2021-08-13
Scale feature

aerayth:
This adds a much-needed feature to Intersect. I like to have the logo bigger than what Intersect provides and the logo deserves to be part of the background image in my honest opinion rather than a separate image. It is much harder to scale the logo and position it in the background. This pull request adds a scale feature to the menu with two supported aspect ratios for the background, 16 by 9 and 4 by 3. If this is not accepted it is okay, I just didn't want to mess with my logo at the top of the GUI and change the entire layout to position the logo to the right of the GUI.

### DIFF
--- a/Intersect (Core)/Configuration/ClientConfiguration.cs
+++ b/Intersect (Core)/Configuration/ClientConfiguration.cs
@@ -46,7 +46,11 @@ namespace Intersect.Configuration
 
         public const string DEFAULT_MENU_BACKGROUND = "background.png";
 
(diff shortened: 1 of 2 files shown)


lodicolo:
- Not tied to an existing feature request
- Targeted to the wrong branch
- Unnecessarily rigid configuration structure
  - Does not accommodate other potentially desired form factors
  - Is not compatible with existing configurations
  - Has a concept of a form-factor based texture set but individually addresses specific form factors (and the names are not informative)
- Inconsistent formatting of code

Issue #911, opened 2021-08-14
No fullscreen screen scale texture


aerayth:
**Is your feature request related to a problem? Please describe.**
I'm trying to use a background image in my game that is scaled rather than padded.

**Describe the solution you'd like**
I would like to have a way to draw a background image that scales the image, centers it, and adds black borders.

PR #912, opened 2021-08-14
FEATURE #911: Scale Background Images

aerayth:
This is the initial commit, with just the drawing function of textures which scales the x and y, I was seeing if the development team was interested in this for background to scale based on the resolution, because right now the background images in Intersect are very boring with how it draws them. You can't really have background images with like characters or anything.

### DIFF
--- a/Intersect.Client/Core/Graphics.cs
+++ b/Intersect.Client/Core/Graphics.cs
@@ -769,6 +769,22 @@ namespace Intersect.Client.Core
             }
         }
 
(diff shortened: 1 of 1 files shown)


Cheshire92:
For future reference (in case anyone reads this), if you're going to start a PR that's a first draft of an idea to gauge interest, please mark it as a draft PR.

That way we know its not supposed to be a finished product and will behave accordingly.

PR #920, opened 2021-08-20
chore: Document IGameRenderer and cleanup

Cheshire92:
Figured I should document these methods and clean up some of the variable names.
Might as well expose them properly if we're going to use them in plugins!


Cheshire92:
Just noticed I did a booboo. Converting to a draft so I can fix it. :)

lodicolo (approved):

PR #926, opened 2021-08-29
Moves 'MaxHotBar' setting to Server's PlayerOptions.

Arufonsu:
This Pull Request should solve the issue #896 

Also: 

- Re-ordered Server's PlayerOptions Alphabetically.
- Added an XML documentation comment for the class PlayerOptions.
- Completely Removed 'MaxHotBar' setting from Core's Config.Options.
- Removed some unused imports.

Preview/Example:

imagen

imagen

Issue #896, opened 2021-07-30
Moving 'MaxHotBar' into server config


froggytrav:
According to JC: "It's actually not in the config. Instead it's hard coded in the source. https://github.com/AscensionGameDev/Intersect-Engine/blob/b48cea53f5e757b7f4961373e01fa777f768d0e6/Intersect%20(Core)/Config/Options.cs#L299"

**Describe the solution you'd like**
Moving the option into config.json or client config

blinkuz:
I can do a PR for this if no one else is doing it, it doesn't look like a complex task to do.

Arufonsu:
Opened a PR that solves this issue.
Should consider to close #895 as well, in order to avoid redundancy in the topic.

PR #928, opened 2021-08-31
PR #926 (Closed) Review Changes.

Arufonsu:
This Pull Request should solve the issue #896 

Also: 

- Re-ordered Server's PlayerOptions Alphabetically.
- Added an XML documentation comment for the class PlayerOptions.
- Completely Removed 'MaxHotBar' setting from Core's Config.Options.
- Removed some unused imports.

Preview/Example:

imagen

imagen

31-08-2021: Closed PR 926 in order to:

* Make a more clear PR along with @Cheshire92 's Review Changes.
* RollBack and Update to remote dev branch in order to prevent further migration conflicts.
* Undid changes that were unrelated to the Pull Request.
* Edited a more descriptive XML Comment for ExpLossOnDeathPercent PlayerOption.

### DIFF
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -1,8 +1,25 @@
 ﻿namespace Intersect.Config
 {
-
+    /// <summary>
+    /// Contains configurable options pertaining to the way Players are handled by the engine.
+    /// </summary>
     public class PlayerOptions
     {
+
+        /// <summary>
+        /// Unlinks the timers for combat and movement to facilitate complex combat (e.g. kiting)
+        /// </summary>
+        public bool AllowCombatMovement = true;
+
+        /// <summary>
+        /// If true, it will remove the associated exp, otherwise players will lose experience based on the amount required to level up.
+        /// </summary>
+        public bool ExpLossFromCurrentExp = true;
+
+        /// <summary>
+        /// A percentage between 0 and 100 which determines the experience that players will lose when they die.
+        /// </summary>
+        public int ExpLossOnDeathPercent = 0;
 
         /// <summary>
         /// A percentage between 0 and 100 which determines the chance in which they will lose any given item in their inventory when killed.
@@ -18,6 +35,13 @@
         /// Number of characters an account may create.
         /// </summary>
         public int MaxCharacters = 1;
+
+        /// <summary>
+        /// Number of hotbar slots a player has.
+        /// Default value is 10. If you change the default value (e.g. 16 slots), make sure to properly
+        /// modify the Client's HotbarWindow layout at: 'resources\gui\layouts\game\HotbarWindow.json'.
+        /// </summary>
+        public int MaxHotbar = 10;
 
         /// <summary>
         /// Number of inventory slots a player has.
@@ -45,29 +69,14 @@
         public int RequestTimeout = 300000;
 
         /// <summary>
-        /// Distance (in tiles) between players in which a trade offer can be sent and accepted.
-        /// </summary>
-        public int TradeRange = 6;
-
-        /// <summary>
-        /// Unlinks the timers for combat and movement to facilitate complex combat (e.g. kiting)
-        /// </summary>
-        public bool AllowCombatMovement = true;
-
-        /// <summary>
         /// Configures whether or not the level of a player is shown next to their name.
         /// </summary>
         public bool ShowLevelByName = false;
 
         /// <summary>
-        /// First thanks for letting me share an idea, of losing XP when dying <3
+        /// Distance (in tiles) between players in which a trade offer can be sent and accepted.
         /// </summary>
-        public int ExpLossOnDeathPercent = 0;
-
-        /// <summary>
-        /// If true, it will remove the associated exp, otherwise you will lose the exp based on the exp required to level up.
-        /// </summary>
-        public bool ExpLossFromCurrentExp = true;
+        public int TradeRange = 6;
 
     }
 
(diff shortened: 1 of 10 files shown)


Cheshire92:
I'll have to read through this again later comparing it to the other pr. Would rather have seen that one fixed as it has the history and there's reading involved. But I'll get to it somewhere during the week.

Cheshire92:
Looks good, was just re-reading and suddenly thought of what the option should be called though. MaxHotbar isn't a very descriptive options name. Wouldn't something like HotbarSlotCount or anything along those lines work better?

@jcsnider 
@lodicolo

Arufonsu:
> HotbarSlotCount

I think it sounds good and makes more sense👍🏽, how about others?

MaxCharacters > MaxCharacterSlots
MaxInventory > MaxInventorySlots
MaxBank > MaxBankSlots

Altho, that might require another PR.

lodicolo:
> Looks good, was just re-reading and suddenly thought of what the option should be called though. MaxHotbar isn't a very descriptive options name. Wouldn't something like HotbarSlotCount or anything along those lines work better?
 
 I actually agree, but for a slightly different reason -- MaxHotbar is misleading because there is not 1 slot all the way up to `MaxHotbar`, there are always exactly `MaxHotbar` number of slots. `HotbarSlotCount` is exactly accurate to how it is actually used. Also as a note -- do _not_ do `HotbarCount` because that should be used for a future implementation of multiple hotbars. :)

lodicolo:
> > HotbarSlotCount
> 
> I think it sounds good and makes more sense👍🏽, how about others?
> 
> MaxCharacters -> CharacterSlotCount
> MaxInventory -> InventorySlotCount
> MaxBank -> BankSlotCount
> 
> _i have my doubts about purging out the "Max" preffix tho._
> 
> MaxHotbar -> MaxHotbarSlots
> MaxCharacters -> MaxCharacterSlots
> MaxInventory -> MaxInventorySlots
> MaxBank -> MaxBankSlots
> 
> Altho, that might require another PR.

> MaxCharacters -> CharacterSlotCount
No, because you can create characters up to the limit, and at least if I remember correctly not-yet-created characters aren't displayed in the player selection screen (or maybe only one to signify that more can be created).

> MaxInventory -> InventorySlotCount
> MaxBank -> BankSlotCount
These two I would also suggest not changing because IMO `InventorySlotCount` and `BankSlotCount` should actually be saved to the player or account itself -- Max defining the limit to upgrades, and your suggestions being used for the current value (base + upgrades). No, there isn't such a system of upgrading slot count at this time, but we should not introduce naming collisions when there's a clear space for a feature here.

> MaxCharacters -> MaxCharacterSlots
> MaxInventory -> MaxInventorySlots
> MaxBank -> MaxBankSlots
These suggestions are better, but these are just renames not moving. Should be one separate PR for renames.

Arufonsu:
>
> These suggestions are better, but these are just renames not moving. Should be one separate PR for renames.

alright👍🏽 then, shall i commit the renaming for MaxHotbar -> HotbarSlotCount? Let me know if that's okay, and i will commit right away, it's basically already done locally :

imagen

Cheshire92:
Sounds good to me.
Just wanted someone else's opinion besides my own on that change. :)

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -41,7 +41,7 @@ public class PlayerOptions
         /// Default value is 10. If you change the default value (e.g. 16 slots), make sure to properly
         /// modify the Client's HotbarWindow layout at: 'resources\gui\layouts\game\HotbarWindow.json'.
         /// </summary>
-        public int MaxHotbar = 10;
+        public int HotbarSlotCount = 10;
```
lodicolo, line 44: This should be a property not a field

lodicolo (changes requested):

Intersect (Core)/Config/PlayerOptions.cs
Arufonsu: Alright, will it do as AutoProperty? and how about the rest of PlayerOptions?

Arufonsu: Alright, will it do as AutoProperty?
Also, how about the rest of PlayerOptions fields?

Arufonsu: Yeah, realized about the difference afterwards, an Auto-property would made it a readonly-sort of property, which is not what we want here, since we want to be able to change this though the Json config files. Thanks for clarifying ! 👍🏽

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -1,9 +1,26 @@
...
+        /// <summary>
+        /// If true, it will remove the associated exp, otherwise players will lose experience based on the amount required to level up.
+        /// </summary>
+        public bool ExpLossFromCurrentExp = true;
+
+        /// <summary>
+        /// A percentage between 0 and 100 which determines the experience that players will lose when they die.
+        /// </summary>
+        public int ExpLossOnDeathPercent = 0;
+
```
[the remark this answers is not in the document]
Arufonsu: 👍🏽 Undid.

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -19,6 +36,13 @@ public class PlayerOptions
         /// </summary>
         public int MaxCharacters = 1;
 
+        /// <summary>
+        /// Number of hotbar slots a player has.
+        /// Default value is 10. If you change the default value (e.g. 16 slots), make sure to properly
+        /// modify the Client's HotbarWindow layout at: 'resources\gui\layouts\game\HotbarWindow.json'.
```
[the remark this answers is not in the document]
Arufonsu: Done 👍🏽

Arufonsu:
https://github.com/Arufonsu/Intersect-Engine/compare/7cd43a8..Moves_MaxHotBar_Setting?diff=unified

lodicolo (approved):

PR #934, opened 2021-09-11
Add Resource Interface so client plugins can figure out the base resource

Cheshire92:
Just a little something I ran into while trying to make the minimap plugin distinguish between resource types. :)

### DIFF
--- a/Intersect.Client/Entities/Resource.cs
+++ b/Intersect.Client/Entities/Resource.cs
@@ -15,14 +15,14 @@ using Intersect.Network.Packets.Server;
 namespace Intersect.Client.Entities
 {
 
-    public partial class Resource : Entity
+    public partial class Resource : Entity, IResource
     {
 
         private bool _waitingForTilesets;
 
-        public ResourceBase BaseResource;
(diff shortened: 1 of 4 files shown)


Intersect.Client/Entities/Resource.cs
```diff
@@ -15,14 +15,14 @@
...
+    public partial class Resource : Entity, IResource
     {
 
         private bool _waitingForTilesets;
 
-        public ResourceBase BaseResource;
+        public ResourceBase BaseResource { get; set; }
 
-        public bool IsDead;
+        public bool IsDead { get; set; }
```
lodicolo, line 25: Should this be exposed by the interface?

Cheshire92: I figured it could be?

Was thinking someone might want to figure out if a resource is consumed to render something on top of it (ie the tree falling over) or do something else with said information. Maybe render a shine over it when it's not consumed but stop rendering that when it's consumed?

Just a few random use cases.

lodicolo (approved):

PR #938, opened 2021-09-13
[Fix]Chatbox (Blank|Empty)Msgs Behaviour with SpamCheck

Arufonsu:
### The Issue
I believe that the chat log's warnings towards spammers are being checked in the wrong moment / place, this is the current behaviour with blank/empty messages, which in my opinion doesn't makes much sense if no message is being send at all:

(Simply press the "Return"/"Enter" Key repeatedly to toggle on and off the chatbox input)

https://user-images.githubusercontent.com/17498701/133119051-095f0b57-d3e4-4a0c-94b1-cfd544cfcb76.mp4

### The Fix
This PR fixes that client side behaviour, in adition with a server side check in the packet handler, to basically do the same (nothing) if the message content is empty/blank or null (just in case that someone decides to mess with the client in order to send spammy empty messages):

https://user-images.githubusercontent.com/17498701/133118893-74ad6e2b-0c75-48dd-9f10-986346aceab8.mp4

https://user-images.githubusercontent.com/17498701/133119377-97e416e7-e61e-465c-becd-14c91b5f3801.mp4

### DIFF
--- a/Intersect.Client/Interface/Game/Chat/Chatbox.cs
+++ b/Intersect.Client/Interface/Game/Chat/Chatbox.cs
@@ -15,6 +15,7 @@ using Intersect.Client.Networking;
 using Intersect.Configuration;
 using Intersect.Enums;
 using Intersect.Localization;
+using Intersect.Utilities;
 
 namespace Intersect.Client.Interface.Game.Chat
 {
@@ -372,25 +373,26 @@ namespace Intersect.Client.Interface.Game.Chat
 
         void TrySendMessage()
         {
-            if (mLastChatTime > Globals.System.GetTimeMs())
-            {
-                ChatboxMsg.AddMessage(new ChatboxMsg(Strings.Chatbox.toofast, Color.Red, ChatMessageType.Error));
-                mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
-
-                return;
-            }
-
-            mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
-
-            if (mChatboxInput.Text.Trim().Length <= 0 || mChatboxInput.Text == GetDefaultInputText())
+            var msg = mChatboxInput.Text.Trim();
+            if (string.IsNullOrWhiteSpace(msg) || string.Equals(msg, GetDefaultInputText(), StringComparison.Ordinal))
             {
                 mChatboxInput.Text = GetDefaultInputText();
 
                 return;
             }
+            
+            if (mLastChatTime > Timing.Global.MillisecondsUtc)
+            {
+                ChatboxMsg.AddMessage(new ChatboxMsg(Strings.Chatbox.toofast, Color.Red, ChatMessageType.Error));
+                mLastChatTime = Timing.Global.MillisecondsUtc + Options.MinChatInterval;
+
+                return;
+            }
+
+            mLastChatTime = Timing.Global.MillisecondsUtc + Options.MinChatInterval;
 
             PacketSender.SendChatMsg(
-                mChatboxInput.Text.Trim(), byte.Parse(mChannelCombobox.SelectedItem.UserData.ToString())
+                msg, byte.Parse(mChannelCombobox.SelectedItem.UserData.ToString())
             );
 
             mChatboxInput.Text = GetDefaultInputText();
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -734,6 +734,12 @@ namespace Intersect.Server.Networking
 
             var msg = packet.Message;
             var channel = packet.Channel;
+            
+            if (string.IsNullOrWhiteSpace(msg))
+            {
+                return;
+            }
+            
             if (client?.User.IsMuted ?? false) //Don't let the tongueless toxic kids speak.
             {
                 PacketSender.SendChatMsg(player, client?.User?.Mute?.Reason, ChatMessageType.Notice);


Cheshire92:
So this is still sending blank messages to the server?
4

Arufonsu:
> 
> 
> So this is still sending blank messages to the server?

It shouldn't, git's diff is odd on this uhh, its probably easier to read here : https://diffy.org/diff/d52bbdfeb4cdf

i simply moved up the check for empty/blank messages, which for some reason was being checked after the spawm warning.
and just in case that someone *modifies* the client to avoid this and send spammy blank/white messages, i added this:
https://github.com/AscensionGameDev/Intersect-Engine/blob/45a122c6ef32e43e2c195a480884d73c9d16f94f/Intersect.Server/Networking/PacketHandler.cs#L738

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -734,6 +734,13 @@ public void HandlePacket(Client client, ChatMsgPacket packet)
 
             var msg = packet.Message;
             var channel = packet.Channel;
+
+            // Don't handle anything if Msg is empty or null.
+            if (msg.Trim().Length <= 0)
```
lodicolo, line 739: `string.IsNullOrWhiteSpace()`

Intersect.Client/Interface/Game/Chat/Chatbox.cs
```diff
@@ -372,23 +372,23 @@ void ChatBoxSendBtn_Clicked(Base sender, ClickedEventArgs arguments)
...
+            if (mLastChatTime > Globals.System.GetTimeMs())
             {
-                mChatboxInput.Text = GetDefaultInputText();
+                ChatboxMsg.AddMessage(new ChatboxMsg(Strings.Chatbox.toofast, Color.Red, ChatMessageType.Error));
+                mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
 
                 return;
             }
 
+            mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
```
lodicolo, line 390: `Timing.Global`?

Intersect.Client/Interface/Game/Chat/Chatbox.cs
```diff
@@ -372,23 +372,23 @@ void ChatBoxSendBtn_Clicked(Base sender, ClickedEventArgs arguments)
 
         void TrySendMessage()
         {
-            if (mLastChatTime > Globals.System.GetTimeMs())
+            if (mChatboxInput.Text.Trim().Length <= 0 || mChatboxInput.Text == GetDefaultInputText())
```
lodicolo, line 375: - Copy `.Text` into a local variable
- Then check with `string.IsNullOrWhiteSpace()`
- Then check with `string.Equals(<var>, GetDefaultInputText(), StringComparison.Ordinal)`

lodicolo (changes requested):

Intersect.Client/Interface/Game/Chat/Chatbox.cs
Arufonsu: whops 🤣
 fixed.

Intersect.Client/Interface/Game/Chat/Chatbox.cs
```diff
@@ -372,27 +372,24 @@ void ChatBoxSendBtn_Clicked(Base sender, ClickedEventArgs arguments)
...
-                mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
+                mChatboxInput.Text = GetDefaultInputText();
 
                 return;
             }
 
-            mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
-
-            if (mChatboxInput.Text.Trim().Length <= 0 || mChatboxInput.Text == GetDefaultInputText())
+            if (mLastChatTime > Timing.Global.Milliseconds)
```
lodicolo, line 383: UTC

Intersect.Client/Interface/Game/Chat/Chatbox.cs
```diff
@@ -372,27 +372,24 @@ void ChatBoxSendBtn_Clicked(Base sender, ClickedEventArgs arguments)
...
+                mLastChatTime = Timing.Global.Milliseconds + Options.MinChatInterval;
 
                 return;
             }
 
-            PacketSender.SendChatMsg(
-                mChatboxInput.Text.Trim(), byte.Parse(mChannelCombobox.SelectedItem.UserData.ToString())
-            );
-
+            mLastChatTime = Timing.Global.Milliseconds + Options.MinChatInterval;
```
lodicolo, line 391: UTC

lodicolo (changes requested):

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -734,6 +733,13 @@ public void HandlePacket(Client client, ChatMsgPacket packet)
 
             var msg = packet.Message;
             var channel = packet.Channel;
+
+            // Don't handle anything if Msg is empty or null.
+            if (string.IsNullOrWhiteSpace(msg.Trim()))
```
lodicolo, line 738: `Trim()` is useless here

lodicolo (changes requested):

Intersect.Client/Interface/Game/Chat/Chatbox.cs
Arufonsu: Timing.Global.MillisecondsUtc  (?)
what about the one in L:386 ?

Intersect.Server/Networking/PacketHandler.cs
Arufonsu: got it 👍🏽

Intersect.Client/Interface/Game/Chat/Chatbox.cs
```diff
@@ -372,27 +372,24 @@ void ChatBoxSendBtn_Clicked(Base sender, ClickedEventArgs arguments)
...
             }
 
-            mLastChatTime = Globals.System.GetTimeMs() + Options.MinChatInterval;
-
-            if (mChatboxInput.Text.Trim().Length <= 0 || mChatboxInput.Text == GetDefaultInputText())
+            if (mLastChatTime > Timing.Global.Milliseconds)
             {
-                mChatboxInput.Text = GetDefaultInputText();
+                ChatboxMsg.AddMessage(new ChatboxMsg(Strings.Chatbox.toofast, Color.Red, ChatMessageType.Error));
+                mLastChatTime = Timing.Global.Milliseconds + Options.MinChatInterval;
```
Arufonsu, line 386: @lodicolo UTC as well here (?)

Arufonsu:
My bad, sorry for the commit naming mistake that referenced something that's not related at all.
should been something like "2nd review changes" instead.

PR #952, opened 2021-10-12
chore: Simplify and Cache Entity / Paperdoll rendering

### LINKED ISSUE #929: Paperdoll Drawing logic should be consolidated.
Paperdoll drawing is handled (in some cases, slightly differently) in multiple areas across the project. It would be nice if at least some of that logic could be abstracted away, so that making modifications to paperdoll rendering is less of a hassle moving forward.

As it stands, the following places need to know how to render an entity's paperdolls, and do so separately of one another:

- Entity#Draw
- SelectCharacterWindow#UpdateDisplay
- CharacterWindow#Update
- EntityBox#Update

Perhaps an entity should know how to draw itself, with its paperdolls - the other places all must have a hook into the entity in order to get accurate display information, anyway.

The trick as I've seen it is that each area needs to render things slightly differently - I.E, the interface areas need to create Gwen image elements, and the game rendering areas need to draw to the game window.

Perhaps a solution would be some sort of method on an entity that outputs the paperdolls to some sort of simple data structure - a dictionary containing the TextureType and the string filename, maybe - that each area could use to more easily iterate over and draw.

Cheshire92:
Resolves #929 

By default we render paperdolls in various different ways..
1. On screen we render it every frame endlessly calculating what we need to render and where.
2. On the GUI we render it through endless UI elements that we need to hide/unhide on every change and its done in a slightly different way every single UI element that draws paperdolls.

This is super annoying and hard to work with.
In this PR I've made it so the client will cache textures of each animation type in memory for every entity and draw these to the screen, this way we won't be rendering and calculating the equipment and whatnot every frame for every player and we can easily re-use this texture on the GUI to render our entities in a quick easy to understand manner.

Tested this on my own system with a few characters with different equipment loads spamming them on and off and seems to work pretty well so far.

### DIFF
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -301,6 +301,8 @@ namespace Intersect.Client.Entities
 
         public Dictionary<SpriteAnimations, GameRenderTexture> TextureCache { get; set; } = new Dictionary<SpriteAnimations, GameRenderTexture>();
 
+        IReadOnlyDictionary<SpriteAnimations, GameRenderTexture> IEntity.TextureCache => TextureCache;
+
         public IMapInstance MapInstance => Maps.MapInstance.Get(MapId);
 
         public virtual Guid MapId { get; set; }
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -235,41 +235,44 @@ namespace Intersect.Client.Entities
             set => mDir = (byte)((value + 4) % 4);
         }
 
-        public virtual string TransformedSprite
-        {
-            get => mTransformedSprite;
-            set
-            {
-                if (mTransformedSprite != value)
-                {
-                    mTransformedSprite = value;
-                    if (value == "")
-                    {
-                        Texture = Globals.ContentManager.GetTexture(Framework.Content.TextureType.Entity, mMySprite);
-                        LoadAnimationTextures(mMySprite);
-                    }
-                    else
-                    {
-                        Texture = Globals.ContentManager.GetTexture(Framework.Content.TextureType.Entity, mTransformedSprite);
-                        LoadAnimationTextures(mTransformedSprite);
-                    }
-                }
-            }
-        }
-
         public virtual string Sprite
         {
-            get => mMySprite;
+            get
+            {
+                var sprite = mMySprite;
+
+                //If the entity has transformed, apply that sprite instead.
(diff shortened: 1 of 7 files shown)


Intersect.Client/Entities/Entity.cs
```diff
@@ -1016,69 +1014,90 @@ public virtual void Draw()
...
+                            {
+                                itemId = Equipment[equipSlot];
+                            }
 
-                                var item = ItemBase.Get(itemId);
-                                if (item != null)
-                                {
-                                    if (Gender == 0)
-                                    {
-                                        DrawEquipment(item.MalePaperdoll, renderColor.A);
```
Arufonsu, line 1061: Does "renderColor" cover this part? if you look back at this previous implementation, equipment wasn't being rendered properly with it's defined RBGA settings, but only with alpha for some reason.

Around last week, i tried to modify the DrawEquipment function to cover up this issue, by doing something like this:
imagen

Of course, this new logic that caches what's being rendered **is way better right now, thanks a lot for this!**, back to my question, do this PR allows it properly render equipment with their colour settings? if so, that's great! and probably should be pointed out in the PR info as well, also, not sure if there's any issue opened for this right now, but is something that may be worth mentioning.

Cheshire92: I believe I fixed the render colours when I found out it was broken along the way here yes. Seeing how it's cached as one image and reused it should draw the same anywhere now.

jcsnider:
I like the cleanup, and reduction of complicated logic.

Render Textures are something I would like to keep to a minimum as much as possible though. Due to their nature they require additional video memory per entity (so a lot more for bigger games, even if they end up containing the same sprite config) and it prevents any opportunity for sprite batching to occur when the same entities are scattered all over the map.

I guess my request would be to use the render textures but only when they are being displayed in the UI (which should only occur for the current logged in player, and whoever is being targeted).

To keep the amount of duplicated logic to a minimum we could add a 'targetTexture' parameter to the draw function -- if that parameter is passed as null then it would draw to the game screen, and otherwise it would draw to the provided render texture for the ui element maybe?

Cheshire92:
Hmm, that might require some rethinking of the logic and add a flag to the entities to mark them as rendered to the GUI so the game knows to create and render their texture properly.

I understand your concern though. Will have to look into that when I find the time and see what I can do to salvage it. Lol

Cheshire92:
Marking as draft for now, mostly because I'll probably have to start over with this as a base concept.

Issue #940, opened 2021-09-16
Target influencing projectile


WeylonSantana:
**Describe the bug**
If the map is safe, you cannot fire projectiles if you click on a player (put the player on target)

**Version Encountered**
0.7.0.126

**To Reproduce**
1. Make a projectile and safe map
2. put other player in map
3. select that player and try fire a projectile

**Expected behavior**
![image](https://user-images.githubusercontent.com/63019821/133623445-4c04a11c-57bf-45da-b275-32741a6ec54a.png)
Target should not influence projectile shooting, on the safe map just do not cause damage

**Screenshots**

**Logs**
If there are error logs please attach them below.

**Operating System:**
Windows

**Additional context**

Arufonsu:
There's a way to do this, it involves to move the "isSafe" check that's present around the targeting code, down to the part where you actually *attack*. if the map is not safe, only then there will be a damage output. This would allow the ability to even shot to targets (while still avoiding stealthed ones), while preventing the damage output when required.
i can try opening up a PR for this one in the following weeks, unless someone else tackles it down.
may be worth to mention that, moving the check over to the attack handler, may not be what everyone actually wants as behaviour? we may need more opinions on this one, in the other hand, we could simply create an option that toggles whether the dev wants to check for safe areas when targeting, or when attacking is about to do it's thing (attack).

PR #973, opened 2021-10-19
Chore: clean up CanCastSpell on Entity and Player.

### LINKED ISSUE #972: condition player can attack and AoE
**Describe the bug**
NPCs take damage from aoe despite the Player can attack condition not being met.

**Version Encountered**
0.7.0.126

**To Reproduce**
1. Go to Npc editor
2. Put a variable in condition of player can attack.
3. Use AoE spell

**Screenshots**
image
image

**Operating System:**
Windows

Cheshire92:
Resolves #972  #940 

No idea why the spell system in general is in as sore a state as it is right now, but I figured I'd start cleaning SOMEWHERE.

Entities, Npcs and Players each had their own ways of checking the same things.
This really shouldn't be the case, generic checks should be done on the Entity class and Npc and Player specific checks on their respective classes.

I've tried to reduce it down to something resembling some form of coherence, got rid of the out of place checks on Npc and Player classes and replaced them with inherited methods, an override and some semblance of logic.

Also made sure all entities have to pass the same basic checks before allowing them to cast spells, Entities other than  players were allowed to bypass certain checks entirely.

Also fixes the following bug(s):
- AoE spells ignore attack conditions.
- Targets could influence Projectile casts.

### DIFF
--- a/Intersect (Core)/Enums/SpellCastFailureReason.cs
+++ b/Intersect (Core)/Enums/SpellCastFailureReason.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Intersect.Enums
+{
+    public enum SpellCastFailureReason
+    {
+        None,
+
+        InvalidSpell,
+
+        InsufficientHP,
+
+        InsufficientMP,
+
+        InvalidTarget,
+
+        Silenced,
+
+        Stunned,
+
+        Asleep,
+
+        InvalidProjectile,
+
+        InsufficientItems,
+
+        Snared,
+
+        OutOfRange,
+
+        ConditionsNotMet,
+
+        OnCooldown
+    }
+}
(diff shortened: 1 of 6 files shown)


Cheshire92:
Because someone didn't feel like posting the issue here:
SirTodd
Auto attacking players is causing dc
It does, the issue is on Intersect.Server.Entities.Player.CanAttack, at line 1351
But I'll fix myself

naziozeno:
I found a bug.
When you try use an AOE skill in friendly mode with any type of buff, for example: Invulnerable, and have a target locked, you can't use that, it's say "Invalid target for this spell".

Cheshire92:
> When you try use an AOE skill in friendly mode with any type of buff, for example: Invulnerable, and have a target locked, you can't use that, it's say "Invalid target for this spell".
This should be fixed now.

> Auto attacking players is causing dc
> It does, the issue is on Intersect.Server.Entities.Player.CanAttack, at line 1351
Unable to replicate this issue in any meaningful way as attacking works fine for me? Need more information.

Arufonsu (approved):

lodicolo (approved):

PR #989, opened 2021-10-22
Forcibly reset Npcs on a map once we stop processing it

### LINKED ISSUE #988: Npcs get stuck at the entrance / exit of areas when all players leave
**Describe the bug**
When an Npc chases a player around and they leave the map with no players remaining on the map they remain in position until someone comes back on the map sometimes not moving anymore and getting stuck or instantly attacking the player that just joined the map due to their aggression setting.

**Version Encountered**
Any, really.

**Expected behavior**
All Npcs should reset appropriately when we stop processing a map.

Cheshire92:
Resolves #988 

This should stop Npcs from hugging corners, entrances, walls etc. when a player joins a map somebody managed to lure all Npcs into one corner for.

When a reset radius is allowed and configured, the Npc will simply reset to the location it was pulled from, assuming it has one.
When no reset radius is configured or allowed.. things get a little trickier and we check if the Npc is in combat or not, if it is we forcibly kill it dropping no items and respawn it in its designated spawn location or randomly.

### DIFF
--- a/Intersect.Server/Core/LogicService.LogicThread.cs
+++ b/Intersect.Server/Core/LogicService.LogicThread.cs
@@ -155,8 +155,8 @@ namespace Intersect.Server.Core
                                     var mapInstance = MapInstance.Get(map);
                                     if (mapInstance != null)
                                     {
-                                        mapInstance.DespawnEverything();
-                                        mapInstance.RespawnEverything();
+                                        mapInstance.DespawnNpcs();
+                                        mapInstance.SpawnMapNpcs();
                                     }
                                 }
                             }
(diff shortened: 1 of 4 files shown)


naziozeno:
Bug found.
When you die for npc or be out of sight, it can disappear, just despawn, but when you press tab it's seens be at respawn, you can see the entity box for some ms.

Cheshire92:
I believe that's an issue with the engine in general, not this change as I've seen blank entities around at times before. Just never found the cause of it yet.

Arufonsu (approved):

lodicolo (approved):

PR #1000, opened 2021-10-26
Resolve #995 - Add the ability to make NPCs immune to knockback; bonus effects

### LINKED ISSUE #995: Allow certain enemies to be immune to knockback, and other status effects
Knockback is a cool feature, but it can be overpowered during something like a boss fight.

The same can be said for things like stun, snare, and sleep.

Allow config on NPCs to determine which status effects/knockback they are immune to. Could even be a percent value.

**Additional context**

Daywalkrr:
Resolves #995 

This adds the ability in the editor to have an NPC be immune to negative-bonus effects, such as stun or silence, but also to Knockback.

This feature also allows an NPC to have a "Tenacity" attribute, so that they can still be affected by these effects, but at a lesser rate than other enemies.

There are expansion opportunities on this system that I've not implemented - as a whole, I would understand if we would like to wait on this until spell effects/statuses have been re-worked, maybe with a "Effect" editor and the like. But, barring that, this is a solution that I'm proposing.

The menu looks like this, within the NPC Editor:

image

### DIFF
--- a/Intersect (Core)/GameObjects/NpcBase.cs
+++ b/Intersect (Core)/GameObjects/NpcBase.cs
@@ -39,11 +39,7 @@ namespace Intersect.GameObjects
             get => JsonConvert.SerializeObject(Immunities);
             set
             {
-                Immunities = JsonConvert.DeserializeObject<List<StatusTypes>>(value ?? "");
-                if (Immunities == null)
-                {
-                    Immunities = new List<StatusTypes>();
-                }
+                Immunities = JsonConvert.DeserializeObject<List<StatusTypes>>(value ?? "") ?? new List<StatusTypes>();
             }
         }
 


lodicolo (changes requested):
I have no issue with the feature itself, but we have two improperly named enums.

`EffectType` is Item-only effects.

`StatusTypes` is Spell-only effects.

Both of them are named wrong, but correctly their names right now would alter far too much of the code base. Renaming them should be done in Beta 8, but this PR (if my other comments are addressed), is perfectly ok to go into 7.1.

The actual enum names should be `ItemEffect` and `SpellEffect`, and then "Immunity to a SpellEffect" and "Knockback is a SpellEffect" suddenly makes sense.

Intersect (Core)/GameObjects/NpcBase.cs
```diff
@@ -12,6 +12,48 @@
...
+        Transform,
+        /// <summary>
+        /// Whether the NPC can be affected by sleep
+        /// </summary>
+        Sleep,
+        /// <summary>
+        /// Whether the NPC can be affected by taunt
+        /// </summary>
+        Taunt
+    }
```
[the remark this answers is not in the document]
AlexVild: Done

Intersect.Server/Entities/Entity.cs
```diff
@@ -2681,6 +2699,40 @@ public StatusPacket[] StatusPackets()
...
+                case StatusTypes.Snare:
+                    return Immunities.Snare;
+                case StatusTypes.Taunt:
+                    return Immunities.Taunt;
+                case StatusTypes.Transform:
+                    return Immunities.Transform;
+                default:
+                    return Immunities.None;
+            }
+        }
```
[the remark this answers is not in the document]
AlexVild: Done

Intersect.Server/Intersect - Backup.Server.csproj
```diff
@@ -0,0 +1,526 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
lodicolo, line 1: Don't commit this

Intersect.Server/Localization/Strings.cs
```diff
@@ -368,6 +368,9 @@ public readonly LocalizedString
                 }
             );
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString immunetoeffect = @"IMMUNE!";
```
lodicolo, line 372: `ImmuteToEffect` and honestly I'd say `Immune!` by default

Intersect (Core)/GameObjects/NpcBase.cs
```diff
@@ -30,6 +29,24 @@ public class NpcBase : DatabaseObject<NpcBase>, IFolderable
 
         [NotMapped] public int[] VitalRegen = new int[(int) Vitals.VitalCount];
 
+        [NotMapped]
+        public Dictionary<StatusTypes, bool> Immunities = new Dictionary<StatusTypes, bool>();
```
lodicolo, line 33: I didn't realize this before but this could totally just be a list of immunities. If it's not in the list, it's not immune. Checkboxes for immunities can just add/remove the immunity from the list.

Caveat: when serialized/checked for differences the immunities should be sorted (by name or enum-erical value, I don't care, just so long as they are consistent so that even if someone unchecked and rechecks all the same things in a different order it doesn't consider it a change).

lodicolo (changes requested):

Intersect.Editor/Localization/Strings.cs
```diff
@@ -4020,6 +4020,25 @@ public struct NpcEditor
...
+            public static Dictionary<string, LocalizedString> Immunities = new Dictionary<string, LocalizedString>
+            {
+                {"knockback", @"Knockback"},
+                {"silence", @"Silence"},
+                {"stun", @"Stun"},
+                {"snare", @"Snare"},
+                {"blind", @"Blind"},
+                {"transform", @"Transform"},
+                {"sleep", @"Sleep"},
+                {"taunt", @"Taunt"},
```
lodicolo, line 4039: The keys should just be the status type enum

lodicolo (changes requested):

Intersect (Core)/GameObjects/NpcBase.cs
AlexVild: Now a list

Intersect (Core)/GameObjects/NpcBase.cs
```diff
@@ -30,6 +29,24 @@ public class NpcBase : DatabaseObject<NpcBase>, IFolderable
...
+        [Column("Immunities")]
+        public string ImmunitiesJson
+        {
+            get => JsonConvert.SerializeObject(Immunities);
+            set
+            {
+                Immunities = JsonConvert.DeserializeObject<List<StatusTypes>>(value ?? "");
+                if (Immunities == null)
+                {
+                    Immunities = new List<StatusTypes>();
```
lodicolo, line 45: `?? new List` at the end of the Deserialize line, no reason to have an if statement for this

Intersect.Server/Intersect - Backup.Server.csproj
AlexVild: Whoops, removed

lodicolo (changes requested):

Intersect (Core)/GameObjects/NpcBase.cs
AlexVild: Done

PR #1002, opened 2021-10-27
Resolve #994: Add a "check bank" checkbox to the "Has Item" event conditional.

### LINKED ISSUE #994: An event that can check a player's inventory AND bank for some item
The engine is not 100% guaranteed to keep players items and stuff - while it is pretty darn good these days, rollbacks _can_ happen, among other strange quirks and glitches, and items can be lost.

Sometimes, a dev may want to add a Quest Item to their game - some item that is "bound", more or less, to a player and is maybe given through some event that the player can only access once.

Another example is a quest reward - maybe by doing some quest, you unlock a bound grappling hook, but through some sort of engine glitch, that grappling hook is lost. The player would have no way of getting it back, unless the dev included some sort of way for the event to give the player the item again - but currently, that event only has the power to look at a player's _inventory_ to check if they don't have the item.

Since these items are BOUND, having a dev replenish them means that, currently, a player could, theoretically, fill up their bank and inventory with bound items that they have no way of destroying.

A protection against this would be to check to see if a player owns, ANYWHERE, some item. So, just a new condition that can check both a player's inventory and a bank. This does not protect against a player dropping an item and starting the event, then picking up the dropped item - but devs can already control which items can be dropped or not.

**Additional context**

AlexVild:
Resolves #994

Adds a lil' check box to the Has Item event conditional that allows the conditional to check into a user's bank as well as their inventory. Special shoutouts to whoever made the "CountItems" method.

Looks like this in-editor:

image


Intersect.Editor/Localization/Strings.cs
```diff
@@ -2181,6 +2181,9 @@ public struct EventConditional
 
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public static LocalizedString MapZoneTypeLabel = @"Zone Type:";
+
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public static LocalizedString checkbank = @"Check Bank?";
```
lodicolo, line 2186: `CheckBank`

lodicolo (changes requested):

lodicolo (approved):

Issue #998, opened 2021-10-23
server crash when granting invalid access to players


TheGodFatherTGF:
**Version Encountered**
Version 0.7.0.126

**To Reproduce**
1. Go to your intersect server console
2. give an offline player admin access but give them an access level that is not recognized by `Intersect.Enums.Access` for example `/poweracc dummyaccount 3`
3. Server will instantly crash
4. scroll down to see error

**Expected behavior**
Complete server crash

**Screenshots**
N/A

**Logs**
```
2021-10-22 14:23:06.739 [Error] Received unhandled exception from System.Threading.Thread.
2021-10-22 14:23:06.743 [Error] 
  Message: Exception of type 'System.ArgumentOutOfRangeException' was thrown.
Parameter name: access
Actual value was 3.
    Stack:   at Intersect.Server.Database.PlayerData.Security.AccessExtensions.AsUserRights (Intersect.Enums.Access access) [0x00031] in C:\TeamCity\buildAgent\work\40dc5d27074313da\Intersect.Server\Database\PlayerData\Security\UserRights.cs:151 
  at Intersect.Server.Core.Commands.PowerAccountCommand.HandleTarget (Intersect.Server.Core.ServerContext context, Intersect.Server.Core.CommandParsing.ParserResult result, Intersect.Server.Database.PlayerData.User target) [0x00010] in C:\TeamCity\buildAgent\work\40dc5d27074313da\Intersect.Server\Core\Commands\PowerAccountCommand.cs:27 
  at Intersect.Server.Core.Commands.TargettedCommand`1[TTarget].HandleValue (Intersect.Server.Core.ServerContext context, Intersect.Server.Core.CommandParsing.ParserResult result) [0x00018] in C:\TeamCity\buildAgent\work\40dc5d27074313da\Intersect.Server\Core\Commands\TargettedCommand.cs:25 
  at Intersect.Server.Core.CommandParsing.Commands.HelpableCommand`1[TContext].Handle (TContext context, Intersect.Server.Core.CommandParsing.
```

**Operating System:**
Windows 10 (Server is intersect hosting)

**Additional context**
N/A

PR #1003, opened 2021-10-27
Fix when granting an out of range access to a player [https://github.com/AscensionGameDev/Intersect-Engine/issues/998]

lucas-tardivo:
Fixing issue #https://github.com/AscensionGameDev/Intersect-Engine/issues/998


lodicolo:
No this wasn't the correct fix, #1131 was.

PR #1011, opened 2021-11-02
Correctly link new quest task completion events and restore old ones

### LINKED ISSUE #915: Quest Editor - Task Editor is causing the Intersect Editor to crash due to null reference.
**Describe the bug**
In a project that was started from prerelease around the end of May 2021, and has since been updated to the latest prerelease as of August 16th 2021, _some_ quest events can not be edited without crashing the Intersect Editor.

Some events work fine.

For others - if a Task is double clicked to edit, will get in a bad state. Clicking on "Cancel" or "Edit Completion Event" will cause the editor to crash with the following stacktrace:

```
Message: Object reference not set to an instance of an object.
    Stack:    at Intersect.Editor.Forms.Editors.Quest.QuestTaskEditor.btnEditTaskEvent_Click(Object sender, EventArgs e) in ...Intersect-Engine\Intersect.Editor\Forms\Editors\Quest\Quest_TaskEditor.cs:line 171
```

**Version Encountered**

0.7.0.0. Most recent prerelease build as of August 16th 2021. There have been minor, isolated, source edits.

**To Reproduce**
1. My guess is - have an older project that has been upgraded to current prerelease spec
2. Modify old quest tasks
3. Some will throw the error

**Expected behavior**
Am able to modify quest tasks without crashing the editor.

**Screenshots**
Looking in the source, the problem lies here:
image

Or, wherever we're trying to reference `task.EditingEvent` - it's null, for some reason.

**Logs**
```
Message: Object reference not set to an instance of an object.
    Stack:    at Intersect.Editor.Forms.Editors.Quest.QuestTaskEditor.btnEditTaskEvent_Click(Object sender, EventArgs e) in ...Intersect-Engine\Intersect.Editor\Forms\Editors\Quest\Quest_TaskEditor.cs:line 171
```

**Operating System:**
Windows

Cheshire92:
Resolves #915 

After finding the issue in Leafling and finding out it was the likely culprit for an issue we had in base Intersect I wrote the fix supplied here.

See #915 for the full explanation.


AlexVild:
FWIW, this bug _did_ crop up again for me after testing the migration. 

I noticed it mostly when changing the _type_ of a quest objective, then saving, _then_ going back and editing a task completion event.

As always, I am happy to supply game files that have the quests in this state.

AlexVild:
The migration works to patch old task completion events, but new quests created after the migration will still have the same problem.

However, we have found out how to reproduce the issue:

1) Create a new quest
2) Give it a task
3) Leave the task completion event empty
4) Save the quest
5) Restart the server and editor
6) Go to your new quest, try to modify its completion event
7) The server will crash, as when we are loading completion events, it sees that the `QuestTask.CompletionEvent` is set to null (for reasons undefined at the moment)

AlexVild:
The issue I mentioned above has been fixed. We were not saving the object after modifiying it while handling a `SaveGameObject` packet on the server.

Arufonsu (approved):

Arufonsu (changes requested):
Seems good 👍🏽 just need to solve the merge conflicts !

lodicolo (approved):

lodicolo (approved):

PR #1019, opened 2021-12-13
Modify Asset Packing to pack by file size per batch rather than file count

Cheshire92:
Probably seems like the most pointless thing ever, but batching these by file count rather than size was always a bit of a silly oversight of mine.
Unlucky scenarios could cache like 50MB inside a single file, which takes.. a while to load.

This should let the user set an upper limit for pack file sizes.
Of course, trying to package files above the limit won't make the files magically smaller, it'll just put these in its own packs.


lodicolo (approved):

lodicolo:
@Cheshire92 Will this impact loading of existing packs made with the file count system?

Cheshire92:
It won't as it only modifies the packing method itself.
The layout and index files remain the same format. Just changed how they're populated from here on out.

Arufonsu (approved):

PR #1024, opened 2021-12-30
I need help

idm0504:
Hello I am korean.
I want to chat.
Korean is not supported properly.
You have to press the Hangul key twice to write in the text box


Cheshire92:
Please open an issue rather than a pull request for bugs.

PR #1026, opened 2021-12-30
Random monster looting

dudevelopement:
This will cause the monster to drop from 1 to the maximum set in the editor.

I'm new to the programming business, any editing will be of great help. I accept suggestions to improve this engine and bring a quality of life to everyone.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -13,4 +13,6 @@
 
 [Dog from the Kash Shop](#)
 
+[Fernando Arzola Lagos](https://github.com/Arufonsu)
+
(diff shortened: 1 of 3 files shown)


AlexVild:
A cool idea, but I think it needs some additional control. Maybe I don't want to _always_ randomize an amount.

I think a better option might be to allow the user to input a range of quantities that we randomly choose from?

Cheshire92:
I agree with @AlexVild here. This needs some level of developer control.

Cheshire92:
Closing this, given the time since the last change.

PR #1032, opened 2022-01-04
[GWEN] FIX #1031: Control Base now keeps track of sound instance it plays

### LINKED ISSUE #1031: [GWEN] Sounds played through the Gwen pipeline (clickedsound, hoversound, etc) get prematurely Garbage Collected
**Describe the bug**
The interface JSON files allow a user to sometimes attach sounds to UI events, such as clicking a button. However, playing these sounds is bugged - a user who changes a sound to something of sufficient length may realize that their sound is getting cutoff before the sound is finished.

The THEORY:
This is because when a sound is played through Gwen (through `Control/Base.cs`), we do not hold any sort of reference to the sound instance we create. As a result, the .NET GC cleans up this sound instance before the sound is actually finished playing, which triggers the Destructor of a MonogameSoundInstance to fire; prematurely disposing the Mono SoundEffectInstance.

**Version Encountered**

Prerelease - 0.7

**To Reproduce**
1. Replace some UI sound (clickedsound on a button, for example) with some much longer .wav
2. Continually click the button in-client, and you will notice the sound getting cutoff at random intervals.

**Expected behavior**
A sound that is placed on a UI element should play to completion, non-looping.

**Operating System:**
Windows

AlexVild:
Resolves #1031 

This is me attempting to implement a fix for my theory as to why UI sounds get cutoff prematurely (see linked issue for details). I _haven't_ seen anything like memory leaks or weird issues/crashes come from this change, and while debugging I did see that sound instances I was creating through the UI were being disposed of in some fashion.

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/Base.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Base.cs
@@ -131,8 +131,6 @@ namespace Intersect.Client.Framework.Gwen.Control
 
         private object mUserData;
 
-        private Dictionary<string, GameAudioInstance> mSounds;
-
         /// <summary>
         ///     Initializes a new instance of the <see cref="Base" /> class.
         /// </summary>
@@ -170,8 +168,6 @@ namespace Intersect.Client.Framework.Gwen.Control
             BoundsOutlineColor = Color.Red;
             MarginOutlineColor = Color.Green;
             PaddingOutlineColor = Color.Blue;
-
-            mSounds = new Dictionary<string, GameAudioInstance>();
         }
 
         /// <summary>
@@ -2178,15 +2174,10 @@ namespace Intersect.Client.Framework.Gwen.Control
             if (sound != null)
             {
                 var soundInstance = sound.CreateInstance();
-                // If this sound file hasn't been played by this UI element before
-                if (!mSounds.ContainsKey(filename))
+                if (soundInstance != null)
                 {
-                    // Add it to the list of playing sounds
-                    mSounds.Add(filename, soundInstance);
+                    Canvas.PlayAndAddSound(soundInstance);
                 }
-                // And play the sound
-                mSounds[filename].SetVolume(100, false);
-                mSounds[filename].Play();
             }
         }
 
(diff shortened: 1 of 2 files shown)


Arufonsu:
just gave this a try, i can confirm that it allows for longer sounds to work properly now 👍🏽 💯

Cheshire92:
What happens when a UI element plays two files simultaneously? For example hover and click? One would still get cut off as the reference is replaced I would assume.

AlexVild:
> 

A good question. Yes, the sound stops playing.

I'm wondering what would be _preferred_? If it makes sense tor a sound to get cutoff if the user is, essentially, now doing a different action - I'm imagining someone with a longer hover sound wanting the sound to stop when the player decides to click on something, for example.

_However_, from a dev standpoint, maybe that would still be too confusing. Lemme know what you think - I suspect that a fix might be just to maintain a list instead? I wonder what's a good way of identifying identical sounds, though. I didn't easily see a way to check the filename of a monosoundinstance

Cheshire92:
Personally, I would like to have the GUI play every sound that it is playing without breaking any of them off if we are taking the time to fix this. 

As for duplicate files, how about instead of a list we use a dict<string, monosoundinstance> where the string is the file name to avoid duplicates?

AlexVild:
@Cheshire92 I think I've updated this to work in the way you described, lemme know what you think.

Cheshire92:
@jcsnider we've moved the code to canvas so we're not looping for dead sounds every frame for every UI element. Since you're a little more knowledgeable about Gwen than I am, I'd like your thoughts on this before we go ahead with it though.

Arufonsu (approved):
This PR resolves #1031 , i just gave it a try and GUI element sounds fully play till completion now, no matter how many times are pressed / hovered / etc, which is way better and clean than the cutting off outcome that we were experiencing prior these changes. 👍🏽 

https://user-images.githubusercontent.com/17498701/153782795-a7fd0350-6b6c-4ffc-8547-24b8dc366da5.mp4

Cheshire92 (approved):

PR #1040, opened 2022-01-05
Chore: DistancePointToRectangle (MapSound)

Arufonsu:
This is a chore PR because it cleans up the code structure without altering the actual functionality.

This PR pretends to improve the performance and simplify the code of DistancePointToRectangle() by using variables as a 'grid system' (left, right, top, bottom) instead of several regions to calculate the distances from points to rectangles.

[Check out the split view of the diff here](https://github.com/AscensionGameDev/Intersect-Engine/pull/1040/commits/17bc7611f4d5906f64c2f76fc0bfab363e5b9f9e?diff=split&w=0)

Note: Everything has been tested already and the functionality of DistancePointToRectangle() seems to be working as intended.

### DIFF
--- a/Intersect.Client/Core/Sounds/MapSound.cs
+++ b/Intersect.Client/Core/Sounds/MapSound.cs
@@ -169,99 +169,34 @@ namespace Intersect.Client.Core.Sounds
             return distance;
         }
 
-        //Code Courtesy of  Philip Peterson. -- Released under MIT license.
-        //Obtained, 06/27/2015 from http://wiki.unity3d.com/index.php/Distance_from_a_point_to_a_rectangle
-        public static float DistancePointToRectangle(Point point, Rectangle rect)
+        private static float DistancePointToRectangle(Point p, Rectangle r)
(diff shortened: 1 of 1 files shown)


lodicolo (changes requested):
But this totally changes the results of this function? Diagonals before used sqrt(x^2+y^2) to calculate distance and now they don't.

Arufonsu:
> But this totally changes the results of this function? Diagonals before used sqrt(x^2+y^2) to calculate distance and now they don't.

Do we need to calculate diagonals specifically if we can calculate and return the required value from a simple X & Y grid?
this was more likely an _experiment/test anyway_ (which so far works well with intersect, and is shorter to implement), from the tests i did with this, the returned values are actually accurate when you get closer or farther from specific tiles/events with map sounds. If this approach isn't adequate for the expected outcome, please feel free to close this one PR 👍🏽 !

Arufonsu:
Closing and postponing this chore PR for now.

PR #1046, opened 2022-01-13
Resolve 1045: Refactor Map Processing into Instances

### LINKED ISSUE #1045: Abstract map processing out from the MapInstance class.
On the roadmap for B8 is the ability to "Instance" maps. A large degree of refactoring is required in the way we handle map processing in order for this to work.

The proposal I've heard is to strip out map processing from the `MapInstance` class to an in-memory list of layers that belongs to given map instances. These layers will contain the logic for processing map updates, and the LogicService queueing will need to be re-written to support these layers as opposed to the updating of actual `MapInstances`.

AlexVild:
Resolves #1045 

# Intro
This PR addresses the need for the `MapInstance` class to abstract away its map processing (NPC updates, entity updates, respawns, mapitems... etc) to a new class, or in this case, a dictionary of IDs to class instances.

This PR **DOES NOT** "Add Instancing" in that after this would be merged, engine-users would still not be able to warp their players to instanced maps. Parties can't be warped together, etc.

This PR _will allow for_ us to introduce those features by creating this abstraction layer for map processing.

# Table of Contents
1. [Renames](#renames)
2. [Core Concepts](#concepts)
3. [Wow, That's A Lot Of Files, Champ](#fileschanged)
    1. [The Most Common Change](#commonchange)
    2. [Why This Pattern](#whypattern)
4. [You Changed the Logic Thread](#logicthread)
5. [How Do You Clean These Up](#cleanup)
6. [Testing](#testing)

## Renames<a name="concepts"></a>
Just as an update, and important when skimming this PR message, is this:

A `MapInstance` has been renamed to a `MapController`, and it contains a _dictionary_ whose values point to `MapInstance`s, which is my new class.

## Core Concepts<a name="concepts"></a>
* A `MapController` no longer processes it's entities, NPCs, etc... because it _has_ none of those things
* * Instead, there is a new class called `MapProcessingLayer` that handles that work for us
* * A `MapInstance` will process everything that a "Map" needs to have processed such that some player can play the game.
* A `MapController` has a new private property, `mInstances` that is of type `ConcurrentDictionary<Guid, MapInstance>
* * This is a mapping of an _Instance ID_ (some GUID) to a related _MapInstance_
* Everything that belongs to a map - an Entity, a MapItem, an EventPageInstance, a Projectile... all of it, has a new **MapInstanceId** property of type `Guid`
* * **NO LONGER WILL THESE CLASSES TALK TO THE MAP CONTROLLER THEY ARE "ON"**. Rather, they now talk to the _Instance_ they are on, and the `MapInstance` that shares the same `MapInstanceId` will be what sends the requesting entity their packets/what keeps all the entities et. al. on the layer up-to-date.
* A map MUST have an instance for a player to play the game and for the game to function. As such, the "Root" layer is any layer with a `MapInstanceId` of `Guid.Empty`

## Wow, That's A Lot Of Files, Champ<a name="fileschanged"></a>
Indeed. As such, I would recommend starting off in a few places:

1) `MapController.cs` - to see what was ripped out, and to see how it can talk to it's new dictionary of MPLs
2) `MapInstance.cs` - to see how, for the most part, I simply lifted functionality from `MapInstance` and tossed that into this class.
3) `Player#Warp().cs` - this, and the player's overridden "Move" method, are the _only_ places anyone _ever_ creates a new MapInstance. A player is the only thing that can request a new instance. As such, the warp method is a good starting point to see what all needs to be accounted for when we _go_ to a new instance.

### You will see one change in particular a LOT<a name="commonchange"></a>
There are a lot of places we used to get some entity's map via something like this:
```cs
var map = MapInstance.Get(this.MapId); // Note the old naming - this has been renamed to MapController in my PR
```

This will no longer work, as we need to get the entity's _Processing Layer_ that _exists_ on their current Map.

As such, you will see the above code turn into something like this:
```cs
if (MapController.TryGetInstanceOnMap(this.MapId, this.MapInstanceId, out MapInstance mapInstance)) {
...
```
A **LOT**. So, just be warned that that's _everywhere_, and all it is doing is _instead of going to the MapController, we now need to the MapInstance ON the MapController that pertains to our entity_.

### Why Did I Choose This Pattern For Getting An Instance?<a name="whypattern"></a>
I went with the `Try..(out var)` pattern because it will give us an easy, concise way of having null safety. Rather than getting/storing a `MapInstance` directly, we can instead wrap references to anywhere we need to refer to a `MapInstance` in an if that checks to see if we _can_ get the instance at that time.

Whenever we can't, we simply don't go inside that if. I really don't like this pattern:
```cs
if (mapInstance != null)
```
as it's ugly and forces a developer to _remember_ to check, rather than have the method call explicitly check for null.

## You Changed the Logic Thread<a name="logicthread"></a>
I did. The logic thread was previously in charge of telling `MapInstance`s to update themselves. This had to change.

A MapController no longer even _can_ update itself. So, the logic in the LogicThread has changed such that we are now getting all of the `MapInstance`s (and their neighbors) that players are occupying, and call _their_ update methods.

They also need cleaned up, which...

## How do we clean these up?<a name="cleanup"></a>
We _must_ keep around instances in memory **even after all of their neighboring players have left**. Consider a player who dies on a map, and then leaves that map "orphaned" - with no players around to see it. The player drops their equipment on death.

If we _didn't_ keep a reference to the processing layer around and cleaned it up (like how Intersect used to handle map cleanups), the player would lose their items forever - regardless of an item's despawn timer.

This is because a `MapController` used to have a list of items in memory - and we never cleaned up `MapController`s - they always existed as long as a server existed, or we refreshed the map in-editor. However, _now_, Instances handle processing, and an Instance would have a reference to the dropped items.

The solution proposed here is that we keep MapInstances around based off a setting in the Server configuration. However, if there are no players on a layer or it's neighbors, we simply exit the `Update()` method of the instance early, saving CPU cycles, but costing additional memory. 

**Tl;Dr**: If there are no players on the map layer or it's neighbors, AND a configurable amount of time has passed since the map was last populated with anybody, we will clean up the layer.

## How Can I Test Your Work?<a name="testing"></a>
Simply pull down this PR into your own fork, and make a modification to `Player#Warp()` such that whenever you admin warp, you change the Player's InstanceLayer. From the line `PreviousMapInstanceId = MapInstanceId;`, add this:

```cs
if (adminWarp)
{
    if (MapInstanceId == Guid.Empty)
    {
        MapInstanceId = Guid.NewGuid();
    }
    else
    {
        MapInstanceId = Guid.Empty;
    }
}
```

Then, build the project, sign into your client as an admin, and Shift + Right Click to warp around. Each warp will alternate between requesting a new instance, and going back to the "Root" instance (Guid.Empty)

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -14,7 +14,7 @@
 
   <PropertyGroup Condition="'$(IntersectProjectName)' == 'Network' AND '$(Configuration.ToLower())' != ''">
     <GenerateEachBuild Condition="'$(INTERSECT_GENERATE_EACH_BUILD)' != ''">true</GenerateEachBuild>
-    <IntersectBuildingAssemblyFile>..\..\bin\building\Debug\net462\Intersect.Building.dll</IntersectBuildingAssemblyFile>
+    <IntersectBuildingAssemblyFile>$(IntersectRepoPath)\bin\building\Debug\net462\Intersect.Building.dll</IntersectBuildingAssemblyFile>
   </PropertyGroup>
 
   <PropertyGroup>


Intersect (Core)/Config/Options.cs
```diff
@@ -149,6 +149,8 @@ public class Options
         public static int TileWidth => Instance.MapOpts.TileWidth;
 
         public static int TileHeight => Instance.MapOpts.TileHeight;
+        
+        public static int TimeUntilMapCleanup => Instance.MapOpts.TimeUntilMapCleanup;
```
Cheshire92, line 153: We're going to get rid of top level options.
Can you remove this and refer to it through the appropriate means? (Can't recall if Options.Instance.SubCategory or Options.SubCategory was preferred, may need to read recent changes)

Intersect (Core)/Network/Packets/Server/MapLayerChangedPacket.cs
```diff
@@ -0,0 +1,25 @@
+﻿using MessagePack;
+using System;
+using System.Collections.Generic;
+
+namespace Intersect.Network.Packets.Server
+{
+    [MessagePackObject]
+    public class MapLayerChangedPacket : IntersectPacket
```
Cheshire92, line 8: This is still called MapLayer after your other changes regarding naming conventions.

Intersect.Client/Networking/PacketHandler.cs
```diff
@@ -381,6 +381,33 @@ public void HandlePacket(IPacketSender packetSender, MapEntitiesPacket packet)
             }
         }
 
+        // MapLayerChanged Packet (for traveling between map "instances")
```
Cheshire92, line 384: Same here, inconsistent name after your changes.

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+
+        //Temporary Values
+        private Guid[] mSurroundingMapIds = new Guid[0];
+        private Guid[] mSurroundingMapsIdsWithSelf = new Guid[0];
+        private MapController[] mSurroundingMaps = new MapController[0];
+        private MapController[] mSurroundingMapsWithSelf = new MapController[0];
+
+        [JsonIgnore]
+        [NotMapped]
+        public Guid[] SurroundingMapIds
```
Cheshire92, line 57: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+                    var surroundingMapsIdsWithSelf = new List<Guid>(value);
+                    surroundingMapsIdsWithSelf.Add(Id);
+                    mSurroundingMapsIdsWithSelf = surroundingMapsIdsWithSelf.ToArray();
+                }
+            }
+        }
+
+        [JsonIgnore]
+        [NotMapped]
+        public MapController[] SurroundingMaps
```
Cheshire92, line 75: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+
+        //For New Maps!
+        public MapController(bool newMap = false) : base(Guid.NewGuid())
+        {
+            Name = "New Map";
+            Layers = null;
+        }
+
+        [JsonConstructor]
+        public MapController(Guid id) : base(id)
```
Cheshire92, line 106: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+            {
+                return;
+            }
+
+            Layers = null;
+        }
+
+        public new static MapControllers Lookup => sLookup = sLookup ?? new MapControllers(MapBase.Lookup);
+
+        public static MapController Get(Guid id)
```
Cheshire92, line 118: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+        {
+            if (id == Guid.Empty)
+            {
+                return;
+            }
+
+            Layers = null;
+        }
+
+        public new static MapControllers Lookup => sLookup = sLookup ?? new MapControllers(MapBase.Lookup);
```
Cheshire92, line 116: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+
+        public static MapController Get(Guid id)
+        {
+            return Lookup.Get<MapController>(id);
+        }
+
+        public static bool TryGetInstanceFromMap(Guid mapControllerId, Guid instanceId, out MapInstance mapInstance, bool createIfNew = false)
```
Cheshire92, line 123: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+            {
+                return mapInstance != null;
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        public static List<MapInstance> GetSurroundingMapInstances(Guid mapControllerId, Guid instanceId, bool includeSelf, bool createIfNew = false)
```
Cheshire92, line 137: Documentation?

Intersect.Server/Maps/MapController.cs
```diff
@@ -0,0 +1,540 @@
...
+                        instances.Add(instance);
+                    }
+                }
+            }
+            return instances;
+        }
+
+        //GameObject Functions
+
+        public object GetMapLock()
```
Cheshire92, line 156: Documentation?
(I think you get the idea for the rest of the file.. there's a lot of methods :P)

Cheshire92 (commented):
Read through it for a bit, lots of small moved this from here to there.. Oof.
Doesn't seem too terrible?
If we're renaming MapInstance we should probably have documentation there as well so we don't have to touch it up again after.

Intersect (Core)/Config/Options.cs
AlexVild: Done

Intersect (Core)/Network/Packets/Server/MapLayerChangedPacket.cs
AlexVild: Done

Intersect.Client/Networking/PacketHandler.cs
AlexVild: Done

Intersect.Server/Maps/MapController.cs
AlexVild: Documentation added

PR #1048, opened 2022-01-14
Change Numerical Input field values back-end storage to Double type

### LINKED ISSUE #1047: [GWEN] Numerical Textboxes and Sliders can't handle Int32.MaxValue
**Describe the bug**
Because of how Floats work (the internal storage type for the values), assigning Int32.MaxValue to a Numerical Textbox and / or Slider component the actual value is lost to the ether.
Since we can stack items up to the Int32 cap we should really accept this as an input.

**Expected behavior**
The value should be retained, may need to use a double to store this?

Cheshire92:
Resolves #1047

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/NumericUpDown.cs
+++ b/Intersect.Client.Framework/Gwen/Control/NumericUpDown.cs
@@ -73,7 +73,7 @@ namespace Intersect.Client.Framework.Gwen.Control
         /// <summary>
         ///     Numeric value of the control.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Arufonsu:
👍🏽  !
"Double is costlier, occupies more space and is more effective when more precision is required. For example, currency conversion, financial reports and transactions, scientific calculations etc…" [Source](https://hackr.io/blog/float-vs-double)

Arufonsu (approved):

lodicolo (approved):

PR #950, opened 2021-10-12
Client Settings Revamp with Categories.

Arufonsu:
### About this Pull Request.
* Renamed "Options" to "Settings" in order to keep consistency.
* Added Tabs in order to categorize the client settings.
* "Tabs" are technically Buttons, but we will refer to them as "Tabs" for now, considering what they actually do, visually speaking.
* Game Settings has it's own tab. 
   (it is empty **for now**) 
   -  **Another PR will cover upcoming Game Settings, but first, we need to expand 
   the settings capabilities, see** #936 
   
* Video Settings has it's own tab.
* Audio Settings has it's own tab.
* Keybinding Settings (AKA: "Controls") has it's own tab.
* Apply button behaves as it usually does, but now saves the Keybinding Settings as well.
* When accessing the Settings Window from the Main or Escape Menus, Apply will save and take you back there.
* When accessing the Settings Window from the Main or Escape Menus, Cancel will take you back there without saving.
* When accessing the Settings Window by pressing a configured key, Apply will save and take you back to game.
* When accessing the Settings Window by pressing a configured key, Cancel will take you back to game without saving.

### Preview.
imagen

imagen

imagen

### About the required resources for this PR
* "settings_window.png" replaces "options_horizontal.png".
* "SettingsWindow.json" replaces "OptionsWindow.json".
* "options_dropdown.png" renamed to "settings_dropdown.png"
* Changes to EscapeMenu.json were made due to the "options" renaming.
* Changes to MenuWindow.json were made due to the "options" renaming.

[(Updated 16-10-2021) Download 'resources.zip' with the required files for this PR](https://github.com/AscensionGameDev/Intersect-Engine/files/7358016/resources.zip)
Added the sprites for tabs just in case now, they should be present in the latest asset pack since the addition of the chatbox tabs.
imagen

NOTE: This PR originally included the 3 new settings for entities info visibility, decided to exclude those for now, because they may still require some discussion and attention in a different PR afterwards, so expect more resource updates on this one !

PR #1073, opened 2022-02-12
Fix: Hides the Main Menu Window while in the character select screen.

Arufonsu:
The issue was triggered by a mistake i previously made with the Settings Revamp commit (#950): 

### **The Issue:**

If we navigated to the Settings Window, applied changes or navigated back to the main menu and then we decided to log-in in order to see the character select screen, the Main Menu Window would show up behind the character select window.

imagen

(Test purposes) Made the character select windows transparent to see what the heck was behind:
imagen

### **The Fix:**
The previously explained behavior should be corrected with the commit from this pull request.

imagen

(Test purposes) with transparent character select window
imagen


AlexVild (approved):
Lol, easy enough to miss.

Cheshire92 (approved):

PR #1075, opened 2022-02-13
Inventory Stack limit respected when pulling out of bank

### LINKED ISSUE #1062: Pulling items out of the bank invalidates the inventory stack limit in favor of the bank stack limit
**Describe the bug**
If some item has a different bank stack limit than its inventory stack limit, a player can withdraw an amount > the inventory stack limit from the bank, and the stack in their inventory will reflect that of the bank's limit - allowing players to have more of an item than allowed in their inventory.

This does not appear to work if the item has an inventory stack limit of 1 - then it seems to function okay.

**Version Encountered**

0.7.x

**To Reproduce**
1. Create an item, with a bank stack limit of 10,000 and an inventory limit of 1,000
2. Place 10,000 of the item in your bank, so that the bank contains a stack of 10,000
3.  Withdraw 10,000 of this item.
4. You will now have a stack of 10,000 in your inventory.

**Expected behavior**
On withdraw, we should have:
1. Checked to see how many slots are required for 10,000 of the item. In this case, 10
2. Checked to see if the player had the space for such a withdraw.
3. If so, give them 10 stacks of 1,000 each.
4. Else, tell them they can not perform this action.

AlexVild:
Fixes #1062

### DIFF
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -1786,6 +1786,8 @@ namespace Intersect.Server.Entities
 
             // Get this information so we can use it later.
             var openSlots = FindOpenInventorySlots().Count;
+            var slotsRequired = (int)Math.Ceiling(item.Quantity / (double) item.Descriptor.MaxInventoryStack);
+
             int spawnAmount = 0;
 
             // How are we going to be handling this?
@@ -1806,7 +1808,7 @@ namespace Intersect.Server.Entities
                         GiveItem(item, sendUpdate);
                         return true;
                     }
-                    else if (item.Descriptor.Stackable && openSlots == 0) // Is stackable, but no inventory space.
+                    else if (item.Descriptor.Stackable && openSlots < slotsRequired) // Is stackable, but no inventory space.
                     {
                         spawnAmount = item.Quantity;
                     }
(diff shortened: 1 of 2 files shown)


Arufonsu (approved):
Reviewed and Tested, works pretty fine ;) if you can force push this PR to a [single commit](https://github.com/AscensionGameDev/Intersect-Engine/pull/1075/commits/17791824c25af6fa8fa85a36b9733d9fc9a36dd6), that would be great too 👍🏽 

[Test] 
Bank stack limit: 10k
Inventory Stack Limit: 1k
Max InvSlots: 5

https://user-images.githubusercontent.com/17498701/153765786-bb60e0a4-ca66-4376-ae94-15f507335f76.mp4

Arufonsu (approved):

Intersect.Server/Entities/Player.cs
```diff
@@ -1786,6 +1786,8 @@ public bool TryGiveItem(Item item, ItemHandling handler = ItemHandling.Normal, b
 
             // Get this information so we can use it later.
             var openSlots = FindOpenInventorySlots().Count;
```
Cheshire92, line 1788: How come this is cast to a double?

AlexVild: `Math.Ceiling` requires a double, and integer division returns an integer

Cheshire92 (approved):

PR #1076, opened 2022-02-13
Stacks with available space in them were ignored on item pickup if a full stack was in front of them, and the inventory was full

### LINKED ISSUE #1063: Picking up a stackable item with a full inventory containg a full stack of the item and a not-full stack will not put the item in the inventory
**Describe the bug**
The "Inventory is Full" message fires prematurely, at first failure to place in stack.

**Version Encountered**

0.7.x

**To Reproduce**
1. Get a full stack of some item, say 100 arrows.
2. Get another stack of the same item, but leave room for more. Say, 10 arrows.
3. Fill up your inventory
4. Find a MapItem of the same type as the stackable item, and go to loot it
5. You will get a message complaining your inventory is full
6. Swap the positions of the full stack and the non-full stack, so that the non-full stack comes before
7. You will now be able to pick up the item

**Expected behavior**
When we pick up a stackable item, we should exhaust ALL possible stacks before determining the inventory is full.

**Screenshots**
image

Note the inventory - i have a full stack (beyond full, but that's a separate issue) of arrows, and another non-full stack, but I can't pick up the stack on the map (see chat log)

AlexVild:
Fix #1063

### DIFF
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -1625,31 +1625,36 @@ namespace Intersect.Server.Entities
                 if (item.Descriptor.IsStackable)
                 {
                     // Does the user have this item already?
-                    var existingSlot = FindInventoryItemSlot(item.ItemId);
-                    if (existingSlot != null)
-                    {
-                        // Can we blindly add more to this stack?
-                        var untilFull = item.Descriptor.MaxInventoryStack - existingSlot.Quantity;
-                        if (untilFull >= item.Quantity)
-                        {
-                            return true;
-                        }
+                    var itemSlots = FindInventoryItemSlots(item.ItemId);
+                    var slotsRequired = Math.Ceiling((double)item.Quantity / item.Descriptor.MaxInventoryStack);
 
-                        // Check to see if we have the inventory spaces required to hand these items out AFTER filling up the existing slot!
-                        var toGive = item.Quantity - untilFull;
-                        if (Math.Ceiling((double) toGive / item.Descriptor.MaxInventoryStack) <= FindOpenInventorySlots().Count)
+                    // User doesn't have this item yet.
+                    if (itemSlots.Count == 0)
+                    {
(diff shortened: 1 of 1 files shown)


Arufonsu (changes requested):
Reviewed and Tested,  works 💯 this PR should resolve #1063 
Try to rebase or somehow make this PR into a [single commit](https://github.com/AscensionGameDev/Intersect-Engine/pull/1076/commits/66bdb722747dc53d1a29948697aab2242347b03f) based in the dev branch 🐰 we don't need those commits with bug fixes in this PR 👍🏽 

https://user-images.githubusercontent.com/17498701/153778427-f8426e13-4ef8-448e-a993-c14985369745.mp4

Intersect.Server/Entities/Player.cs
```diff
@@ -1625,31 +1625,36 @@ public bool CanGiveItem(Item item)
...
-                    if (existingSlot != null)
-                    {
-                        // Can we blindly add more to this stack?
-                        var untilFull = item.Descriptor.MaxInventoryStack - existingSlot.Quantity;
-                        if (untilFull >= item.Quantity)
-                        {
-                            return true;
-                        }
+                    var itemSlots = FindInventoryItemSlots(item.ItemId);
+                    var slotsRequired = Math.Ceiling((double)item.Quantity / item.Descriptor.MaxInventoryStack);
```
Cheshire92, line 1629: How come this is cast to a double?
I see it used to be before as well but I'm curious what the reason was. lol

Cheshire92: But does math.ceiling make an actual difference here? Wouldn't an integer already round by default?  Curious if that would just work as well though might need some testing.

AlexVild: Integer division _rounds_, but we need to get the ceiling, else we could lie about the amount of slots we need.

Cheshire92 (approved):

PR #1077, opened 2022-02-13
A player's unmodified stats are what is considered when assigning stat points

### LINKED ISSUE #1064: Assigning Stat Points is based on your stat points with gear equipped, not your base stats.
**Describe the bug**
When you level and have points to spend, you are assigning points based on your _calculated_ stats, with gear, and not your base stats.

This allows you to either:
- Equip gear that lowers a stat, then pump points into that stat to get PAST stat limits
- Equip gear that raises a stat, and not be allowed to put points into some stat until you unequip the gear.

**Version Encountered**

0.7x

**To Reproduce**
1. Equip some equipment that raises your stats to the limit or beyond
2. Level up
3. Try to assign points to the stat - you will not be able to
4. Unequip the gear
5. You can now assign the points

**Expected behavior**
Spending points should be based on your _base_ stats, not your calculated stats.

AlexVild:
Fix #1064

### DIFF
--- a/Intersect (Core)/Intersect (Core).csproj
+++ b/Intersect (Core)/Intersect (Core).csproj
@@ -405,6 +405,7 @@
     <Compile Include="Network\Packets\Server\AnnouncementPacket.cs" />
     <Compile Include="Network\Packets\Server\CancelCastPacket.cs" />
     <Compile Include="Network\Packets\Server\PlayAnimationPackets.cs" />
+    <Compile Include="Network\Packets\Server\PlayerStatsPacket.cs" />
     <Compile Include="Network\Packets\SlotQuantityPacket.cs" />
     <Compile Include="Network\Packets\Client\SwapBankItemsPacket.cs" />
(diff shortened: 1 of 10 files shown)


Cheshire92:
Instead of adding a true stats packet and whatnot, why don't we just send the correct stats over in the first place and have a calculated and raw stat collection on the client?

AlexVild:
> Instead of adding a true stats packet and whatnot, why don't we just send the correct stats over in the first place and have a calculated and raw stat collection on the client?

Interesting - does the client know about what a player has equipped/its stats and stuff? I guess it probably must

Arufonsu:
> guess it probably must

from what i can tell visually in the client, is that it somewhat knows that you are equipped with X equipment that modifies Y stats for sure, which is why the + button re-appears on the affected stats when you equip something, despite the fact that we already reached the base stat limit (this is not right as behavior). 

Also, like i said in the issue report, i'm still unable to reproduce the issue where pressing the wrongly shown up + button raises the stat past it's own limit.

PR #1078, opened 2022-02-13
Fix issues where DoTs would stack multiple times for the same spell

### LINKED ISSUE #1065: Damage over Times are not actually unique to the caster
**Describe the bug**
Damage over times have strange behavior where they can be applied multiple times, stacking the _effect_ rather than the duration.

**Version Encountered**

0.7.x

**To Reproduce**
1. Create a projectile with a big animation, that causes a DoT. Make it move slowly and cover a wide array of space
2. Log in as another player, and have the original player cast this DoT. As this player, move around in the spawned projectiles
3. You will have these DoT stack tremendously.

**Expected behavior**
DoTs when acquired from an attacker who has already placed a DoT should NOT cause the DoT to deal a new set of damage - rather, it should just reset the DoTs timer. Effectively, it should clear the old DoT, and create a new one, for each attacking entitythat has triggered a DoT on some affected entity.

**Screenshots**
You can see this happen here:
https://www.youtube.com/watch?v=V1SAyvcjPUM

**Additional context**
I think this is the offending code (from the end of `Entity#TryAttackSpell`):

```cs
if (spellBase.Combat.HoTDoT)
{
    var doTFound = false;
    foreach (var dot in target.CachedDots)
    {
        if (dot.SpellBase.Id == spellBase.Id && dot.Target == this)
        {
            doTFound = true;
        }
    }

    if (doTFound == false) //no duplicate DoT/HoT spells.
    {
        new DoT(this, spellBase.Id, target);
    }
}
```

Most suspiciously, note that we check that the _TARGET_ is equivalent to `this` (the attacking entity. I believe that that line should read:

```cs
if (dot.SpellBase.Id == spellBase.Id && dot.Attacker == this)
```

AlexVild:
Fix #1065


Intersect.Server/Entities/Entity.cs
```diff
@@ -1571,21 +1571,14 @@ byte projectileDir
...
             {
-                var doTFound = false;
-                foreach (var dot in target.CachedDots)
-                {
-                    if (dot.SpellBase.Id == spellBase.Id && dot.Target == this)
-                    {
-                        doTFound = true;
-                    }
-                }
+                target.CachedDots.ToList()
```
Cheshire92, line 1577: Can you provide performance metrics for the benefit of using linq here?
Generally speaking it's a much slower alternative to a regular foreach loop with an if.

Intersect.Server/Entities/Entity.cs
```diff
@@ -1571,21 +1571,14 @@ byte projectileDir
             }
 
             //Handle DoT/HoT spells]
+
```
Cheshire92, line 1574: Unneeded newline

Intersect.Server/Entities/Entity.cs
AlexVild: Nope lol.

In general, I prefer LINQ because of its readability, writeability, and the fact that by leveraging an external lib, you leave the door open for improvements to be made down the line without you doing anything.

I found a good article, however, on performance here: https://medium.com/swlh/is-using-linq-in-c-bad-for-performance-318a1e71a732

This all being said, LMAO this was a dumb implementation because I was doing two loops.

PR #1084, opened 2022-02-24
Fix: "Player is Busy" on Party Invite.

### LINKED ISSUE #1083: False "Player is Busy" messages prevent a user from interacting with another until they relog
**Describe the bug**
A player can get in a state where no other player can trade, friend, or party with them.

**Version Encountered**

0.7.x, prerelease

**To Reproduce**
1. Login with two separate players
2. Send a friend request to one player
3. Accept the friend request on the other player
4. Now, on the original client, try to send a party invite to the same player
5. You will see a chat message: "That player is currently busy"
6. This will persist until that player relogs.

**Expected behavior**
"This player is busy" is a message that should only appear when a player, truly, already has a request of some sort displayed to them.

**Additional context**
This bug is happening because we consider someone busy if they fail the following check:

```cs
if (Trading.Requester == null && PartyRequester == null && FriendRequester == null)
```

(which, additionally, should probably be a private method on the `Player` called `isBusy()` or some equivalent)

Crucially, `FriendRequester` only gets set to null:

1. On logout
2. In a _specific case_ when a player _denies_ an incoming friend request. It _never_ gets set to null if a player _accepts_ a friend request.

Arufonsu:
* Should resolve #1083

### DIFF
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -3149,10 +3149,14 @@ namespace Intersect.Server.Entities
         }
 
         //Business
-        public bool IsBusy()
-        {
-            return InShop != null || InBank || CraftingTableId != Guid.Empty || Trading.Counterparty != null;
-        }
+        private bool IsBusy() =>
+            InShop != null ||
+            InBank ||
+            CraftingTableId != Guid.Empty ||
+            Trading.Counterparty != null ||
+            Trading.Requester != null ||
+            PartyRequester != null ||
+            FriendRequester != null;
 
         //Bank
         public bool OpenBank(bool guild = false)
@@ -3536,7 +3540,7 @@ namespace Intersect.Server.Entities
 
             if (!FriendRequests.ContainsKey(fromPlayer) || !(FriendRequests[fromPlayer] > Timing.Global.Milliseconds))
             {
-                if (Trading.Requester == null && PartyRequester == null && FriendRequester == null)
+                if (!IsBusy())
                 {
                     FriendRequester = fromPlayer;
                     PacketSender.SendFriendRequest(this, fromPlayer);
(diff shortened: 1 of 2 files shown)


Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -2395,6 +2395,8 @@ public void HandlePacket(Client client, FriendRequestResponsePacket packet)
 
                     PacketSender.SendFriends(target);
                 }
+
+                player.FriendRequester = null;
```
AlexVild, line 2399: Perhaps we could just have _one_ `player.FriendRequester = null` at the end of this method if this implementation is safe.

AlexVild:
If it's not too much to ask, I wouldn't mind seeing this logic:

`if (Trading.Requester == null && PartyRequester == null && FriendRequester == null)`

extrapolated out into a method on the Player class as well. Just throwing that out there. If we ever want to change what makes someone "busy", we'd currently have to change it in three different places.

AlexVild (approved):

Intersect.Server/Networking/PacketHandler.cs
Arufonsu: you are right, maybe we don't need to call this twice, i will give it a try by placing it elsewhere.

Arufonsu: Done ! 👍🏽

AlexVild (approved):

Intersect.Server/Entities/Player.cs
```diff
@@ -3149,10 +3149,14 @@ public bool CheckCrafting(Guid id)
         }
 
         //Business
-        public bool IsBusy()
-        {
-            return InShop != null || InBank || CraftingTableId != Guid.Empty || Trading.Counterparty != null;
-        }
+        private bool IsBusy() =>
```
AlexVild, line 3152: Hmm, I wonder if adding on to this method will cause problems for the _other_ places that use it? I'm not sure. But, pre-change, it _appears_ that a player would still be able to attack and move around when they have an invite open, which might actually be desired, as otherwise someone could spam them with invites to prevent a player from attacking them, for example.

We might need like, an `CanAcceptRequest` function instead that _only_ checks for requesters :P sorry for leading you astray

AlexVild (changes requested):

AlexVild (approved):

Cheshire92 (approved):

PR #1091, opened 2022-03-01
Fixes a visual glitch when moving spells within the spells window.

### LINKED ISSUE #1090: [Visual Glitch] Moving elements within the spells window.
**Describe the bug**
Whenever we move spells within the spells window, you can clearly see that right before the spell is placed in the desired slot, the previous slot will have the spell we are moving re-rendered for a small fraction of time, which feels odd and bad looking.

**Version Encountered**
0.7.+ (Dev Branch too).

**To Reproduce**
This video is quite self explainatory, you will notice the described issue for sure:

https://user-images.githubusercontent.com/17498701/156085401-544e177b-d267-420e-81b5-3b058e278c10.mp4

**Expected behavior**
Dragging stuff around within the spells window should behave without this bad looking visual glitch.

**Operating System:**
Windows, Linux.

Arufonsu:
_Should_ fix #1090 

### Before the Fix

https://user-images.githubusercontent.com/17498701/156087884-4bab3145-d198-4a92-8e8e-23387a7dde7b.mp4

### After the Fix

https://user-images.githubusercontent.com/17498701/156087894-ab028e17-2551-45ab-b6d4-ac4ec106d952.mp4

### DIFF
--- a/Intersect.Client/Interface/Game/Spells/SpellsWindow.cs
+++ b/Intersect.Client/Interface/Game/Spells/SpellsWindow.cs
@@ -63,23 +63,7 @@ namespace Intersect.Client.Interface.Game.Spells
             for (var i = 0; i < Options.MaxPlayerSkills; i++)
             {
                 var spell = SpellBase.Get(Globals.Me.Spells[i].Id);
-
-                if (spell != null)
-                {
-                    if (Items[i].IsDragging)
-                    {
-                        Items[i].Pnl.IsHidden = true;
-                    }
-                    else
-                    {
-                        Items[i].Pnl.IsHidden = false;
-                    }
-                }
-                else
-                {
-                    Items[i].Pnl.IsHidden = true;
-                }
-
+                Items[i].Pnl.IsHidden = spell == null || Items[i].IsDragging;
                 Items[i].Update();
             }
         }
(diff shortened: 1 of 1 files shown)


AlexVild (dismissed):

Intersect.Client/Interface/Game/Spells/SpellsWindow.cs
```diff
@@ -61,19 +62,9 @@ public void Update()
...
-                        Items[i].Pnl.IsHidden = true;
-                    }
-                }
-                else
-                {
-                    Items[i].Pnl.IsHidden = true;
-                }
+                var spell = SpellBase.Get(Globals.Me.Spells[i].Id);
+                Items[i].Pnl.IsHidden = spell == null || Items[i].IsDragging;
+                Items[i].Update();
```
Cheshire92, line 67: I don't believe there's a reason to run an update when the spell is blank?

Arufonsu: oh, you are right, for a moment i though that we needed to updated on those as well in order to properly hide the null ones, just did it so it updates after checking against != null spells and everything seems fine 👍🏽

Cheshire92 (approved):

AlexVild (approved):

lodicolo (approved):

PR #1133, opened 2022-03-21
feat: add getvar/listvars, get/setvar accepts name or id

### LINKED ISSUE #1132: feature: GetVariable as a server command
It would be cool if we could also get server variable values from the server's command line.

A new command, `getvar < id >` that returns the value of some server variable with the appropriate ID.

lodicolo:
Resolves #1132 

image


AlexVild (changes requested):
One tiny change, otherwise lg2m

AlexVild (approved):

PR #1177, opened 2022-04-02
feat: ui improvements, fixes, skin changes

### LINKED ISSUE #1137: feature: gui layout hot reload
Modifying the GUI is torture, but having to restart the client every time a change is made is even more torturous.

Loading layouts should start a file watcher on the loaded layouts, and if a file change from outside of the client is detected, the layout file will be reloaded and automatically applied to the active user interface.

**Additional context**
AscensionGameDev/Intersect-Engine#1136 is an MVP (minimum viable product) feature. This is for the stable/safe implementation that has crash protection.

lodicolo:
Resolves #1137

- added checkbox, radio button to skins
  - made them composed and added extra states
- debug window persists on all screens
- improvements to settings window
- improvements to layout hot reloading (toggleable switch, but it still will break the UI in game)
  - while we continue to use GWEN the way we do, layout hot reloading simply will not be completely functional, every hot reload makes the UI disjoint from the data it's representing, and also messes up the loaded files, and additionally will cause windows that shouldn't be visible to be visible prematurely, causing crashes **use at your own risk**
- admin window can be styled

### DIFF
--- a/Intersect (Core)/Color.cs
+++ b/Intersect (Core)/Color.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 using Intersect.Localization;
@@ -6,7 +7,7 @@ using MessagePack;
 namespace Intersect
 {
     [MessagePackObject]
-    public class Color
+    public class Color : IEquatable<Color>
     {
 
         public enum ChatColor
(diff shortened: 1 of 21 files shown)


lodicolo:
Asset PR: https://github.com/AscensionGameDev/Intersect-Assets/pull/12

Arufonsu (approved):

PR #1201, opened 2022-04-07
Add a Slot Check Condition

### LINKED ISSUE #1189: feature: Add a Check Slot Condition
I'm always frustrated when I want to check if the slot is equipped with an item.

Add condition that checks if slot has item equipped, regardless of item, inside the menu check if the item is equipped

**Additional context**

WeylonSantana:
Resolves #1189

Adds a new condition that checks whether or not the slot is busy.

Testing: https://www.ascensiongamedev.com/resources/filehost/d041471d7587942aa4ee656fd4c0ba25.mp4

### DIFF
--- a/Intersect (Core)/GameObjects/Events/Condition.cs
+++ b/Intersect (Core)/GameObjects/Events/Condition.cs
@@ -44,7 +44,7 @@ namespace Intersect.GameObjects.Events
 
         MapZoneTypeIs,
 
-        CheckEquipment
+        CheckEquipment,
 
     }
 
--- a/Intersect (Core)/GameObjects/Events/Condition.cs
+++ b/Intersect (Core)/GameObjects/Events/Condition.cs
@@ -44,7 +44,7 @@ namespace Intersect.GameObjects.Events
 
         MapZoneTypeIs,
 
-        CheckEquippedSlot
+        CheckEquipment
 
     }
 
@@ -339,11 +339,9 @@ namespace Intersect.GameObjects.Events
     public class CheckEquippedSlot : Condition
     {
 
-        public override ConditionTypes Type { get; } = ConditionTypes.CheckEquippedSlot;
+        public override ConditionTypes Type { get; } = ConditionTypes.CheckEquipment;
 
-        public int SlotIndex { get; set; }
-
-        public string SlotName { get; set; }
+        public string Name { get; set; }
 
     }
 
--- a/Intersect (Core)/GameObjects/Events/Condition.cs
+++ b/Intersect (Core)/GameObjects/Events/Condition.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 using Intersect.Enums;
 
@@ -43,6 +43,8 @@ namespace Intersect.GameObjects.Events
         InGuildWithRank,
 
         MapZoneTypeIs,
+
+        CheckEquippedSlot
 
     }
 
@@ -331,6 +333,17 @@ namespace Intersect.GameObjects.Events
         public StringVariableComparators Comparator { get; set; } = StringVariableComparators.Equal;
 
         public string Value { get; set; }
+
+    }
+
+    public class CheckEquippedSlot : Condition
+    {
+
+        public override ConditionTypes Type { get; } = ConditionTypes.CheckEquippedSlot;
+
+        public int SlotIndex { get; set; }
+
+        public string SlotName { get; set; }
 
     }
 
--- a/Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
+++ b/Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
@@ -253,8 +253,8 @@ namespace Intersect.Editor.Forms.Editors.Events.Event_Commands
             chkBank.Text = Strings.EventConditional.CheckBank;
 
             //Check Equipped Slot
-            grpCheckEquippedSlot.Text = Strings.EventConditional.checkequippedslotgrp;
-            lblCheckEquippedSlot.Text = Strings.EventConditional.checkequippedslotlbl;
+            grpCheckEquippedSlot.Text = Strings.EventConditional.checkequipment;
+            lblCheckEquippedSlot.Text = Strings.EventConditional.equipmentslot;
 
             btnSave.Text = Strings.EventConditional.okay;
             btnCancel.Text = Strings.EventConditional.cancel;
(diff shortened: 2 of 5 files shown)


Intersect.Server/Entities/Events/Conditions.cs
```diff
@@ -443,6 +443,24 @@ QuestBase questBase
...
+        {
+            for (var i = 0; i < Options.EquipmentSlots.Count; i++)
+            {
+                if (player.Equipment[condition.SlotIndex] >= 0)
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
lodicolo, line 461: This should be just

```csharp
var equipmentIndex = Options.EquipmentSlots.IndexOf(condition.Name);
return player.Equipment[equipmentIndex] >= 0;
```

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2085,7 +2094,8 @@ public struct EventConditional
                 {17, @"Item Equipped is..."},
                 {18, @"Has X free Inventory slots..." },
                 {19, @"In Guild With At Least Rank..." },
-                {20, @"Map Zone Type is..." }
+                {20, @"Map Zone Type is..." },
+                {21, @"Check Equipped Slot... " }
```
lodicolo, line 2098: Delete the extra space ` ` in the string, and end the line with `,` so that future additions don't need to modify two lines.

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2042,6 +2047,10 @@ public struct EventConditional
 
             public static LocalizedString canstartquest = @"Can Start Quest";
 
+            public static LocalizedString checkequippedslotlbl = @"Slot";
```
lodicolo, line 2050: Name the string `EquipmentSlot` and change the default text to "Equipment Slot" please.

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2042,6 +2047,10 @@ public struct EventConditional
...
+
+            public static LocalizedString checkequippedslotgrp = @"Check Item Slot:";
```
lodicolo, line 2052: Name the string `Check Equipment` and change the text to `Check Equipment`, do not include `:` in the string.

Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
```diff
@@ -1271,6 +1293,11 @@ private void SetupFormValues(MapZoneTypeIs condition)
             }
         }
 
+        private void SetupFormValues(CheckEquippedSlot condition)
+        {
+            cmbCheckEquippedSlot.SelectedIndex = condition.SlotIndex;
```
lodicolo, line 1298: Instead of `condition.SlotIndex` it would be set to `Options.EquipmentSlots.IndexOf(condition.Name)`

Intersect (Core)/GameObjects/Events/Condition.cs
```diff
@@ -334,4 +336,15 @@ public class StringVariableComparison : VariableCompaison
 
     }
 
+    public class CheckEquippedSlot : Condition
+    {
+
+        public override ConditionTypes Type { get; } = ConditionTypes.CheckEquippedSlot;
+
+        public int SlotIndex { get; set; }
```
lodicolo, line 344: This isn't necessary to include, it's basically a duplicate index.

Intersect (Core)/GameObjects/Events/Condition.cs
```diff
@@ -334,4 +336,15 @@ public class StringVariableComparison : VariableCompaison
...
+
+        public string SlotName { get; set; }
```
lodicolo, line 346: Rename to just `Name`.

Intersect (Core)/GameObjects/Events/Condition.cs
```diff
@@ -44,6 +44,8 @@ public enum ConditionTypes
 
         MapZoneTypeIs,
 
+        CheckEquippedSlot
```
lodicolo, line 47: Rename this to `CheckEquipment`

lodicolo (changes requested):

Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
```diff
@@ -383,6 +387,14 @@ private void ConditionTypeChanged(ConditionTypes type)
                         cmbMapZoneType.SelectedIndex = 0;
                     }
 
+                    break;
+                case ConditionTypes.CheckEquipment:
+                    Condition = new CheckEquippedSlot();
+                    if (cmbCheckEquippedSlot.Items.Count > 0)
+                    {
+                        cmbCheckEquippedSlot.SelectedIndex = 0;
+                    }
```
lodicolo, line 396: Look at this again, the `if` is not needed, just set the SelectedIndex to 0, just like on L387 (for MapZoneTypeIs)

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2042,6 +2047,10 @@ public struct EventConditional
 
             public static LocalizedString canstartquest = @"Can Start Quest";
 
+            public static LocalizedString equipmentslot = @"Slot";
```
lodicolo, line 2050: Change the name to `EquipmentSlot`, with that exact capitalization.

Doing all lowercase was a bad design choice by JC, when the strings can be referenced case-insensitively during loading/saving. We are trying to fix this with each new PR rather than repeat the mistake since it requires humans to fix the casing.

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2042,6 +2047,10 @@ public struct EventConditional
...
+
+            public static LocalizedString checkequipment = @"Check Equipment";
```
lodicolo, line 2052: The name should be `CheckEquipment` with that capitalization

Intersect (Core)/GameObjects/Events/Condition.cs
```diff
@@ -44,6 +44,8 @@ public enum ConditionTypes
 
         MapZoneTypeIs,
 
+        CheckEquipment
```
lodicolo, line 47: Since you are need to make one more revision anyway and I forgot last time, please add a trailing comma here so new additions only modify one line just like in the other review I did in this PR.

lodicolo (changes requested):

Intersect.Editor/Localization/Strings.cs
WeylonSantana: I understand, I didn't know that, when you asked to rename it I thought it was just a writing habit, and I continued following the pattern I was, I will change

lodicolo (approved):

AlexVild:
It _looks_ like it should, but if someone could verify:

1) The editor populates the fields correctly when editing an existing conditional that's using this conditition
2) The command can be copied/pasted correctly

then this looks fine

AlexVild (approved):

PR #1218, opened 2022-04-18
Random Crafting Options

### LINKED ISSUE #1217: feature: Random Crafting Options
It would be nice if crafting had options for randomness in crafting success. This way we would cover a larger audience of creators who want the feature, if they want the standard intersect, just leave the option 0%. 

Chance of crafting to fail. And bonus if you have a chance to NOT lose items on failure.

**Additional context**

WeylonSantana:
resolves #1217

Two new options for the crafting editor.

Failure (%) - What is the chance of failure?
Item Loss (%) - What is the chance of losing the item?

How it should work: 
https://www.ascensiongamedev.com/resources/filehost/b5fc6817ff28f1e478ce33f1979df16c.mp4


Intersect.Server/Entities/Player.cs
```diff
@@ -3605,6 +3605,8 @@ public void CraftItem(Guid id)
                     invbackup.Add(item.Clone());
                 }
 
+                var mCraftItem = CraftBase.Get(id);
```
lodicolo, line 3608: Just `craftItem` -- `m` is used for member variables on a class, this is only local.

Intersect.Server/Entities/Player.cs
```diff
@@ -3646,52 +3648,90 @@ public void CraftItem(Guid id)
                     }
                 }
 
-                //Take the items
-                foreach (var c in CraftBase.Get(id).Ingredients)
+                //Return true if the craft was a success
+                if (Randomization.Next(1, 101) > mCraftItem.FailureChance)
```
lodicolo, line 3652: This should be `Next(0, 101)`.

Intersect.Server/Entities/Player.cs
```diff
@@ -3646,52 +3648,90 @@ public void CraftItem(Guid id)
...
+                            return;
+                        }
+                    }
+
+                    //Give them the craft
+                    var quantity = Math.Max(mCraftItem.Quantity, 1);
+                    var itm = ItemBase.Get(mCraftItem.ItemId);
+                    if (itm == null || !itm.IsStackable)
+                    {
+                        quantity = 1;
```
lodicolo, line 3676: Separate this.

- if not stackable, quantity should be 1 like you have here
- if null, log an error to the console with information about the item id, craft id, and quantity (Log.Error), send some sort of generic error message to the client, and then just `return`

lodicolo (changes requested):

Intersect.Server/Entities/Player.cs
WeylonSantana: I won't know how to do this item, I don't know how to send a log to the server, nor to the client.

Intersect.Server/Localization/Strings.cs
```diff
@@ -497,6 +497,12 @@ public sealed class CraftingNamespace : LocaleNamespace
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public readonly LocalizedString nospace = @"You do not have enough inventory space to craft {00}!";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString CraftFailure = @"The attempt to create the item {00} failed!";
+
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString LostItems = @"You lost the materials!";
```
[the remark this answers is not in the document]
WeylonSantana: Could you give me some idea? as to why it won't always be that the player will lose the items, so I thought it might be necessary to let you know when you lose them. I don't like the idea of two lines either, but I couldn't think of anything else at that time.

Now I could only think about this:
image
image

Intersect.Server/Entities/Player.cs
```diff
@@ -3646,52 +3648,90 @@ public void CraftItem(Guid id)
...
+                            PacketSender.SendInventory(this);
+                            CraftId = Guid.Empty;
+
+                            return;
+                        }
+                    }
+
+                    //Give them the craft
+                    var quantity = Math.Max(mCraftItem.Quantity, 1);
+                    var itm = ItemBase.Get(mCraftItem.ItemId);
```
Arufonsu, line 3673: instead of "itm", try with "craftItemId" for this local variable's name.

Intersect.Server/Entities/Player.cs
Arufonsu: > I won't know how to do this item, I don't know how to send a log to the server, nor to the client.

here is a little approach/example on how you could handle panda's suggestion:

```
var craftItemId = ItemBase.Get(mCraftItem.ItemId);

if (craftItemId == null)
{
    PacketSender.SendChatMsg(this, Strings.Crafting.craftError, ChatMessageType.Crafting, CustomColors.Alerts.Error);
    Log.Error("[Error]:" + this.Name + " attempted to craft a null item!");
    return;
}
if (!craftItemId.IsStackable)
{
    quantity = 1;
}
```

EDIT: you don't want to hardcode the Log.Error message, this is just an example !

Arufonsu (changes requested):

Intersect.Server/Entities/Player.cs
WeylonSantana: @Arufonsu Thanks for your time on this question, but I'm not really going to know how to do it haha, as I said to panda, I didn't do that part of the code, I just replace all craftbase.get() in the method with the variable. If you can send me the code to replace it would help me, because I really don't understand about it. (for example, I saw the code above, I understand, but you said that hardcoded is not ideal, so I have no idea what it would be like. Strings on the server maybe?)

Intersect.Server/Entities/Player.cs
lodicolo: Uh, it shouldn't be craftItemId, it's not an id, it's the descriptor. `item` is better than `itm` though.

Intersect.Server/Entities/Player.cs
lodicolo:
```csharp
if (item == null)
{
    PacketSender.SendChatMsg(this, Strings.Errors.UnknownErrorTryAgain, ChatMessageType.Error, CustomColors.Alerts.Errors);
    Log.Error($"Unable to find item descriptor {craftItem.ItemId} for craft {craftItem.Id}.");
    return;
}

if (!craftItemId.IsStackable) {
    quantity = 1;
}
```

Where UnknownErrorTryAgain is `An unknown error occurred, please try again.`

lodicolo (changes requested):

WeylonSantana:
Finished the changes. Everything still working.
image

Intersect.Server/Entities/Player.cs
```diff
@@ -3605,7 +3605,8 @@ public void CraftItem(Guid id)
                     invbackup.Add(item.Clone());
                 }
 
-                var mCraftItem = CraftBase.Get(id);
+                var craftItem = CraftBase.Get(id);
+                var finalItem = ItemBase.Get(craftItem.ItemId);
```
lodicolo, line 3609: This is on me for having a brain fart earlier but `craftItem` should be `craft`, `finalItem` should be `item`.

Intersect.Server/Entities/Player.cs
```diff
@@ -3669,21 +3670,30 @@ public void CraftItem(Guid id)
...
-                    var quantity = Math.Max(mCraftItem.Quantity, 1);
-                    var itm = ItemBase.Get(mCraftItem.ItemId);
-                    if (itm == null || !itm.IsStackable)
+                    var quantity = Math.Max(craftItem.Quantity, 1);
+                    if (finalItem == null)
+                    {
+                        PacketSender.SendChatMsg(this, Strings.Errors.UnknownErrorTryAgain, ChatMessageType.Error, CustomColors.Alerts.Error);
+                        Log.Error($"Unable to find item descriptor {finalItem.Id} for craft {craftItem.Id}.");
+                        return;
+                    }
```
lodicolo, line 3679: You can move this to the top of the method directly beneath `item = ItemBase...`

lodicolo (changes requested):

Intersect.Server/Entities/Player.cs
WeylonSantana: haha lol, okay

lodicolo (approved):

Arufonsu (approved):

PR #1229, opened 2022-04-27
feat: ✨ add diagonal movement

blinkuz:
This is a proof of concept with the base code for what the diagonal movement would be.

All the base code used in this PR is taken from the patches shared by the user Shenmue in the forum:
https://www.ascensiongamedev.com/topic/4908-diagonal-movement/

Notes:  
- This PR contains the code for the optional diagonal projectile logic mentioned in the forum post.
- This PR contains the code for the optional function of attacking enemies if they are in range of 5 tiles in front of him.

image

### DIFF
--- a/Intersect (Core)/Enums/Directions.cs
+++ b/Intersect (Core)/Enums/Directions.cs
@@ -10,8 +10,15 @@ namespace Intersect.Enums
 
         Left,
 
-        Right
+        Right,
+        
+        UpLeft,
 
+        UpRight,
+
+        DownLeft,
+
+        DownRight
     }
 
 }
--- a/Intersect (Core)/GameObjects/Events/Enums.cs
+++ b/Intersect (Core)/GameObjects/Events/Enums.cs
@@ -93,6 +93,14 @@
         SetGraphic,
 
         SetAnimation,
+        
+        MoveUpLeft,
+
+        MoveUpRight,
+
+        MoveDownLeft,
+
+        MoveDownRight
 
     }
 
--- a/Intersect (Core)/Network/Packets/Client/AttackPacket.cs
+++ b/Intersect (Core)/Network/Packets/Client/AttackPacket.cs
@@ -12,13 +12,17 @@ namespace Intersect.Network.Packets.Client
 
         }
         
-        public AttackPacket(Guid target)
+        public AttackPacket(Guid target, bool targetOnFocus)
         {
             Target = target;
+            TargetOnFocus = targetOnFocus;
         }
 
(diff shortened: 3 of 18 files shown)


jcsnider:
Few things:

Please make sure to adjust movement speed/time based on diagonal or not. Diagonal movement is more distance traveled than just moving left/right or up/down (think of it like the long side of a triangle). The distance traversed when moving diagonally would be sqrt(2) or 1.41421356237 tiles. Knowing that, the movement time should be longer when players/npcs/events are moving diagonally otherwise they will appear to get speed boosts when moving diagonally in game.

Secondly, diagonal movement, and the options within for projectiles/attack range should be options in the servers config so games which do not want this functionality can turn it off.

Thirdly, please make sure that diagonal movement applies and works for map critters. I think they came out long after the original Shenmue patches.

jcsnider:
There will be functions for GetMovementTime() in the client and server:
https://github.com/AscensionGameDev/Intersect-Engine/blob/d0f74d5569cdae63a0dc5928123650748ff380d4/Intersect.Server/Entities/Entity.cs#L818

I think all you will really need to do is take the result and multiply it by 1.41421356237 IF the entity is moving diagonally. 

I added this before the Math.Min() line in Ambardia in order to accomplish it:
```
             if (diag)
            {
                time = (float)Math.Sqrt((time * time) * 2);
            }
```

Cheshire92:
Something similar (though more lenient) on timing is added to Leafling's diagonal movement as well.
The proposed delay felt very slow for players and it was changed based on feedback.

jcsnider:
> Something similar (though more lenient) on timing is added to Leafling's diagonal movement as well. The proposed delay felt very slow for players and it was changed based on feedback.

I'm sure it would feel insanely slow after experiencing the original logic where you were literally traversing distances 40% faster XD

Cheshire92:
Yeah the logic used is as follows:

`time = (float)Math.Sqrt((time * time) * 1.5);`

Not a huge difference, but enough to make it flow just a little faster.

Cheshire92:
The proper fix is as JC provided it.
I was just pitching in that some people may find it too slow for their liking.

dudevelopement:
Basta adicionar

```

            if (Dir > 3)
            {
                time = (float)Math.Sqrt((time * time) * 2);
            }
```
e está funcionando para mim.

blinkuz:
I proceed to close this PR since the feature was developed in this https://github.com/AscensionGameDev/Intersect-Engine/pull/1708 solving several points that the current one lacks.

PR #1264, opened 2022-05-23
resolve: (Day) Add sound effect selection to warp attribute

AlexVild:
Adds the ability to play a sound effect when warping a player via map attribute.

The attribute editor will look like this when "Warp" is selected:
image


lodicolo (approved):

lodicolo (approved):

Arufonsu (approved):
Interesting, I just tried this and it is very **useful**, the usual way would be to trigger the sounds to play through the event itself during the warp, but this is quite useful to do the same thing in one go without requiring additional event actions. 👍🏽

PR #1302, opened 2022-06-22
Blocking Refactor I

### LINKED ISSUE #1300: bug: attack animations runs after blocking
**Describe the bug**
When the sprite has custom attack sprites (attack, weapon or shoot), this animation is played after blocking.

**Version Encountered**
0.7.1.22

**To Reproduce**
1. Make extra sprite animation
2. Equip a shield and release
3. After release de block button see error

**Expected behavior**
Do not run attack animation, if not attacking

**Screenshots**
https://www.ascensiongamedev.com/resources/filehost/04fa0e69f0facca18722f5342a14343b.mp4

**Logs**
If there are error logs please attach them below.

**Operating System:**
Windows

**Additional context**

WeylonSantana:
Resolves #1300

I believe that the block was created to act as a "status", the player while holding the button is in a "defense state" until released, I changed that.

I'm using the attack logic as a base to modify the blocking, now, it is working similar to the attack, as an action and not as a "state", so while holding the block button, it keeps making several attempts to block, as it works in attack.
Basically, the player either attacks or defends, never both, and [this fixes the attack animation playing after blocking.](https://github.com/AscensionGameDev/Intersect-Engine/issues/1300)

I just changed the logic a little bit of how it works to make it look like the attack with "speed modifier".

### DIFF
--- a/Intersect.Client/Core/Input.cs
+++ b/Intersect.Client/Core/Input.cs
@@ -185,7 +185,10 @@ namespace Intersect.Client.Core
                                         break;
 
                                     case Control.Block:
-                                        Globals.Me?.TryBlock();
+                                        if(Globals.Me.IsBlocking == false)
+                                        {
+                                            Globals.Me?.TryBlock();
+                                        }
 
                                         break;
 
@@ -300,11 +303,6 @@ namespace Intersect.Client.Core
             {
                 return;
             }
-
-            if (Controls.Controls.ControlHasKey(Control.Block, modifier, key))
-            {
-                Globals.Me.StopBlocking();
-            }
         }
 
         public static void OnMouseDown(Keys modifier, MouseButtons btn)
@@ -406,11 +404,6 @@ namespace Intersect.Client.Core
             if (Globals.Me == null)
             {
                 return;
-            }
-
-            if (Controls.Controls.ControlHasKey(Control.Block, modifier, key))
-            {
-                Globals.Me.StopBlocking();
             }
 
             if (btn != MouseButtons.Right)
--- a/Intersect.Client/Core/Input.cs
+++ b/Intersect.Client/Core/Input.cs
@@ -185,10 +185,7 @@ namespace Intersect.Client.Core
                                         break;
 
                                     case Control.Block:
-                                        if(Globals.Me.IsBlocking == false)
-                                        {
-                                            Globals.Me?.TryBlock();
-                                        }
+                                        Globals.Me?.TryBlock();
 
                                         break;
 
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -242,6 +242,15 @@ namespace Intersect.Client.Entities
                         }
                     }
                 }
+
+                //Holding block button for "auto blocking"
+                if(Controls.KeyDown(Control.Block))
+                {
+                    if(!IsBlocking)
+                    {
+                        TryBlock();
+                    }
+                }
             }
 
             if (TargetBox == default && this == Globals.Me && Interface.Interface.GameUi != default)
@@ -1381,23 +1390,11 @@ namespace Intersect.Client.Entities
                 if (item != null)
                 {
                     PacketSender.SendBlock(true);
-                    IsBlocking = true;
-
                     return true;
                 }
             }
 
             return false;
-        }
-
-        public void StopBlocking()
-        {
-            if (IsBlocking)
-            {
-                IsBlocking = false;
-                PacketSender.SendBlock(false);
-                AttackTimer = Timing.Global.Ticks / TimeSpan.TicksPerMillisecond + CalculateAttackTime();
-            }
         }
 
         public bool TryAttack()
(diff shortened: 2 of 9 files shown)


Intersect.Client/Core/Input.cs
```diff
@@ -185,7 +185,10 @@ public static void OnKeyPressed(Keys modifier, Keys key)
                                         break;
 
                                     case Control.Block:
-                                        Globals.Me?.TryBlock();
+                                        if(Globals.Me.IsBlocking == false)
```
lodicolo, line 188: Instead of checking outside of TryBlock, this check belongs inside of TryBlock:

```
if (IsBlocking)
{
    return true;
}
```

Intersect.Client/Entities/Player.cs
```diff
@@ -242,6 +242,15 @@ public override bool Update()
...
                     }
                 }
+
+                //Holding block button for "auto blocking"
+                if(Controls.KeyDown(Control.Block))
+                {
+                    if(!IsBlocking)
+                    {
+                        TryBlock();
+                    }
```
lodicolo, line 252: See above `IsBlocking` check comment

Intersect.Client/Networking/PacketHandler.cs
```diff
@@ -1014,10 +1014,30 @@ public void HandlePacket(IPacketSender packetSender, EntityAttackPacket packet)
...
+                    if (attackTimer > -1)
+                    {
+                        en.AttackTimer = Timing.Global.Ticks / TimeSpan.TicksPerMillisecond + attackTimer;
+                        en.IsBlocking = true;
+                    }
+                    else
+                    {
+                        en.IsBlocking = false;
+                    }
+                }
```
lodicolo, line 1040: I think this can be greatly simplified to something like:

```csharp
var isSelf = en == Globals.Me;
if (isSelf)
{
    en.IsBlocking = packet.IsBlocking;
}

if (attackTimer > -1)
{
    en.AttackTimer = Timing.Global.Ticks / TimeSpan.TicksPerMillisecond + attackTimer;
    if (!isSelf)
    {
        en.AttackTime = attackTimer;
    }
}
```

Intersect.Server/Entities/Entity.cs
```diff
@@ -346,6 +346,13 @@ public virtual void Update(long timeMs)
                     {
                         status.TryRemoveStatus();
                     }
+
+                    //Blocking timers
+                    if(AttackTimer < Timing.Global.Milliseconds && Blocking)
```
lodicolo, line 351: Make this be `Blocking && ...`.

Intersect.Server/Entities/Entity.cs
```diff
@@ -812,7 +819,7 @@ public virtual float GetMovementTime()
             var time = 1000f / (float) (1 + Math.Log(Stat[(int) Stats.Speed].Value()));
             if (Blocking)
             {
-                time += time * Options.BlockingSlow;
+                time += time * (float)Options.BlockingSlow;
```
lodicolo, line 822: What effect does this change have?

Intersect.Server/Networking/PacketSender.cs
```diff
@@ -955,9 +955,9 @@ public static void SendEntityDir(Entity en)
         }
 
         //EntityAttackPacket
-        public static void SendEntityAttack(Entity en, int attackTime)
+        public static void SendEntityAttack(Entity en, int attackTime, bool isBlocking = false)
```
lodicolo, line 958: I think this parameter should always be explicit, no default `= false` here.

lodicolo (changes requested):

Intersect.Server/Entities/Entity.cs
WeylonSantana: I think I saw that on the client side it was like this, so I left both the same, client and server, nothing really changes

Intersect.Server/Networking/PacketSender.cs
WeylonSantana: The intention is to make blocking separate from the attack, later on (maybe after this merge), so it's implicit only for the blocking use case, to make it easier to remove when separating from the attack

Intersect.Client/Networking/PacketHandler.cs
```diff
@@ -1014,29 +1014,18 @@ public void HandlePacket(IPacketSender packetSender, EntityAttackPacket packet)
...
+            var isSelf = en == Globals.Me;
+            if (isSelf)
             {
-                //attack handler
-                if (attackTimer > -1 && en != Globals.Me)
-                {
-                    en.AttackTimer = Timing.Global.Ticks / TimeSpan.TicksPerMillisecond + attackTimer;
-                    en.AttackTime = attackTimer;
-                }
+                en.IsBlocking = packet.IsBlocking;
```
lodicolo, line 1020: I just realized that my suggestion is not actually compatible with your logic, this is my bad.

I think this actually needs to be (in order to match the functionality you were adding):

```
if (isSelf && packet.IsBlocking)
{
    en.IsBlocking = attackTimer > -1;
}
```

lodicolo (changes requested):

lodicolo (approved):

Arufonsu (approved):

PR #1304, opened 2022-06-22
fix: (Day) Prevent NPCs from despawning when their reset positions are blocked.

### LINKED ISSUE #1303: bug: NPCs will die if their reset location is blocked
**Describe the bug**
If you have NPC reset radius turned on, and block an NPC's reset position, either with a player, global event, or NPC, they will eventually just die. This is problematic, especially for things such as bosses.

**Version Encountered**

Latest prerelease

**To Reproduce**
1. Turn on NPC reset behavior in the server config
2. Make an NPC, not aggressive by default, with some reset radius
3. Log in with two clients
4. On one client, draw the NPC with an attack
5. On the other client, stand in the location the NPC was located before being aggro'd
6. Move the NPC away until it begins to reset
7. When the NPC arrives at its reset location, it will freak out for a bit, and then despawn, awarding no one the kill.

**Expected behavior**
The NPC should instead just consider their new location their reset.

AlexVild:
Resolves #1303

### DIFF
--- a/Intersect (Core)/Config/NpcOptions.cs
+++ b/Intersect (Core)/Config/NpcOptions.cs
@@ -44,6 +44,15 @@
         /// </summary>
         public bool ShowLevelByName = false;
 
+        /// <summary>
+        /// If true, NPCs that are resetting will walkthrough players
+        /// </summary>
+        public bool IntangibleDuringReset = true;
+
+        /// <summary>
+        /// If true, NPCs will go to reset state if their combat timer is exceeded
+        /// </summary>
+        public bool ResetIfCombatTimerExceeded = true;
     }
 
 }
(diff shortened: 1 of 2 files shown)


Cheshire92:
How does this fix the issue exactly?
Like I said on Discord, part of the reset system is meant to PREVENT NPCs from being dragged all over the world. This will allow players to do exactly that again.

You've effectively replaced one bug with another intentionally this way.
I think a cleaner solution would be to allow NPCs to always bypass players and other entities while they are resetting.

AlexVild:
@Cheshire92 Changed to reflect our discussion - intangibility on reset can now be configured (default true), as can resetting due to an NPC combat timer exception

WeylonSantana (approved):

Arufonsu (approved):
well, making them able to pass thru players uppon reset seems to work so far 👍🏽

lodicolo (approved):

PR #1336, opened 2022-06-24
fix bags with deleted items have slots taken by ghost items

### LINKED ISSUE #1335: bug: deleted items inside bags take up space and cannot be withdrawn
### Description

When items which are stored in bags get deleted from the editor they become 'ghosts' which continue to take up the bag slots despite nothing being there. The slot is taken and cannot be reused whatsoever.

img

### Steps to Reproduce

1. Add items to bag.
2. Delete item from editor which is inside bag.
3. Try to store items in that slot.

### Version with bug

v7.1.22

Unknown

### Affected platforms

Windows, macOS, Linux

N/A

### Relevant log output

```shell
N/A
```

### Duplicate Bug Check

jcsnider:
Resolves #1335

### DIFF
--- a/Intersect.Server/Database/Item.cs
+++ b/Intersect.Server/Database/Item.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel.DataAnnotations.Schema;
 
 using Intersect.Enums;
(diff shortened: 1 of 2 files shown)


Intersect.Server/Database/Item.cs
```diff
@@ -122,6 +122,18 @@ public bool TryGetBag(out Bag bag)
...
             bag = Bag;
 
+            if (bag != null)
+            {
+                //Remove any items from this bag that have been removed from the game
+                foreach (var itm in bag.Slots)
+                {
+                    if (itm.ItemId != Guid.Empty && ItemBase.Get(itm.ItemId) == null)
+                    {
+                        itm.Set(new Item());
```
AlexVild, line 132: Just wondering - in other places I've seen `Item.Set(Item.Empty)` instead - which do we prefer?

AlexVild (approved):

Arufonsu (dismissed):

WeylonSantana (dismissed):

Intersect.Server/Database/Item.cs
```diff
@@ -122,6 +122,18 @@ public bool TryGetBag(out Bag bag)
...
+                //Remove any items from this bag that have been removed from the game
+                foreach (var itm in bag.Slots)
+                {
+                    if (itm.ItemId != Guid.Empty && ItemBase.Get(itm.ItemId) == null)
+                    {
+                        itm.Set(Item.None);
+                    }
+                }
+            }
+
```
lodicolo, line 136: Can you add this as an else condition to the if statement immediately below this?

lodicolo (changes requested):

PR #1341, opened 2022-06-24
Added hide equipment map property

### LINKED ISSUE #1332: enhancement: hide paperdoll on maps via map property
### Description

A map property to hide player paperdolls on individual maps would be nice.

### Intended Use-Case

This would allow players to visually 'take a load off' in safe areas like homes/town centers/etc. And while not the intent it would also reduce rendering loads on the game clients if used in populated areas.

### Duplicate Check

jcsnider:
Resolves #1332

Property

[Preview of feature in action.](https://www.ascensiongamedev.com/resources/filehost/7e74420c754e186d25005728565c4bf3.mp4)

Other notes: when doing the game database migration it added a missing Npc.Tenacity record into the GameGontextModelSnapshot which looks like was omitted [here](https://github.com/AscensionGameDev/Intersect-Engine/commit/c962ccde730f4b497d22618a28a1695df852a916#diff-4aceccd5f41d657cbea16ca79737915d97970762b0d20012fb55c9e04c409089). As it should be in that file I am including it in this PR.


Intersect.Server/Migrations/Game/GameContextModelSnapshot.cs
```diff
@@ -468,6 +468,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Property<bool>("Swarm");
 
+                    b.Property<double>("Tenacity");
```
WeylonSantana, line 471: Lately when I've been doing migrations I delete this line

Intersect.Server/Migrations/Game/GameContextModelSnapshot.cs
```diff
@@ -786,7 +788,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.ToTable("Time");
                 });
 
-            modelBuilder.Entity("Intersect.Server.Maps.MapInstance", b =>
+            modelBuilder.Entity("Intersect.Server.Maps.MapController", b =>
```
WeylonSantana, line 791: And also I return this line to MapInstance

WeylonSantana:
That way your changes are only related to the subject of the pull request,As well, they continue to work

lodicolo (approved):

PR #1343, opened 2022-06-24
enhance: custom item attribute respawn timers and custom item despawn timers

### LINKED ISSUE #1331: enhancement: granular item spawn/respawn timers
### Description

It would be nice if we could set specific item respawn delays when placing map items via the map item attribute (which can default to the server config if set to 0).

Furthermore it would be nice if per-item (in the item editor) we could set granular despawn times so we can configure items to vanish or stick around longer.

### Intended Use-Case

This would grant us overall a lot more control over when items respawn and or vanish.

### Duplicate Check

jcsnider:
Resolves #1331

### DIFF
--- a/Intersect.Editor/Localization/Strings.cs
+++ b/Intersect.Editor/Localization/Strings.cs
@@ -963,6 +963,8 @@ namespace Intersect.Editor.Localization
 
             public static LocalizedString ResourceSpawn = @"Resource";
 
+            public static LocalizedString RespawnTime = @"Respawn Time (ms):      [0 for server default]";
+
             public static LocalizedString Slide = @"Slide";
 
             public static LocalizedString Sound = @"Sound";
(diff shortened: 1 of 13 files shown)


WeylonSantana (approved):

Intersect.Editor/Localization/Strings.cs
```diff
@@ -748,6 +748,8 @@ public partial struct Attributes
 
             public static LocalizedString quantity = @"Quantity:";
 
+            public static LocalizedString respawntime = @"Respawn Time (ms):      [0 for server default]";
```
lodicolo, line 751: should the text in `[]` not be a tooltip instead of making the label huge?

Intersect.Editor/Localization/Strings.cs
```diff
@@ -3222,6 +3224,9 @@ public partial struct ItemEditor
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public static LocalizedString DeathDropChance = @"Drop chance on Death (%):";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public static LocalizedString DespawnTime = @"Item Despawn Time (ms):     [0 for server default]";
```
lodicolo, line 3228: should the text in `[]` not be a tooltip instead of making the label huge?

lodicolo (changes requested):

lodicolo (approved):

PR #1345, opened 2022-06-24
fix #1344: force show the ui when user input is required

### LINKED ISSUE #1344: bug: hidden ui does not reappear when user input is required
### Description

If the ui is toggled to be hidden then it doesn't start rendering when chat, event dialogs, bags, banks, trading, crafting, or shopping is open and needs the users attention.

### Steps to Reproduce

1. Toggle UI Off
2. Open Bank In Game

### Version with bug

v7.1.22

Unknown

### Affected platforms

Windows, macOS, Linux

N/A

### Relevant log output

```shell
N/A
```

### Duplicate Bug Check

jcsnider:
This resolves #1344 

I found that the UI elements although hidden could still be closed, so the extra code changes in this pr fixes that as well.

### DIFF
--- a/Intersect.Client/Core/Main.cs
+++ b/Intersect.Client/Core/Main.cs
@@ -365,6 +365,14 @@ namespace Intersect.Client.Core
             Globals.JoiningGame = false;
             Globals.NeedsMaps = true;
             Globals.Picture = null;
+
+            Globals.InBag = false;
+            Globals.InBank = false;
+            Globals.GameShop = null;
+            Globals.InTrade = false;
+            Globals.EventDialogs?.Clear();
+            Globals.InCraft = false;
+
             Interface.Interface.HideUi = false;
 
             //Dump Game Objects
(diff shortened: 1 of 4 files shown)


Arufonsu (approved):
tested and confirmed to be working as intended 👍🏽 

https://user-images.githubusercontent.com/17498701/175697831-fb1003e5-d67d-419f-b604-89d0241c0306.mp4

WeylonSantana (approved):

Arufonsu:
Just a doubt: would this disallow us from hiding specific UI elements from view on specific situations?
ie. i was currently hiding every UI element except for the dialogue window during event dialogue for the sake of minimalism and ease of usage and reading.

Arufonsu:
Just a wondering: would this disallow us from hiding specific UI elements from view on specific situations?
ie. i was currently hiding every UI element except for the dialogue window during event dialogue for the sake of minimalism and ease of usage and reading.

jcsnider:
This doesn't force specific ui elements to show. It only forces the UI to render when an element is open that requires interaction.

lodicolo (approved):

PR #1346, opened 2022-06-24
Added crafting requirements to individual crafts

### LINKED ISSUE #852: Crafting Effects and Conditions (XReq #3)
The current crafting system is not scalable if the developer intends to implement any sort of conditions or skill requirements. As an example let's take a crafting table of 12 crafts. If we want each item to have some unique condition then we'd need to create 2^12 crafting tables (allowed or disallowed state for each craft required). Another thing is that crafting doesn't allow us to cast effects - this severely limits options. If you wanted to for example grant crafting experience for each time you craft something, you're out of luck.

- Condition
`Add the ability to add condition lists for each craft. Also add a checkbox on whether or not we'd like to hide the item from a workstation or simply gray it out as an option when the condition's not fulfilled.`
- Common Event after craft
`Add the ability to trigger a common event after a craft successfully completes.`
- IN-GAME Crafting Station
`Do not show items that don't have their condition fulfilled or gray them out as an option. If all options are hidden (not grayed out, though) throw some default message to chat saying something like "You're unable to use this crafting station.", as we shouldn't show an empty crafting menu to the player.`

**Additional context**
This proposed feature's not complicated and it would grant a lot more options to the user in regards to the crafting system. It's also very general, so it'd be of use to pretty much any game that uses crafting.

**Personal Note**
I've discovered the engine recently and I'm examining its functionality for my project. I come from a Unity/Godot environment, but my plan to make an MMO made me go looking for alternatives better than just coding the entire thing myself (from my experiences, I really hate dealing with net code). I'm impressed with the engine's back-end as it handles all of the networking bull-crap I didn't want to deal with. Unfortunately, every time I try to implement something using the editor, I find it extremely lacking in basic functionality, which is why I'll be creating a few feature suggestions which I think would massively improve the engine's functionality and are really general to the entire MMORPG experience. I want to contribute to the project in the future, but the lackluster (most of the time completely absent) use of comments in the code and no code documentation multiply the amount of time I'll need to grasp the code base a hundred-fold.

jcsnider:
Resolves #852 since #999 was closed.

This hides the crafts the user is not able to make, so they are unaware that they are being hidden. While I remain open to better clarity over requirements and when things are unlocked I feel like that can be leave to the game dev for now. This pr just implements the initial functionality.


Arufonsu (approved):

Intersect.Server/Entities/Player.cs
```diff
@@ -3609,6 +3609,25 @@ public void CraftItem(Guid id)
...
+                var table = CraftingTableBase.Get(CraftingTableId);
+                if (table == null)
+                    return;
+
+                var craftToCheck = CraftBase.Get(id);
+
+                if (craftToCheck == null)
+                    return;
+
+                if (!table.Crafts.Contains(craftToCheck.Id))
```
lodicolo, line 3621: Isn't `craftToCheck.Id` equal to `id`? If so, this check can be moved to _before_ `CraftBase.Get(id)` and avoid an expensive lookup call.

Intersect.Server/Entities/Player.cs
```diff
@@ -3609,6 +3609,25 @@ public void CraftItem(Guid id)
...
+                    return;
```
lodicolo, line 3622: In general I would prefer we keep return statements inside of explicit braces.

lodicolo (changes requested):

PR #1347, opened 2022-06-26
Custom Animations

### LINKED ISSUE #745: Option to use sprite animations on spells
Having a option that allows to use a sprite animation from the existing ones (_attack, _weapon, _shoot) when using a spell.

A dropdown menu in the spell editor that allows to choose which of the three animations to use.

**Additional context**
This would allow to have a more immersive game with sprite animations for spells and not only for weapons.

### LINKED ISSUE #1199: feature: Animation Manager
### Discussed in https://github.com/AscensionGameDev/Intersect-Engine/discussions/1191

<sup>Originally posted by **WeylonSantana** April  7, 2022</sup>

Sit down and get your coffee, this is a discussion for developers (not that I'm one), I appreciate if you leave your opinion so we can reach a logical and quick conclusion.

Currently the animations are defined and coded internally. We have _attack, _cast, _idle, _shoot, _weapon and that doesn't bring the developers a dynamic interaction between the developer and the players experience.

"But how Weylon? Explain yourself better!"

So, it is normal that we have different types of weapons in our games, but only 1 animation defined for that type of weapon. Considering the case of the _shoot animation, if we have a bow and arrow we can make this animation, but this animation will be the same for shooting shuriken, shooting crossbows, throwing knives, because these other types of weapons require different animations, and we have only 1 _shoot animation, which is not dynamic. The same goes for _weapon for example, different types of weapons require different types of animation.

So we got to the point that animation management needs to become dynamic

I'm going to present a solution (which may or may not be crazy, because I'm not a developer), and you developer reading this feel free to refute me, throw my idea in the trash, or improve it, it doesn't matter, I'd just like it we came to a conclusion that possibly some well-known programmer would do it (since I have no knowledge so I won't do it)

My idea is that the animations were defined in the server's config, through an array. And for each weapon's attack animations to be set in the item editor.
In the item editor there is already "attack animation" that runs an animation in front of the attack. This for ages has been a simple solution for the intersect, but today it is no longer necessary due to _attack or _weapon, but which is still useful and can complement these aforementioned animations by adding more effect and sound.

An array on the server - In the same way that we can put items into categories, or something similar, we would define the suffixes that sprite images should have on the server. For example:

"SpriteAnimations" : [
   "lance",
   "sword",
   "bow",
   "crossbow",
   ...
]

So with some magic (programming obviously), we would define in the item editor, which animation it will use, and then with more magic (more fucking programming) the client would get the base sprite that is being used by the class editor and would try to search for the sprite with the selected suffix to render the animation at the time of the attack.

In idiotic terms: we create the animations on the server, we define it in the editor, so on the client it will search for "player_sword.png", "player_bow.png" etc.

Could I be being an idiot and talking nonsense? Yes, but I'm being a pretty logical idiot, aren't I?

Well, in my view this would make the animations dynamic, each weapon could have its own animation (or for the laziest like me, use the same animation, but not stick to it).

Logically there should be a lot of bug prevention if the image doesn't exist, etc. And also configuration (like the ones that currently exist) for the amount of frames.

You can have a dictionary approach to this: Dictionary<Animation, Number of Frames>
"SpriteAnimations" : [
   "lance" => 4
   "sword" => 2
   "bow" => 3
   "crossbow" => 4
   ...
]

**Additional context**
But anyway, I started the discussion, and I hope there can be more contributions to a really viable solution (I don't think mine really is), it was just a kick-off to make games made with the intersect visually better in their quality, and more fun for the player experience.

I'd like to say that I think it's fair that only the walking animation should be hardcoded.
About the cast, it should follow the same model as above, because different spells might need different casts, but I don't know how it would be implemented.
And about idle, we could have more than one, but that would be too luxury haha</div>

WeylonSantana:
resolves #745
resolves #1198
resolves #1199

Added to the Class Editor, Item Editor and Spell Editor, a new combobox, allowing users to select their own custom "_animations", for attack(unarmed and armed) and cast.

In the case of items, if the weapon has a custom animation
the _shoot will be ignored and the custom animation will be prioritized.
_shoot will only be used if the weapon is projectile and does not have custom animation.

Priority will always be given to custom animation, if not set, the engine will look for the default "_animation" (attack, cast, weapon and shoot), if none then uses default 3rd frame for attacks and nothing for cast.

Paperdolls support the new animations defined in the editor.

CAUTION: The files from this pr must follow the format "<base-name>_<animation>_<custom-animation>.png"
Examples:
base-male.png
base-male_attack.png ==> Default sprite attack animation
base-male_attack_sword.png ==> Custom animation
base-male_cast.png ==> Default sprite cast animation
base-male_cast_kamehameha.png ==> Custom animation
and so on...

The "_" is used as a separator between the base filename, animation and custom animation
"base_male_attack.png" will not work for custom animation.

Modifications to Entity.cs and Player.cs are authored by: @lodicolo


lodicolo:
@WeylonSantana Can you provide a clip demonstrating the custom animations?

WeylonSantana:
All working now

Intersect (Core)/GameObjects/ClassBase.cs
```diff
@@ -91,6 +91,8 @@ public AnimationBase AttackAnimation
 
         public int AttackSpeedValue { get; set; }
 
+        public string SpriteAttackAnimation { get; set; }
```
lodicolo, line 94: I think this should be `AttackSprite` instead.

Intersect (Core)/GameObjects/ItemBase.cs
```diff
@@ -132,6 +132,8 @@ public AnimationBase EquipmentAnimation
 
         public int AttackSpeedValue { get; set; }
 
+        public string SpriteWeaponAnimation { get; set; }
+
```
lodicolo, line 136: `WeaponSprite`

Intersect (Core)/GameObjects/SpellBase.cs
```diff
@@ -108,6 +108,8 @@ public string JsonCastRequirements
 
         public string CannotCastMessage { get; set; } = "";
 
+        public string SpriteCastAnimation { get; set; }
+
```
lodicolo, line 112: `CastSprite`

Intersect.Editor/Forms/Editors/frmItem.Designer.cs
```diff
@@ -214,6 +215,7 @@ private void InitializeComponent()
             this.toolStripItemPaste = new System.Windows.Forms.ToolStripButton();
             this.toolStripSeparator3 = new System.Windows.Forms.ToolStripSeparator();
             this.toolStripItemUndo = new System.Windows.Forms.ToolStripButton();
+            this.txtSpriteAttackAnimation = new DarkUI.Controls.DarkTextBox();
```
lodicolo, line 218: I think this should be a dropdown like other sprite selectors.

Intersect.Editor/Forms/Editors/frmSpell.Designer.cs
```diff
@@ -175,6 +176,7 @@ private void InitializeComponent()
             this.btnClearSearch = new DarkUI.Controls.DarkButton();
             this.txtSearch = new DarkUI.Controls.DarkTextBox();
             this.lstGameObjects = new Intersect.Editor.Forms.Controls.GameObjectList();
+            this.txtSpriteCastAnimation = new DarkUI.Controls.DarkTextBox();
```
lodicolo, line 179: I think this should be a dropdown like other sprite selectors.

Intersect.Editor/Forms/Editors/frmClass.Designer.cs
```diff
@@ -174,6 +175,7 @@ private void InitializeComponent()
             this.toolStripItemUndo = new System.Windows.Forms.ToolStripButton();
             this.mnuExpGrid = new System.Windows.Forms.ContextMenuStrip(this.components);
             this.btnExpPaste = new System.Windows.Forms.ToolStripMenuItem();
+            this.txtSpriteAttackAnimation = new DarkUI.Controls.DarkTextBox();
```
lodicolo, line 178: I think this should be a dropdown like other sprite selectors.

lodicolo (changes requested):
@WeylonSantana I would like a demo video showing this with paperdolls too.

WeylonSantana:
moved to #1422

WeylonSantana:
using this
image
and this
image

for testing

https://user-images.githubusercontent.com/63019821/187078265-6c3a1120-96e0-4285-992d-dfce1f0d1173.mp4


https://user-images.githubusercontent.com/63019821/187078358-a8562fce-17b5-4f3b-a228-5051f0606912.mp4

lodicolo (approved):

Issue #1281, opened 2022-06-14
feature: Screen Shake


Erald557:
**Is your feature request related to a problem? Please describe.**
When you attack a monster or take damage sometimes there isnt much effect to the action and its a bit boring 

**Describe the solution you'd like**
A small tint of screen shake whenever we attack or take damage it will give a nice combat effect, also its a good alert to tell if you are getting attacked.

**Additional context**
Whether the game devs want to add it it should be optional

PR #1348, opened 2022-06-28
Animation camera shake feature

lucas-tardivo:
Solving the feature request of #1281

Created a new animation feature for camera shaking. 
- Animations now have a new Integer property "CameraShakeInterval" that represents how many ms the camera will shake.
- If there is two animations shaking the camera, only the longer one will prevail.
- Camera shake intensity can be configured at client options. (Also can be disabled).

Long time i promised this feature, sorry, but now there is.

### DIFF
--- a/Intersect.Client.Framework/Database/GameDatabase.cs
+++ b/Intersect.Client.Framework/Database/GameDatabase.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace Intersect.Client.Framework.Database
 {
@@ -38,6 +38,8 @@ namespace Intersect.Client.Framework.Database
 
         public bool PlayerOverheadInfo;
 
+        public int CameraShakeIntensity;
+
         //Saving password, other stuff we don't want in the games directory
         public abstract void SavePreference(string key, object value);
 
@@ -72,6 +74,7 @@ namespace Intersect.Client.Framework.Database
             NpcOverheadInfo = LoadPreference("NpcOverheadInfo", true);
             PartyMemberOverheadInfo = LoadPreference("PartyMemberOverheadInfo", true);
             PlayerOverheadInfo = LoadPreference("PlayerOverheadInfo", true);
+            CameraShakeIntensity = LoadPreference("CameraShakeIntensity", 2);
         }
 
         public virtual void SavePreferences()
@@ -91,6 +94,7 @@ namespace Intersect.Client.Framework.Database
             SavePreference("NpcOverheadInfo", NpcOverheadInfo.ToString());
             SavePreference("PartyMemberOverheadInfo", PartyMemberOverheadInfo.ToString());
             SavePreference("PlayerOverheadInfo", PlayerOverheadInfo.ToString());
+            SavePreference("CameraShakeIntensity", CameraShakeIntensity.ToString());
         }
 
         public abstract bool LoadConfig();
--- a/Intersect.Client/Interface/Shared/SettingsWindow.cs
+++ b/Intersect.Client/Interface/Shared/SettingsWindow.cs
@@ -96,6 +96,12 @@ namespace Intersect.Client.Interface.Shared
 
         private readonly LabeledCheckBox mLightingEnabledCheckbox;
 
+        private readonly ImagePanel mCameraShakeBackground;
+
+        private readonly Label mCameraShakeLabel;
+
+        private readonly ComboBox mCameraShakeList;
+
         // Audio Settings.
         private readonly HorizontalSlider mMusicSlider;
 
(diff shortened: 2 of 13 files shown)


Intersect.Server/Migrations/Game/20220628180051_AddedCameraShakeInterval.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public partial class AddedCameraShakeInterval : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<int>(
+                name: "CameraShakeInterval",
+                table: "Animations",
+                nullable: false,
+                defaultValue: 0);
```
Arufonsu, line 13: We still have to get rid of a couple of 'defaultValue' column properties in order to allow migrations to MySQL, adding more to it is should be a no no since we want to get rid of this issue eventually.

Intersect.Client.Framework/Database/GameDatabase.cs
```diff
@@ -38,6 +38,8 @@ public abstract partial class GameDatabase
 
         public bool PlayerOverheadInfo;
 
+        public int CameraShakeIntensity;
```
Arufonsu, line 41: Considering that you have 5 options for the intensity here, i would say that an Enum list should be used instead of int values.

Intersect.Server/Migrations/Game/GameContextModelSnapshot.cs
```diff
@@ -786,7 +788,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.ToTable("Time");
                 });
 
-            modelBuilder.Entity("Intersect.Server.Maps.MapInstance", b =>
+            modelBuilder.Entity("Intersect.Server.Maps.MapController", b =>
```
Arufonsu, line 791: why are you changing this?
pretty sure that MapInstance is what we use now since the latest changes on map's instancing 😕

Intersect.Editor/Localization/Strings.cs
```diff
@@ -642,6 +642,8 @@ public partial struct AnimationEditor
 
             public static LocalizedString simulatedarkness = @"Simulate Darkness: {00}";
 
+            public static LocalizedString camerashake = @"Camera Shake (ms): {00}";
```
Arufonsu, line 645: Don't forget to add: 

`[JsonProperty(NullValueHandling = NullValueHandling.Ignore)]`

Also use Pascal casing (C# Coding Conventions)

Arufonsu (changes requested):
> * Camera shake intensity can be configured at client options. (Also can be disabled).

This is fine, but i personally dislike the fact that this **only allows to enable/disable the shaking effect for every single attack animation?**, i would personally like it to be configurable as per item/weapon in the editor, and would also like it to have a boolean toggle so the shaking effect only triggers with critical hits optionally, and not on every single animation attack which can be quite annoying after a while for some.

Intersect.Server/Migrations/Game/GameContextModelSnapshot.cs
WeylonSantana: He is not changing that, creating any migration causes this line to change, I can say from experience, every time I create a migration, I manually change this line to MapInstance again, I believe this edit should be done here too, and something should be done about this change that occurs every migration

Intersect.Client/Interface/Shared/SettingsWindow.cs
```diff
@@ -812,6 +861,33 @@ private void SettingsApplyBtn_Clicked(Base sender, ClickedEventArgs arguments)
                 Globals.Database.TargetFps = newFps;
             }
 
+            var newCameraShakeIntensity = 0;
+            if (mCameraShakeList.SelectedItem.Text == Strings.Settings.CameraShakeIntensityNone)
```
lodicolo, line 865: Couldn't you just use the selected index and then add/subtract to get the correct value?

That said, I almost think this should be a number slider instead.

Intersect.Editor/Localization/Strings.cs
lodicolo, line 645: Name this `CameraShake`, I don't expect you to change all of the other string names but I don't want to perpetuate the bad string names.

Intersect.Server/Migrations/Game/20220628180051_AddedCameraShakeInterval.cs
lodicolo: What issue are you referring to here? @Arufonsu

Intersect.Server/Migrations/Game/GameContextModelSnapshot.cs
lodicolo: Since this was changed in JCs PR that got merged, this change should be omitted from this PR.

lodicolo (changes requested):
I am not sure I'm on-board with a 4-tier shake intensity system, number or enum my opinion remains the same.

- I feel like the user should have their own intensity slider, similar to volume, that goes from 0 to 10 (0 disabled, 10 is full) -- this would get "multiplied" against the camera shake intensity specified for the given trigger
- Various different camera shake triggers will shake the screen with varying intensity
  - e.g. A sword collision will not shake the camera as much as an earthquake cutscene
- Camera shake should not be tied to animations
  - melee attacks (which are not animations), animation-less events, etc. are all legitimate camera shake triggers
  - what happens when we have 10 different animations/camera shake triggers all playing on the screen, all with camera shake on them? which camera shake takes priority? how do we know that the camera shake trigger should actually impart a camera shake on the viewer? (e.g. an earthquake will cause camera shake for all viewers on the map, but a sword collision will only shake the camera for the attacker and the recipient of the attack)
  
I am not opposed to us adding the system that does the actual shaking of the camera _without_ any ability to trigger it, and then separate PRs for different shake triggers in the event you only want to implement one at first.

Possible camera shake triggers:
- when melee attacks actually hit (misses wouldn't shake)
- when melee attacks hit an entity that is actively blocking (or has some sort of knockback on hit)
- when knockback is applied to a user
- some step in an event (for cutscenes)
- casting spells
- on spell hit (shakes the camera for the recipient)
  - optionally shakes the camera for all entities within a specific range whether or not they were directly impacted by the spell, for example a fire bomb spell could shake the screens for everyone within 5 tiles
- when the person is struck (by anything?)

Again, I don't think shake should be tied directly to animations. It _should_ be tied to what triggers the animations though (and more). We also need logic for when to ignore certain shake triggers.

Intersect.Server/Migrations/Game/20220628180051_AddedCameraShakeInterval.cs
lodicolo: After discussion with Aru in Discord, I think that this defaultValue is ok because it is for an int column type, while the migration error is for BLOB, TEXT, GEOMETRY, and JSON column types.

This doesn't however change all of the other comments I made last night.

lodicolo:
To follow up on my comments from last night, we need to have a larger discussion in #github-features or #engine-development on the Discord server to better describe this feature and how it should be implemented.

Intersect.Editor/Localization/Strings.cs
Arufonsu: For a reference: we need LocalizedString properties to follow the PascalCase naming convention 👍🏽 
Also, make sure to add the following JsonProperty:
`[JsonProperty(NullValueHandling = NullValueHandling.Ignore)]`

Intersect.Server/Migrations/Game/20220628180051_AddedCameraShakeInterval.cs
Arufonsu: Yup, like @lodicolo , this was discussed in discord, my comment doesn't applies to int column types so feel free to ignore this comment review @lucas-tardivo 👍🏽

Issue #1308, opened 2022-06-22
bug: Set Graphic not working


GabrielTadachi:
**Describe the bug**
Set graphic is not actually setting the graphic that I say to set.

**Version Encountered**
Last Prerelease

**To Reproduce**
1. Create an event and click on Set Move Route
2. Click to Change Graphic and select the sprite you want
3. Save Event and try it. The sprite will not change

**Expected behavior**
On interacting with the event, the graphic should change, as it changed in previous versions.

**Screenshots**
Output after Move Route:
![image](https://user-images.githubusercontent.com/25730747/175134772-62e45890-0104-4e44-938a-6a302020f9fd.png)

Event configuration:
![image](https://user-images.githubusercontent.com/25730747/175134865-9c42dad0-679c-4621-b953-fb934cbde26a.png)
![image](https://user-images.githubusercontent.com/25730747/175134824-99e59e81-00f9-4b55-bf8c-ca697d34987b.png)
![image](https://user-images.githubusercontent.com/25730747/175136200-f96e0300-be49-4f5a-8e07-987f15493b9b.png)

**Logs**
If there are error logs please attach them below.

**Operating System:**
Windows, Mac, Or Linux

**Additional context**

WeylonSantana:
I confirm, it's a bug

PR #1351, opened 2022-06-29
Fix updating the event graphic resolving issue #1308

lucas-tardivo:
Basically the EventGraphic object does not change the event sprite according to the Filename property.
Worked for both Sprite/Tileset.

This pull request fix the issue  #1308

### DIFF
--- a/Intersect.Client/Entities/Events/Event.cs
+++ b/Intersect.Client/Entities/Events/Event.cs
@@ -62,6 +62,11 @@ namespace Intersect.Client.Entities.Events
             _drawCompletedWithoutTexture = Graphic.Type != EventGraphicType.Tileset;
 
             base.Load(packet);
+
+            if (!string.IsNullOrEmpty(Graphic?.Filename))
+            {
+                Sprite = Graphic.Filename;
+            }
         }
 
         public override bool Update()
@@ -77,11 +82,6 @@ namespace Intersect.Client.Entities.Events
 
         protected bool TryEnsureTexture(out GameTexture texture)
         {
-            if (Graphic != null && !string.IsNullOrEmpty(Graphic.Filename) && Sprite != Graphic.Filename)
-            {
-                Sprite = Graphic.Filename;
-            }
-
             if (_drawCompletedWithoutTexture)
             {
                 texture = Texture;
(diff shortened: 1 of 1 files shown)


Intersect.Client/Entities/Events/Event.cs
```diff
@@ -77,6 +77,11 @@ public override bool Update()
 
         protected bool TryEnsureTexture(out GameTexture texture)
```
lucas-tardivo, line 78: Dont know if thats the best solution for this.
Actually this method seems a bit confusing, but as it changes the event texture, sound appropriated.

Intersect.Client/Entities/Events/Event.cs
```diff
@@ -77,6 +77,11 @@ public override bool Update()
...
         {
+            if (Graphic != null && !string.IsNullOrEmpty(Graphic.Filename) && Sprite != Graphic.Filename)
```
lodicolo, line 80: Remove `Sprite != Graphic.Filename` -- the setter for `Sprite` already makes this check.

Intersect.Client/Entities/Events/Event.cs
lodicolo: @lucas-tardivo When the graphic is changed, is it being changed via `Event`'s `Load(EntityPacket)` method (L59)?

If so, we can expand the `Graphic` property's setter to do what you are doing here to set the sprite, but at the time of change instead of checking and re-setting it every time `TryEnsureTexture(out GameTexture)` is called.

lodicolo (changes requested):

Intersect.Client/Entities/Events/Event.cs
lucas-tardivo: > 

I tried that first, but it does'nt work due to the inheritance from Entity class :(

Intersect.Client/Entities/Events/Event.cs
```diff
@@ -62,6 +62,11 @@ public override void Load(EntityPacket packet)
             _drawCompletedWithoutTexture = Graphic.Type != EventGraphicType.Tileset;
 
             base.Load(packet);
+
+            if (!string.IsNullOrEmpty(Graphic?.Filename))
```
lodicolo, line 66: @lucas-tardivo this where I was suggesting to move the fix

lodicolo (approved):

PR #1359, opened 2022-07-08
Fix 1358 "Loot All" across map boundaries

### LINKED ISSUE #1358: bug: Cannot use "Loot All" across map boundaries
### Description

Since instancing was introduced, you could not use the loot box's "loot all" functionality across map boundaries.

### Steps to Reproduce

1. Drop an item across a map boundary
2. Attempt to pick it up with "loot all"

### Version with bug

7.1.22

Pre-instancing

### Affected platforms

You can pick up the items individually in the loot window by clicking on them.

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
Fixes #1358

### DIFF
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -1628,7 +1628,7 @@ namespace Intersect.Server.Networking
                                 tmpInstance.RemoveItem(mapItem);
 
                                 // Try to give the item to our player.
-                                if (player.TryGiveItem(mapItem, ItemHandling.Overflow, false, true, mapItem.X, mapItem.Y))
+                                if (player.TryGiveItem(mapItem, ItemHandling.Overflow, false, -1, true, mapItem.X, mapItem.Y))
                                 {
                                     var item = ItemBase.Get(mapItem.ItemId);
                                     if (item != null)
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -1569,17 +1569,20 @@ namespace Intersect.Server.Networking
                     foreach (var itemMap in map.FindSurroundingTiles(new Point(player.X, player.Y), Options.Loot.MaximumLootWindowDistance))
                     {
                         var tempMap = itemMap.Key;
-                        if (tempMap.TryGetInstance(player.MapInstanceId, out var tempMapInstance))
-                        {
-                            if (!giveItems.ContainsKey(itemMap.Key))
-                            {
-                                giveItems.Add(tempMap, new List<MapItem>());
-                            }
 
-                            foreach (var itemLoc in itemMap.Value)
-                            {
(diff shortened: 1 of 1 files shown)


Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1590,59 +1593,62 @@ public void HandlePacket(Client client, PickupItemPacket packet)
                 foreach (var itemMap in giveItems)
                 {
                     var tempMap = itemMap.Key;
-                    var toRemove = new List<MapItem>();
-                    foreach (var mapItem in itemMap.Value)
+                    if (tempMap.TryGetInstance(player.MapInstanceId, out var tmpInstance))
```
lodicolo, line 1596: Can this just be inverted and `continue;`d? Rather than indenting everything inside. This will also make this massively easier to review.

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1569,14 +1569,17 @@ public void HandlePacket(Client client, PickupItemPacket packet)
                     foreach (var itemMap in map.FindSurroundingTiles(new Point(player.X, player.Y), Options.Loot.MaximumLootWindowDistance))
                     {
                         var tempMap = itemMap.Key;
-                        if (!giveItems.ContainsKey(itemMap.Key))
+                        if (tempMap.TryGetInstance(player.MapInstanceId, out var tempMapInstance))
```
lodicolo, line 1572: Same inversion comment as below.

lodicolo (changes requested):
I'd like the two things inverted that I mentioned, and then I will review the PR.

That said, before merging this PR I will want a video demonstrating this fix and/or verification from another person saying this works.

Intersect.Server/Networking/PacketHandler.cs
AlexVild: done

Intersect.Server/Networking/PacketHandler.cs
AlexVild: done

Intersect.Server/Networking/PacketHandler.cs
```diff
@@ -1598,29 +1609,29 @@ public void HandlePacket(Client client, PickupItemPacket packet)
...
+                        if (mapItem.Owner == Guid.Empty || Timing.Global.Milliseconds > mapItem.OwnershipTime)
+                        {
+                            // The ownership time has run out, or there's no owner!
+                            canTake = true;
+                        }
+                        else if (mapItem.Owner == player.Id)
+                        {
+                            // The current player is the owner.
+                            canTake = true;
+                        }
```
lodicolo, line 1623: I just realized there's no point in the else-if, the entire thing can be reduced down into

`var canTake = mapItem.Owner == Guid.Empty || mapItem.Owner == player.Id || Timing.Global.Milliseconds > mapItem.OwnershipTime;` but it was already an unnecessary clown fiesta so I will just make this change on my own in a separate PR.

lodicolo (approved):

Issue #454, opened 2020-12-28
Projectiles Bug : Projectiles don't ignore resource


Spectreye:
**Describe the bugs**
1) When you use a spell or weapon which use projectile that ignore collision with active and inactive ressources, the animation fade away on ressources who are not walkable before harvesting, however the damage are still done.

2) When you use a spell with projectiles and change map, the projectiles disappear from view, sometimes they even do not do any damage when fired from another map.

I've made a video where you can see each bug  individually (Look at the slime getting hit by projectiles but not getting any damage)

**Version Encountered**
0.6.2.453

**To Reproduce**
Missiles Animation Fading
1. Make a spell with projectile which ignore collision with active and inactive ressources.
2. Make a ressource which is not walkable before harvesting (and random skin)
3. Try to launch a projectile through it.

1. Make a spell with slow missiles and two map
2. Launch the projectiles and go on the other map
3. See error

**Expected behavior**
1)Projectiles animation should not end when passing through an active ressources if "Ignore collison with active ressources" is checked.

2) Projectiles should not disappear visually (And sometimes not doing damage at all) when moving to another map or shooting from another map.

**Video of bug**
https://ascensiongamedev.com/resources/filehost/107d0ffec3625212bf1079f5a76e4815.mp4

**Operating System:**
Windows

blinkuz:
What disappearing when you move to another map happens between connected maps? Because this behavior is very strange unless what is done is to eliminate all the entities / objects that you are far away, have you tried placing a player on the other map and launching the projectile to confirm that the other player can see it?

Spectreye:
Player on other map can see it, but if they change map, the projectiles will visually disappear too. I'm almost sure it has to do with the way the code deal with transition between map.

blinkuz:
I just tried and I can confirm this error too, this has happened to me with my projectiles.

Spectreye:
Update : The animation dysfonction seems to be corrected in 0.7.0.108.

Still, the Animation is still fading when colliding with a ressources, even if the "Ignore collision" with active ressources is checked.

WeylonSantana:
The second bug is fixed, shooting projectiles and traversing the map doesn't make them disappear anymore.

but, the first bug still occurs, a projectile does not traverse a resource even when the checkbox is marked to ignore

Issue #1100, opened 2022-03-06
chore: refactor Entity#CanMove()'s return types


AlexVild:
**Is your feature request related to a problem? Please describe.**
The `CanMove` method on an `Entity` returns a variety of integers as it's return type:

```cs
/// <summary>
///     Determines if this entity can move in the direction given.
///     Returns -5 if the tile is completely out of bounds.
///     Returns -3 if a tile is blocked because of a Z dimension tile
///     Returns -2 if a tile is blocked by a map attribute.
///     Returns -1 for clear.
///     Returns the type of entity that is blocking the way (if one exists)
/// </summary>
/// <param name="moveDir"></param>
/// <returns></returns>
```

This is, uh, disgusting, and very very unclear when referenced in areas of the code removed from this comment block.

**Describe the solution you'd like**
A more elegant solution might be to leverage an `out` variable for the blocking REASON (which should be some enum, not just an int), and returning a `bool` from the method.

This way, we don't need to check if `CanMove == -1`, which is strange to read, and can replace its usages with `if (CanMove())`. We can leverage an out variable that can contain a reason for areas where we want to make sure we know the reason why the movement isn't successful (projectiles, pathfinding).

PR #1360, opened 2022-07-08
Projectile Changes and New Options to Editor

### LINKED ISSUE #652: Projectiles blocked by friendly targets
Projectiles could have the ability to cross party members, on the contrary, in a narrow passages, whoever has projectiles style combat behind someone who fights hand-to-hand will never participate in the battle

Projectiles pass through party members without causing damage or disappearing, this should happen without the need for piercing to be activated.

**Additional context**
1 - Open 2 clients
2 - Place two players on the same tile and face the same side
3 - shoot a projectile, the projectile does not hit the enemy ahead

WeylonSantana:
Summary:
Projectiles can pass through players if they are on a safe map.
Projectiles can pass through players if they are partying with the target.
Added new options to Projectile Editor to check hook.
and refactor things related to projectile checks.

Details:
I created the GrappleHookOptions in the projectile editor so that we can choose what should hook the player.
I modified IsTileBlocked on the client side to return an enum instead of an int, to make the code more readable (maybe a kick-start to #1100)
I added one more parameter to IsTileBlocked to know which entity invoked the method and thus make the checks better (this fixes bug #454) (I accept better suggestions)
Added one more check in CheckForCollision in Projectile on the server, not to enter the if if entity[z] is the owner player (for some reason he checked this when casting the spell/projectile)
I refactored the HitEntity into ProjectileSpawn on the server:
- Removed the scalingStat that wasn't being used (I believe, well, I didn't see that in both tryattack() methods).
- I created a method to check the new options
- And another method to execute the hook, to decrease the code repetition
- Added an extra check for when the player shoots the projectile and the map is safe or if they are partying with the target (this fixes #652)


WeylonSantana:
Full test on everything above, found 2 bugs
1 - Grapple doesn't work with events yet
2 - When the projectile collides with any map attribute, it despawns, needs better checks.
The others things is fine...


https://www.ascensiongamedev.com/resources/filehost/74add1e35569b726cdb7b5c6fe659af3.mp4

PR #1363, opened 2022-07-09
Fix 454/652 and new grapple options

WeylonSantana:
Summary:
Projectiles can pass through players if they are on a safe map.
Projectiles can pass through players if they are partying with the target.
Added new options to Projectile Editor to check hook.
and refactor things related to projectile checks.

Fixes for #454 and #652.


WeylonSantana:
Grapple on event still doesn't work, but for the rest everything is 100%.

https://www.ascensiongamedev.com/resources/filehost/1bf97547d000aac98cd67b85dc5457e7.mp4

WeylonSantana:
Removed collision when target is on same tile as player

https://www.ascensiongamedev.com/resources/filehost/60c66dfb29425e7a181ae345899f9d5b.mp4

Intersect.Server/Entities/ProjectileSpawn.cs
```diff
@@ -176,6 +164,51 @@ public bool HitEntity(Entity en)
...
+                return ProjectileBase.GrappleHookOptions.Contains(Enums.GrappleOptions.Player);
+            }
+            else if(entityType == typeof(Npc))
+            {
+                return ProjectileBase.GrappleHookOptions.Contains(Enums.GrappleOptions.NPC);
+            }
+            else if(entityType == typeof(Resource))
+            {
+                return ProjectileBase.GrappleHookOptions.Contains(Enums.GrappleOptions.Resource);
+            }
```
lodicolo, line 192: Use this format instead, do not do this if stuff
```csharp
switch (en)
{
    case Player _:
        return ...;
    ... etc ...
}

**Review by lodicolo — CHANGES_REQUESTED** [review]

File: Intersect.Server/Entities/Projectile.cs [hunk 2]
```diff
@@ -357,10 +362,10 @@ public bool CheckForCollision(ProjectileSpawn spawn)
                 }
 
                 //Check for grapplehooks.
-                if (attribute != null &&
-                    attribute.Type == MapAttributes.GrappleStone &&
-                    Base.GrappleHook == true &&
-                    !spawn.Parent.HasGrappled)
+                if (attribute.Type == MapAttributes.GrappleStone &&
+                    Base.GrappleHookOptions.Contains(GrappleOptions.MapAttribute) == true &&
```
lodicolo, line 366: The `== true` check is implicit and not needed

Intersect.Server/Entities/ProjectileSpawn.cs
```diff
@@ -66,80 +66,74 @@ public bool IsAtLocation(Guid mapId, int x, int y, int z)
...
+        public bool HitEntity(Entity targetEntity)
         {
-            var targetEntity = en;
-            if (targetEntity is EventPageInstance) return false;
-
-            var scalingStat = Enums.Stats.StatCount;
-
-            if (Parent.Spell != null && Parent.Spell.Combat != null)
+            Player targetPlayer = null;
+            if (targetEntity is Player)
```
lodicolo, line 72: `if (targetEntity is Player targetPlayer)` if you can add more logic into the body of the if statement, or `var targetPlayer = targetEntity as Player;` otherwise. 

`(Player) targetEntity` I think will throw a cast exception, but `targetEntity as Player` will just result in null if it is not a player.

Intersect.Server/Entities/ProjectileSpawn.cs
```diff
@@ -66,80 +66,74 @@ public bool IsAtLocation(Guid mapId, int x, int y, int z)
...
+                            {
+                                return false;
+                            }
+
                             return true;
                         }
                     }
                 }
-                else if (targetEntity.GetType() == typeof(Resource))
+                else if (targetEntity is Resource)
```
lodicolo, line 125: `targetEntity is Resource targetResource` and then use `targetResource` instead of the re-casted value below.

lodicolo (changes requested):

Intersect.Server/Entities/ProjectileSpawn.cs
```diff
@@ -176,22 +172,20 @@ public bool ShouldHook(Entity en)
...
 
-            return false;
+                case Npc _:
+                    return ProjectileBase.GrappleHookOptions.Contains(Enums.GrappleOptions.NPC);
+
+                case Resource _:
+                    return ProjectileBase.GrappleHookOptions.Contains(Enums.GrappleOptions.Resource);
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(en));
```
lodicolo, line 187: I think this should just be `new ArgumentException($"Unsupported entity type {en.GetType().FullName}", nameof(en))` instead of an out-of-range.

Intersect.Server/Entities/ProjectileSpawn.cs
```diff
@@ -122,16 +118,16 @@ public bool HitEntity(Entity targetEntity)
...
+                        !targetResource.IsDead() && !ProjectileBase.IgnoreActiveResources)
                     {
-                        if (Parent.Owner is Player && !((Resource) targetEntity).IsDead())
+                        if (Parent.Owner is Player && !targetResource.IsDead())
                         {
-                            Parent.Owner.TryAttack(targetEntity, Parent.Base, Parent.Spell, Parent.Item, Dir);
+                            Parent.Owner.TryAttack(targetResource, Parent.Base, Parent.Spell, Parent.Item, Dir);
 
-                            if (Dir <= 3 && ShouldHook(targetEntity) && !Parent.HasGrappled)
+                            if (Dir <= 3 && ShouldHook(targetResource) && !Parent.HasGrappled)
```
lodicolo, line 130: I think this can be simplified by just allowing multiple `return true;`s.

We're checking `IsDead()` 3 times here when realistically we should check once.

Pseudocode:

```
if (dead) {
  if (!IgnoreExhaustedResources) {
    return true;
  }
} else if (!IgnoreActiveResources) {
  if (Parent.Owner is Player) {
    ... tryattack ...
    ... if dir <= 3 ...
    
  }
  
  return true;
}

lodicolo (changes requested):

lodicolo (approved):

Issue #1280, opened 2022-06-14
bug: Buff spell kills player instantly


xXEsiXx:
**Describe the bug**
Buffs casted to the player instantly kills them. Problem to fix this is remove the scaling amount from 100 to 0 and 1.50 crit multiplier

**Version Encountered**
V0.7.1.020

**To Reproduce**
1. Create a normal spell
2. Create Buff stats
3. Get in game
4. Cast it and it will instant kill you

**Expected behavior**
If you dont set the scaling amount from 100 to 0 and 1.50 crit multiplier it will instant kill the player

**Screenshots**

**Logs**
If there are error logs please attach them below.

**Operating System:**
Windows

**Additional context**

Arufonsu:
@xXEsiXx try the following (use a minux prefix)

![imagen](https://user-images.githubusercontent.com/17498701/174466893-c713f774-9e35-4b3e-8904-f2d753cd7db6.png)


expected result:

https://user-images.githubusercontent.com/17498701/174466866-40219ab3-b06f-4de6-9131-68ac1912153d.mp4

let us know if that works for you!

WeylonSantana:
Closing because the user confirmed not to be a bug

PR #1367, opened 2022-07-09
Changing the default spell scaling

WeylonSantana:


lodicolo (approved):

lodicolo:
This adjusts the default value that was "causing" #1280 (not a bug, the value just needed to be changed).

PR #1370, opened 2022-07-10
NPC getting aggro when targeted by player spell

### LINKED ISSUE #1369: bug: Npc doesn't get aggressive when debuffed
### Description

If the npc is targeted by a spell, which does not damage, only debuff, he does not get aggressive.

### Steps to Reproduce

Create a spell that does no damage and use it on the npc

What should happen is he gets aggressive

### Version with bug

v7.1.22

Unknown

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

WeylonSantana:
Fixes #1369


Arufonsu (approved):
Tested to be working as described with status changing spells 👍🏽

lodicolo (approved):

PR #1375, opened 2022-07-12
FIX: 1374 - Too many needs map requests

### LINKED ISSUE #1374: bug: We send too many "SendNeedMap" packets when traversing maps
### Description

We don't always listen to our list of processing map requests on the client when traversing to new maps and, as a result, make too many requests to the server for new maps.

We can reduce some server overhead by properly utilizing `MapInstance.MapRequests`, which is a dictionary containing MapIDs and their maximum request timer.

### Steps to Reproduce

1. Set breakpoint in `Client.PacketHandler#SendNeedMap()`
2. Traverse across map boundaries
3. We should only send 3 requests to the server, for the 3 new maps - we send more than that.

### Version with bug

Main

Unknown

### Affected platforms

No

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
Fix #1374


Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1358,6 +1355,17 @@ public void Dispose(bool prep = true, bool killentities = true)
             Delete();
         }
 
+        public static bool MapNotRequested(Guid mapId) => !MapRequests.ContainsKey(mapId) || MapRequests[mapId] < Timing.Global.Milliseconds;
+
+        public static void UpdateMapRequestTime(Guid mapId)
+        {
+            if (MapRequests.ContainsKey(mapId))
+            {
+                MapRequests[mapId] = Timing.Global.Milliseconds + 2000;
```
lodicolo, line 1364: you actually only need this as the only line in this method -- indexers will add the key if it does not exist, ContainsKey not needed

lodicolo (changes requested):

lodicolo (approved):

PR #1388, opened 2022-07-16
Fix: Re-allows to use Key actions while hovering the GUI

### LINKED ISSUE #1266: bug: when using the interact/attack hotkey (E by default) and hovering over a part of the UI, it does not do the action
**Describe the bug**
When you hover over an item and other UI items, including spells, buttons, etc. you cannot do the interact/attack hotkey action (E by default).

**Version Encountered**

0.7.1.17

**To Reproduce**
1. Go to ANY part of the UI with your cursor
2. Attempt to use the interact / attack key (E by default)
3. See error

**Expected behavior**
When you press E, no matter where your cursor is, it should complete the interact / attack action. For example, if I am attacking and trying to click on potions, or my cursor happens to be placed on a piece of UI, it should complete the action when I press E.

**Screenshots**
221266658dca661ac2ba25232f06a75f

**Logs**
If there are error logs please attach them below.

**Operating System:**
Windows

**Additional context**
It is ALL UI elements, not just clickable buttons.

Arufonsu:
* Re-allows to use ControlMap **Keyboard Key actions** while hovering the GUI (ie: E key for attack/interact).
* Maintains behaviour: right and left **mouse buttons** don't trigger ControlMap actions (only GUI actions and menus) while hovering the GUI.
* Removed unused "control" parameter from ControlMap - Only "key1" and "key2" parameters are used.
* Resolves #1266 

**Video-Demo _( wear headphones / earphones to listen keys and mouse clicks ! )_**

https://user-images.githubusercontent.com/17498701/179371524-e9c90a76-5b0b-431e-8c8e-1defee624e92.mp4

### DIFF
--- a/Intersect.Client/Core/Controls/ControlMap.cs
+++ b/Intersect.Client/Core/Controls/ControlMap.cs
@@ -1,5 +1,4 @@
-﻿using Intersect.Client.Framework.GenericClasses;
-using Intersect.Client.Framework.Input;
+﻿using Intersect.Client.Framework.Input;
 using Intersect.Client.General;
 
 namespace Intersect.Client.Core.Controls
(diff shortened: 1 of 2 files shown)


Arufonsu:
@lodicolo 's  #1391 proposed a better way to solve this issue, which applies to all keybinds 👍🏽

PR #1390, opened 2022-07-16
Chore: Entities Target-box, Overhead Info, and Item Labels

Arufonsu:
Minor chore changes to prevent drawing the following while having the cursor over GUI elements:
 - Entities Target-box.
 - Entities Overhead Information (when hidden).
 - Dropped Item Labels.

**Before this chore PR:**

https://user-images.githubusercontent.com/17498701/179374839-b02880e4-05ec-49e1-a49b-db2297a3623b.mp4

**After this chore PR:**

https://user-images.githubusercontent.com/17498701/179374847-4e47e038-7040-4317-812a-03b672d85aac.mp4

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -2137,13 +2137,14 @@ namespace Intersect.Client.Entities
         // (when they are hidden by the game settings preferences).
         public void DrawOverheadInfoOnHover()
         {
+            if (Interface.Interface.MouseHitGui())
+            {
+                return;
+            }
+
             var mousePos = Graphics.ConvertToWorldPoint(Globals.InputManager.GetMousePosition());
             foreach (MapInstance map in Maps.MapInstance.Lookup.Values)
             {
-                if (Interface.Interface.MouseHitGui())
-                {
-                    return;
-                }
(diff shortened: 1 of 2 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -2140,6 +2140,10 @@ public void DrawOverheadInfoOnHover()
             var mousePos = Graphics.ConvertToWorldPoint(Globals.InputManager.GetMousePosition());
             foreach (MapInstance map in Maps.MapInstance.Lookup.Values)
             {
+                if (Interface.Interface.MouseHitGui())
+                {
+                    return;
+                }
```
lodicolo, line 2146: Can't you just put this at the start of the method instead of in the for-loop?

Intersect.Client/Entities/Player.cs
```diff
@@ -2299,6 +2303,10 @@ public void DrawTargets()
...
```
lodicolo, line 2309: The entire for-loop should be inside of a `!MouseHitGui()` if

lodicolo (changes requested):

lodicolo (approved):

PR #1400, opened 2022-07-24
chore: (Day) Cleansup calls to RichLabel#AddText() by adding a helper…

### LINKED ISSUE #1399: enhancement: Create helper for adding text to RichLabels
### Description

`RichLabel`s seem to primarily be used by creating a `Label` object to use as the _editable_ fields in Gwen's JSON, then using the properties of that `Label` when adding text to the `RichLabel` to allow for text wrapping.

This rather cumbersome logic is being done all over the place, and a simple helper method that takes in a `Label` and uses its properties as needed on a `RichLabel` would eliminate duplicate logic that's all over the place in the UI, and cleanup the code base a tiny bit.

### Intended Use-Case

Slightly more intuitive to add to RichLabels.

### Duplicate Check

AlexVild:
… method that can take in a label

Resolves #1399

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/RichLabel.cs
+++ b/Intersect.Client.Framework/Gwen/Control/RichLabel.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 
 using Intersect.Client.Framework.File_Management;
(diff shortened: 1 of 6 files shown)


Intersect.Client.Framework/Gwen/Control/RichLabel.cs
```diff
@@ -67,6 +67,22 @@ public override JObject GetJson()
...
+        /// <param name="text">Text to add</param>
+        /// <param name="template">Label to use as a template</param>
+        public void AddText(string text, Label template)
+        {
+            if (template == null)
+            {
+                throw new ArgumentNullException(nameof(template));
+            }
+            AddText(text, template.TextColor,
+                template.CurAlignments.Count > 0 ? template.CurAlignments[0] : Alignments.Left,
```
lodicolo, line 82: I'm not sure but would like to see if `template.CurAlignments.FirstOrDefault() ?? Alignments.Left` works -- it may not because `Alignments` is an enum.

lodicolo (changes requested):

Intersect.Client/Interface/Menu/CreditsWindow.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.IO;
+using System.IO;
```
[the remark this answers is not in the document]
AlexVild: Nah, don't know why my VS still insists on those auto-importing :P

Intersect.Client.Framework/Gwen/Control/RichLabel.cs
AlexVild: No dice :P

lodicolo (approved):

PR #1402, opened 2022-07-24
fix: (Day) fixes bug where party sync did not fire common events for …

### LINKED ISSUE #1401: bug: "Sync Party" on variable change event will not trigger Common Events with variable change triggers for party members
### Description

The "variable change" event command has an option to "sync with party". With this option enabled, the variable change will propagate through to all party members.

Currently, this will NOT fire any common events with that variable as a trigger for party members - only the person handling the event.

### Steps to Reproduce

1. Create a common event with a player var trigger
2. Form a party
3. Create an event with a variable change command with party sync enabled
4. Trigger the event with one of the party members
5. One party member will get the common event triggered, others will not.

### Version with bug

7.1.22

Unknown

### Affected platforms

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
…all members

fix #1401

### DIFF
--- a/Intersect.Server/Entities/Events/CommandProcessing.cs
+++ b/Intersect.Server/Entities/Events/CommandProcessing.cs
@@ -1992,7 +1992,7 @@ namespace Intersect.Server.Entities.Events
                         if (partyMember != player)
                         {
                             partyMember.SetVariableValue(command.VariableId, value.Integer);
-                            partyMember.StartCommonEventsWithTrigger(CommonEventTrigger.PlayerVariableChange, "", command.VariableId.ToString());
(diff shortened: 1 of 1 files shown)


Intersect.Server/Entities/Events/CommandProcessing.cs
```diff
@@ -1992,6 +1992,7 @@ Event instance
                         if (partyMember != player)
                         {
                             partyMember.SetVariableValue(command.VariableId, value.Integer);
+                            partyMember.StartCommonEventsWithTrigger(CommonEventTrigger.PlayerVariableChange, "", command.VariableId.ToString());
```
lodicolo, line 1995: `""` -> `string.Empty`

lodicolo (changes requested):

lodicolo (approved):

PR #1420, opened 2022-08-26
fix 1207: properly sets MouseInput on CellText

### LINKED ISSUE #1207: bug: can't properly click/hover on quest and crafting lists.
**Describe the bug**
I noticed that in the BETA version 0.7.1.3 the quest and crafting system seems to be impossible to click on the name of the quest or craft in order to actually work the mission.

**Version Encountered**
0.7.1.3

**To Reproduce**
- Open the quest menu / Craft item list.
- Try clicking / hovering on the quest name / craftable item.
- See the error. (The highlight will not show up, and you cannot click into the quest when hovering on the name's font).

**Screenshots**
b31f4b9ebbc0e0e2d199923621ae42bf

**Expected behavior**
- As you hover over the quest name's font or craft item name's font, we should be able to click on them and properly see the hovering effect taking place. 

**Operating System:**
Windows

Arufonsu:
Should resolve #1207 by adding the boolean parameter 'enableMouseInput' as false by default into the SetCellText method in order to disable the MouseInput within cell's text, which now allows to properly hover/click their parent buttons/elements.

### Preview:

https://user-images.githubusercontent.com/17498701/186951416-a377a016-40e2-4c4d-b58d-a8beb58d4c2c.mp4

image

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/Layout/TableRow.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Layout/TableRow.cs
@@ -352,7 +352,8 @@ namespace Intersect.Client.Framework.Gwen.Control.Layout
         /// </summary>
         /// <param name="column">Column number.</param>
         /// <param name="text">Text to set.</param>
-        public void SetCellText(int column, string text)
+        /// <param name="enableMouseInput">Determines whether mouse input should be enabled for the cell.</param>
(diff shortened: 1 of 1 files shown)


dudevelopement:
Tested here, works perfectly :)

WeylonSantana (dismissed):

WeylonSantana (approved):

lodicolo (approved):

PR #1429, opened 2022-08-31
chore: Refactor Entity#DropItems() to be virtual/abstract

### LINKED ISSUE #1427: enhancement: Refactor `Entity#DropItems()` to be virtual/abstract
### Description

An `Entity` on the server has a method `DropItems()` that is used to drop loot when an entity has been slain.

The problem comes from the fact that dropping items from a slain player and dropping items from an NPC/Resource differ from one another, and, in the future, perhaps we even may want different implementations between NPCs and resources, or whatever other entities may get created.

`DropItems()` should be made virtual (preferred because NPCs and Resources share implementations at the moment) or abstract, and code should be refactored such that the server's `Player` class overrides the Entity's implementation.

### Intended Use-Case

This would clean up the code base a bit. We could also refactor the first half of the method that fetches luck, potentially, to clean up further and improve reusability.

### Duplicate Check

lodicolo:
Resolves #1427

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -27,7 +27,7 @@ using Newtonsoft.Json;
 namespace Intersect.Server.Entities
 {
 
-    public partial class Entity : IDisposable
+    public abstract partial class Entity : IDisposable
     {
 
         //Instance Values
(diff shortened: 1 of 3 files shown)


AlexVild (approved):

alexcon92:
Hey, I've only just found this project so forgive me if I'm talking nonsense but I pulled this last night and the new ShouldGetItems call in the Npc class is throwing a StackOverflow Exception, looks like it's just repeatedly calling itself. I've made a change in my fork that just involves grabbing the logic from the override in the Player class and using that in the Npc class - is this miles off or is this pretty much what was intended with this change? If so happy to create a PR :)

WeylonSantana:
Hey @alexcon92 thanks for your comment, and another user has already warned us about the error
image
It will be fixed soon
I invite you to be part of the discord server if you have one, and if you are a developer you are welcome to contribute to the project in any way you can if you wish.
Here is the link: https://discord.com/channels/697405212712894504/697776750888878152/766650225800249374

PR #1442, opened 2022-09-05
New Trigger Layout

### LINKED ISSUE #1441: feature: New trigger layout
### Description

it's horrible when we put as triggers in common event variable or even triggers with big name (guilds, etc), because we can't see the full name of the variable, the space is small, we need a small readjustment for that.

### Intended Use-Case

Small changes for dev comfort

### Duplicate Check

- [X] This feature request is not a duplicate to the best of my knowledge.

WeylonSantana:
resolves #1441

https://user-images.githubusercontent.com/63019821/188471470-a209469c-762a-4a7f-94de-87967efa6ccf.mp4


lodicolo (changes requested):
@WeylonSantana can you make it so the group in the winform does not resize depending on the trigger?

PR #1444, opened 2022-09-05
Manasteal

### LINKED ISSUE #1443: feature: Manasteal
### Description

Steal player/npc mana

### Intended Use-Case

It can be used in combat in many different ways.

### Duplicate Check

- [X] This feature request is not a duplicate to the best of my knowledge.

WeylonSantana:
resolves #1443

*Added a new (extra buff) equipment property - Manasteal
If the player has any percentage above 0 in any part of the equipment, mana will be stolen from the enemy in (percentage * damage).
**Enemy mana is decreased along with health**

If the percentage to be stolen is greater than the enemy's mana. Only the amount of mana will be regenerated and the rest will deal extra damage to the enemy.
For example: damage 20, 100% manasteal, enemy has 10 mana.
Will restore 10 mana to the attacker and deal +10 damage to the enemy

If the enemy has no mana then it will take everything as extra damage

The damage will be true, as it ignores defense, as it is damage caused directly by the server.

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -1951,23 +1951,23 @@ namespace Intersect.Server.Entities
                     );
                 }
 
-                var manasteal = thisPlayer.GetEquipmentBonusEffect(EffectType.Manasteal) / 100f;
-                var manaRecovered = Math.Min(enemyVitals[(int) Vitals.Mana], manasteal * baseDamage);
+                var manasteal = (thisPlayer.GetEquipmentBonusEffect(EffectType.Manasteal) / 100f) * baseDamage;
+                var manaRecovered = Math.Min(enemyVitals[(int) Vitals.Mana], manasteal);
 
                 if (manaRecovered > 0) //Don't send any +0 msg's.
                 {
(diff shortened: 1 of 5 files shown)


Intersect.Server/Entities/Entity.cs
```diff
@@ -1950,6 +1950,45 @@ public virtual void TryAttack(Entity target)
...
+                            enemy.SubVital(Vitals.Health, (int) amountToRemove);
+                            PacketSender.SendActionMsg(
+                                enemy, Strings.Combat.removesymbol + amountToRemove, CustomColors.Combat.TrueDamage
+                            );
+                        }
+                    }
+                }
+                else
+                {
+                    //If enemy has 0 mana, all value will be dealt as damage
```
lodicolo, line 1982: Even if it only deals damage, it should still recover mana

Also, because this is computed after life steal, you need to Math.Min() against the _current_ health remaining (which might not be what is stored in `enemyVitals` because lifesteal may have been subtracted already), and use that to compute mana recovered and damage dealt.

lodicolo (changes requested):

lodicolo (approved):

Issue #224, opened 2020-05-17
enhancement: client/editor should use unconnected messages to check server status


lodicolo:
**Summary**: Currently the client/editor establish full-blown connections to the server even on the main menu, where the client could be left inactive but still hogging a connection.

**Solution**:
- Client/Editor should use unconnected messages to communicate with server before login
  - Primarily server status, version verification
- Client/Editor should request full connections on login attempt
- Client should:
  - lose the connection on logout
  - request full connections on registration

Arufonsu:
Not sure if related, but after reading "the client establish full-blown connection" i've realized that this may be related with this issue:

(1)
-Start the server.
-Asap and while it's loading, open a client.
-Server crashes.

(2)
-Start the client. - will show up the server as offline.
-Start the server. The client will update the current statys to "online" as normal
_-No issue here_

(3)
-Start the server.
-Wait until it loads properly.
-Start the client, _no issue here_, but i shouldn't be waiting for the server to fully load, players can't tell about that right?

Why i mention these 3 situations and why i think that "the client establish full-blown connection" may be the fault of the (1) issue:
in the normal basis, when online games are offline due to maintenance or updates, players will keep opening and closing their clients in order to refresh and see if their game is up, this may lead to multiple server crashes (1).

About (2): Maybe.. just maybe, the client shouldn't start up right away, unless the server its completly online.
if the server is offline, instead of the server status in the main menu screen, a notice window should pop up saying "the game server is currently offline / under maintenance / or whatever". This way, situation 3 (where we wait in order to prevent the server crash) and situation 1 could be solved, leaving us with no situation (2) anymore at this point, yet, a more stable server that's able to load up without interruptions its always better.

If anyone can help by replicating the situation (1) that i previously mentioned, then we could confirm that the issue that @lodicolo  is talking about may be related with this as well.

And yes, adding to what @lodicolo says here, the client shouldn't be hogging a connection indefinitely, an online status check should be enough upon client start up, if the server is offline, then don't allow the client to start up. Once the client is up, connections should be somewhat limited, if no registration/login/char select request is made within x minutes, the server should drop the connection and close the user's client (i've seen this in many online games by now, they do this in order to prevent unnecessary flood of packets from afkers or users who simply leave the game open in the main/login screen.

It may not seem like a big deal right now, but when it comes to scalability and a steady increase in players, then the problem will be more than real. 

Cheers <3

jcsnider:
[...]

PR #1464, opened 2022-09-24
fix: prevent time crash if user logs in too quickly after server start

lodicolo:
If a user logs in extremely quickly (before the server has begun ticking) the time range value is still -1 which is out of range and **this causes an unexplained black screen with no client errors on login** -- changing it to 0 instead which is in-bounds and shouldn't cause issues (since SendTimeToAll() wouldn't be called if it was changing to zero or there are no players online).

### DIFF
--- a/Intersect.Server/General/Time.cs
+++ b/Intersect.Server/General/Time.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 using Intersect.GameObjects;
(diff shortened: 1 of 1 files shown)


Arufonsu:
ohh nice !

Related: https://github.com/AscensionGameDev/Intersect-Engine/issues/224#issuecomment-781783405

PR #1563, opened 2022-09-25
Update shield info

WeylonSantana:
The shield information is not updated constantly when the user uses the shield spell effect. It is necessary to keep recasting the spell constantly to know how much shield is left.

I don't know if this solution is the best (there must be other ways), but with this the problem is solved, I'm sending pr for better analysis by professionals


lodicolo:
I don't know that this is the best solution but it at least improves the bad state shields are in.

PR #1577, opened 2022-09-28
Fix warp bug

### LINKED ISSUE #1296: bug: Player disappears when teleported via event
**Describe the bug**
One of the players becomes completely invisible to the other when teleporting to another nearby map.

**Version Encountered**
0.7.1.22

**To Reproduce**
1 - Create 1 NPC that will teleport the player to 1 map above.
2 - Have 2 players logged in.
3 - The first player, Use the npc to have him teleport you to the map above.
4 - The second player, Use the npc to make him teleport you to the map above.
5 - See the error.... The first player will be able to see the second player on the new map, but the second player will not be able to see it.

**Expected behavior**
Both players could see each other

**Operating System:**
Windows, Mac, Or Linux

WeylonSantana:
Resolves #1296

### DIFF
--- a/Intersect.Client/Networking/PacketHandler.cs
+++ b/Intersect.Client/Networking/PacketHandler.cs
@@ -1369,7 +1369,7 @@ namespace Intersect.Client.Networking
             {
                 foreach (var projDeath in packet.ProjectileDeaths)
                 {
-                    if (Globals.Entities.ContainsKey(projDeath) && Globals.Entities[projDeath] is Projectile projectile)
+                    if (Globals.Entities.ContainsKey(projDeath) && Globals.Entities[projDeath] is Projectile)
                     {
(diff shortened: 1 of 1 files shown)


lodicolo (approved):

Intersect.Client/Networking/PacketHandler.cs
```diff
@@ -1372,7 +1371,6 @@ public void HandlePacket(IPacketSender packetSender, ProjectileDeadPacket packet
                 {
                     if (Globals.Entities.ContainsKey(projDeath) && Globals.Entities[projDeath] is Projectile projectile)
```
lodicolo, line 1372: you can change `is Projectile projectile` to `is Projectile` -- the variable is no longer used

lodicolo (changes requested):

lodicolo (approved):

PR #1586, opened 2022-10-02
fix: link cb/rb subcomponents to parents

### LINKED ISSUE #1585: bug: unable to disable LabeledCheckBox controls
### Description

`LabeledCheckBox.IsDisabled = true;` and `LabeledCheckBox.Disable();` does not disable the entire control and it's child elements, currently, the only way to "disable" labeled check boxes, is to by disabling their "Label" and "Checkbox" child elements manually thru the json file. It would be nice to be able to their the entire control along with it's child elements to disable with ease thru the code, that way we could disable specific labeled check boxes by code under specific conditions.

### Steps to Reproduce

1. Create a labeled checkbox 
image

2. Try to disable it (in this specific case/example) 
with `mAutoTurnToTarget.IsDisabled = true;` and/or `mAutoTurnToTarget.Disable();`

3. Outcome: Enabled and check-able
image

4. Expected Outcome: Disabled and unable to check.
image

### Version with bug

0.7.1.60

Unknown

### Affected platforms

Windows

The step 4. on Steps to Reproduce can be achieved by doing the following:

Within the control's children at the json structure, disable it's label and checkbox

image

Outcome:
image

### Relevant log output

```shell
none
```

### Duplicate Bug Check

lodicolo:
Resolves #1585

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/Base.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Base.cs
@@ -27,6 +27,19 @@ namespace Intersect.Client.Framework.Gwen.Control
     public partial class Base : IDisposable
     {
 
+        private bool _inheritParentEnablementProperties;
+
+        internal bool InheritParentEnablementProperties
+        {
+            get => _inheritParentEnablementProperties;
+            set
+            {
+                _inheritParentEnablementProperties = value;
+                IsDisabled = Parent?.IsDisabled ?? IsDisabled;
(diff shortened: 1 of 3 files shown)


Arufonsu:
Works great when disabling a control by editing it's parent `json` property manually now 👍🏽 
but i can't disable the control by code with `.IsDisabled = true;` nor `.Disabled();` 🤔

Arufonsu (approved):

PR #1588, opened 2022-10-03
feat: auto-turn to selected target

### LINKED ISSUE #993: A "aim" system
Aiming projectiles is a nightmare. All too often you end up moving around when you don't mean to.

Introduce functionality where:
- A player can enable a "turn-to-target" option that turns them to face whatever target they've selected whenever a new target is determined
- A button on the keyboard that turns the player toward their current target

Arufonsu:
* Implements a new targeting game setting that allows players to auto-turn to the selected target when enabled.

* Player will auto-turn to their currently selected target (by clicking on them or switching targets) only when not moving and if the direction needs to be changed from the current one.

* The feature can be toggled by the server config as well. When disabled, it will make it so the feature no longer works for the client and the checkbox within the game settings is disabled and unchecked.

* Minor chore at DrawTargets() : reduced nesting by merging and inverting if(s) and used conditional en.Value != null within the foreach loops required for this feature.

* Special thanks to Daywalkr (Middle Ages: Online), took (and changed a bit) the implementation from his repository.

* Should resolve #993 
( #1530 )

[Assets pull request](https://github.com/AscensionGameDev/Intersect-Assets/pull/25)

Video Preview:

https://user-images.githubusercontent.com/17498701/193482524-49ae44a0-d41c-40a8-abbc-15d0e77facc1.mp4

image

### DIFF
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -16,6 +16,11 @@
         public bool AllowCombatMovement { get; set; } = true;
 
         /// <summary>
+        /// Enables de client setting 'Auto-turn to target'.
+        /// </summary>
+        public bool EnableAutoTurnToTarget { get; set; } = true;
+
+        /// <summary>
         /// Enables or disables friend login notifications when a user joins the game.
         /// </summary>
         public bool EnableFriendLoginNotifications { get; set; } = true;
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -16,9 +16,14 @@
         public bool AllowCombatMovement { get; set; } = true;
 
         /// <summary>
-        /// Enables de client setting 'Auto-turn to target'.
+        /// Enables the client feature 'Auto-turn to target'.
         /// </summary>
         public bool EnableAutoTurnToTarget { get; set; } = true;
+
+        /// <summary>
+        /// Sets the delay (milliseconds) for the feature 'Auto-turn to target'.
+        /// </summary>
+        public long AutoTurnToTargetDelay { get; set; } = 50;
 
         /// <summary>
         /// Enables or disables friend login notifications when a user joins the game.
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -23,7 +23,7 @@
         /// <summary>
         /// Sets the delay (milliseconds) for the feature 'Auto-turn to target'.
         /// </summary>
-        public long AutoTurnToTargetDelay { get; set; } = 50;
+        public ushort AutoTurnToTargetDelay { get; set; } = 500;
 
         /// <summary>
         /// Enables or disables friend login notifications when a user joins the game.
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1921,35 +1921,16 @@ namespace Intersect.Client.Entities
 
             var yOffset = originY - targetY;
             var xOffset = originX - targetX;
-            var preferY = (Math.Abs(yOffset) - Math.Abs(xOffset)) > 0;
 
-            switch (preferY)
+            if ((Math.Abs(yOffset) - Math.Abs(xOffset)) > 0)
             {
-                case true when yOffset > 0:
-                    newDir = (byte)Directions.Up;
-                    break;
-                case true when yOffset < 0:
-                    newDir = (byte)Directions.Down;
-                    break;
-                case true when xOffset > 0:
-                    newDir = (byte)Directions.Left;
-                    break;
-                case true when xOffset < 0:
-                    newDir = (byte)Directions.Right;
-                    break;
-
-                case false when xOffset > 0:
-                    newDir = (byte)Directions.Left;
-                    break;
-                case false when xOffset < 0:
-                    newDir = (byte)Directions.Right;
-                    break;
-                case false when yOffset > 0:
-                    newDir = (byte)Directions.Up;
-                    break;
-                case false when yOffset < 0:
-                    newDir = (byte)Directions.Down;
-                    break;
+                newDir = (Math.Abs(yOffset) != 0) ? (yOffset > 0) ? (byte)Directions.Up : (byte)Directions.Down :
+                    (xOffset > 0) ? (byte)Directions.Left : (byte)Directions.Right;
+            }
+            else
+            {
+                newDir = (Math.Abs(xOffset) != 0) ? (xOffset > 0) ? (byte)Directions.Left : (byte)Directions.Right :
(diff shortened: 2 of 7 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -2268,44 +2282,35 @@ public void DrawOverheadInfoOnHover()
...
             {
-                foreach (var en in eventMap.LocalEntities)
+                foreach (var en in eventMap.LocalEntities.Where(en => en.Value != null))
                 {
-                    if (en.Value == null)
+                    if (en.Value.MapId != eventMap.Id ||
+                        ((Event)en.Value).DisablePreview ||
+                        en.Value.IsHidden ||
+                        (en.Value.IsStealthed && (!(en.Value is Player player) || !Globals.Me.IsInMyParty(player))) ||
+                        TargetType != 1 || TargetIndex != en.Value.Id)
```
lodicolo, line 2307: Same comments about the if-continue here

Intersect.Client/Interface/Shared/SettingsWindow.cs
```diff
@@ -425,6 +431,11 @@ void TargetingSettings_Clicked(Base sender, ClickedEventArgs arguments)
             mInterfaceSettings.Hide();
             mInformationSettings.Hide();
             mTargetingSettings.Show();
+            if (!Options.EnableAutoTurnToTarget)
+            {
+                mAutoTurnToTarget.IsChecked = false;
```
lodicolo, line 436: Don't really care about this

Intersect.Client/Interface/Shared/SettingsWindow.cs
```diff
@@ -425,6 +431,11 @@ void TargetingSettings_Clicked(Base sender, ClickedEventArgs arguments)
...
+                mAutoTurnToTarget.IsDisabled = true;
```
lodicolo, line 437: Just make it mAutoTurnToTarget.IsDisabled = !Options......EnableAutoTurnToTarget

lodicolo (changes requested):

Intersect.Client/Entities/Entity.cs
```diff
@@ -1921,35 +1921,16 @@ protected byte DirectionToTarget(Entity en)
 
             var yOffset = originY - targetY;
             var xOffset = originX - targetX;
-            var preferY = (Math.Abs(yOffset) - Math.Abs(xOffset)) > 0;
 
-            switch (preferY)
+            if ((Math.Abs(yOffset) - Math.Abs(xOffset)) > 0)
```
lodicolo, line 1925: Isn't this the same as `Math.Abs(yOffset) > Math.Abs(xOffset)`

Intersect.Client/Entities/Player.cs
```diff
@@ -1403,6 +1403,20 @@ private void SetTargetBox(Entity en)
...
+            if (!Globals.Database.AutoTurnToTarget || !Options.Instance.PlayerOpts.EnableAutoTurnToTarget || IsMoving ||
+                Dir == MoveDir)
+            {
+                return;
+            }
+
+            MoveDir = -1;
+            byte directionToTarget = DirectionToTarget(en);
+            Dir = directionToTarget;
+            PacketSender.SendDirection(directionToTarget);
```
lodicolo, line 1417: Just do
```csharp
Dir = DirectionToTarget(en);
PacketSender.SendDirection(Dir);
```

Intersect.Client/Entities/Entity.cs
```diff
@@ -1878,6 +1878,64 @@ protected virtual bool TryGetAnimationTexture(string textureName, SpriteAnimatio
...
+            {
+                newDir = (Math.Abs(yOffset) != 0) ? (yOffset > 0) ? (byte)Directions.Up : (byte)Directions.Down :
+                    (xOffset > 0) ? (byte)Directions.Left : (byte)Directions.Right;
+            }
+            else
+            {
+                newDir = (Math.Abs(xOffset) != 0) ? (xOffset > 0) ? (byte)Directions.Left : (byte)Directions.Right :
+                    (yOffset > 0) ? (byte)Directions.Up : (byte)Directions.Down;
+            }
```
lodicolo, line 1934: So on further thought I think there is value for switches here but not quite the way you had done before.

I would instead replace the entire if statement with

```csharp
var <x/y>Dir = <x/y>Offset == 0 ? newDir : (<x/y>Offset < 0 ? (byte)Direction.<direction> : (byte)Direction.<direction>);
var <x/y>Dir = <x/y>Offset == 0 ? newDir : (<x/y>Offset < 0 ? (byte)Direction.<direction> : (byte)Direction.<direction>);
newDir = Math.abs(yOffset) > Math.abs(xOffset) ? yDir : xDir;
```

lodicolo: This isn't really much more readable, but it is slightly more readable than what you would have to change the code to in order to fix the case where yOffset isn't 0 and xOffset is (currently Right instead of current newDir), and where xOffset isn't 0 but yOffset is (currently it's Down instead of the current newDir) -- which this change does fix.

Intersect.Client/Entities/Player.cs
```diff
@@ -2296,16 +2322,33 @@ public void DrawTargets()
...
+                    if (en.Value.MapId != eventMap.Id)
                     {
-                        if (TargetType == 1 && TargetIndex == en.Value.Id)
-                        {
-                            en.Value.DrawTarget((int)TargetTypes.Selected);
-                        }
+                        continue;
+                    }
+
+                    if (((Event)en.Value).DisablePreview)
```
lodicolo, line 2330: I realize that this is the same as the original code, but it seems like this needs to say (for safety) `if (en.Value is Event eventEntity && eventEntity.DisablePreview)`

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: btw, ended up requiring it as a variable in the last commit

Arufonsu:
Along with panda's review (II) ; added minor changes in order to prevent the following scenario that i didn't saw until today when testing with a non-admin account:

image

Intersect (Core)/Config/PlayerOptions.cs
```diff
@@ -16,10 +16,15 @@ public partial class PlayerOptions
...
         /// <summary>
-        /// Enables de client setting 'Auto-turn to target'.
+        /// Enables the client feature 'Auto-turn to target'.
         /// </summary>
         public bool EnableAutoTurnToTarget { get; set; } = true;
 
+        /// <summary>
+        /// Sets the delay (milliseconds) for the feature 'Auto-turn to target'.
+        /// </summary>
+        public long AutoTurnToTargetDelay { get; set; } = 50;
```
lodicolo, line 26: Honestly this should be at like 500ms not 50. Also can be a `ushort` not a `long`.

lodicolo (changes requested):

lodicolo (approved):

PR #1602, opened 2022-10-10
fix: harden network security

lodicolo:


Arufonsu (approved):
Safety first ⛑️  , approved 👍🏽 
Basically we now demand HTTPS to whatever goes thru the public net,
while we permit HTTPS by default and to whatever goes thru `Loopback = 1, Subnet = 2, Private = 4`

PR #1619, opened 2022-10-15
feat: (Day) Support for multiple bonus effects on an item

### LINKED ISSUE #1620: enhancement: Multiple bonus effects per item
### Description

It's limiting that an item of type equipment can only support a single bonus effect. They should be able to have as many effects as they want.

### Intended Use-Case

Allow a developer to further customize their equip-able items in their games.

### Duplicate Check

AlexVild:
resolves #1620


Intersect.Server/Migrations/Game/20221015154500_MigrateOldBonusEffectToNewListMigration.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
lodicolo, line 1: This migration is not needed

Intersect.Server/Database/GameData/Migrations/BonusEffectsToListMigration.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Intersect.Logging;
```
lodicolo, line 1: This migration is not needed, we can copy the old values via SQL rather than code

Intersect (Core)/GameObjects/ItemBase.cs
```diff
@@ -140,6 +141,7 @@ public AnimationBase EquipmentAnimation
 
         public bool TwoHanded { get; set; }
 
+        [Obsolete("Use Effects (List<EffectData>) instead")]
```
lodicolo, line 144: Just remove this property and regenerate the migration

Intersect.Server/Migrations/Game/20221015154313_MultipleBonusesMigration.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
lodicolo, line 1: Delete this migration and regenerate it.

Afterwards we'll need to modify it to look something like this (note I need the migration generated first to tell you exactly what needs to change in it, and the Down() migration is _not_ complete -- please ping me after you've regenerated the migration and updated the PR with it so I can help with the manual modification to it)

```csharp
using System;
using System.Diagnostics;

using Microsoft.EntityFrameworkCore.Migrations;

namespace Intersect.Server.Migrations.Game
{
    public partial class MultipleBonusesMigration : Migration
    {
        protected override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.AddColumn<string>(
                name: "Effects",
                table: "Items",
                nullable: false,
                defaultValue: string.Empty);

            switch (migrationBuilder.ActiveProvider)
            {
                case "Microsoft.EntityFrameworkCore.Sqlite":
                    _ = migrationBuilder.Sql("UPDATE Items SET Effects = \"[{\"\"EffectType\"\":\" || Effect_Type || \",\"\"EffectPercentage\"\":\" || Effect_Percentage || \"}]\" WHERE Effect_Percentage <> 0;");
                    break;

                case "Pomelo.EntityFrameworkCore.MySql":
                    _ = migrationBuilder.Sql("UPDATE `Items` SET Effects = CONCAT(\"[{\\\"EffectType\\\":\", Effect_Type, \",\\\"EffectPercentage\\\":\", Effect_Percentage, \"}]\") WHERE Effect_Percentage <> 0;");
                    break;
            }

            migrationBuilder.DropColumn(
                name: "Effect_Percentage",
                table: "Items");

            migrationBuilder.DropColumn(
                name: "Effect_Type",
                table: "Items");
        }

        protected override void Down(MigrationBuilder migrationBuilder)
        {
            switch (migrationBuilder.ActiveProvider)
            {
                case "Microsoft.EntityFrameworkCore.Sqlite":
                    throw new NotSupportedException();

                case "Pomelo.EntityFrameworkCore.MySql":
                    _ = migrationBuilder.Sql("UPDATE `Items` SET Effect_Type = JSON_VALUE(Effects, \"$[0].EffectType\"), Effect_Percentage = JSON_VALUE(Effects, \"$[0].EffectPercentage\") WHERE CHAR_LENGTH(`Effects`) > 2;");
                    break;
            }
        }
    }
}
```

lodicolo (changes requested):

PR #1633, opened 2022-10-18
Revert "fix: drop unused db columns"

lodicolo:
Reverts AscensionGameDev/Intersect-Engine#1631

Temporarily reverting so nobody can get screwed by the dataloss introduced in this PR until I merge the fix


lodicolo:
Forgot to merge it, and in the 15 minutes that have passed I already have the fix, no longer necessary.

PR #1644, opened 2022-10-26
chore: replaces StatBuffs with ItemProperties

WeylonSantana:
Replacing StatBuffs for ItemProperties so that way adding new individual item properties will become a less painful task.

This preset resets the random stats of all items that are currently in the players hand/inventory (maybe in chests and etc?), the random stats will be applied to new items generated after migration


Intersect.Server/Database/Item.cs
```diff
@@ -20,44 +21,24 @@ public Item()
...
         }
 
         public Item(Item item) : this(item.ItemId, item.Quantity, item.BagId, item.Bag)
         {
-            for (var i = 0; i < (int) Stats.StatCount; i++)
-            {
-                StatBuffs[i] = item.StatBuffs[i];
-            }
-
+            ItemProperties = item.ItemProperties;
```
lodicolo, line 41: This should be cloning the item properties as well.

Something like `ItemProperties = new ItemProperties(item.ItemProperties);` or `ItemProperties = item.ItemProperties.Clone()`.

Intersect.Server/Database/Item.cs
```diff
@@ -74,6 +55,7 @@ public Item(Item item) : this(item.ItemId, item.Quantity, item.BagId, item.Bag)
 
         public int Quantity { get; set; }
 
+        //DEPRECATED
```
lodicolo, line 58: Actual deprecation should be via `[Obsolete]` (and it would need to be on StatBuffs not StatBuffsJson), but if it's being replaced it should just be removed, even if that is a pain in the ass.

Even if it's only being deprecated though, you aren't copying the old data into the new ItemProperties, and so this change is effectively causing data loss.

Can you please check and see what ItemProperties looks like when serialized as JSON and stored in the DB when it does have ItemProperties set? This will help write the SQL-only migration to ItemProperties.

Intersect.Server/Database/Item.cs
```diff
@@ -20,44 +21,24 @@ public Item()
         {
         }
 
-        public Item(Guid itemId, int quantity, bool incStatBuffs = true) : this(
-            itemId, quantity, null, null, incStatBuffs
+        public Item(Guid itemId, int quantity, ItemProperties ItemProperties = null) : this(
```
lodicolo, line 24: The parameter name should just be `properties`, we already know they're item properties.

lodicolo: This constructor is also _not_ functionally equivalent anymore.

Intersect (Core)/Network/Packets/Server/ItemProperties.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            for (var i = 0; i < (int)Stats.StatCount; i++)
+            {
+                // TODO: What the fuck?
+                RandomStats[i] = Randomization.Next(-descriptor.StatGrowth, descriptor.StatGrowth + 1);
+            }
+        }
+
+        [Key(1)]
+        public int[] RandomStats { get; set; } = new int[(int)Stats.StatCount];
```
lodicolo, line 37: This should be called `StatModifiers` instead of `RandomStats`. Their _generation_ might be random, but that's not the only way to populate them.

Intersect (Core)/Network/Packets/Server/ItemProperties.cs
```diff
@@ -0,0 +1,39 @@
...
+            var descriptor = ItemBase.Get(itemBaseId);
+            if (descriptor == null)
+            {
+                return;
+            }
+
+            if (descriptor.ItemType != ItemTypes.Equipment)
+            {
+                return;
+            }
```
lodicolo, line 27: This section should remain in the server's Item implementation.

Intersect (Core)/Network/Packets/Server/ItemProperties.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            for (var i = 0; i < (int)Stats.StatCount; i++)
+            {
+                // TODO: What the fuck?
+                RandomStats[i] = Randomization.Next(-descriptor.StatGrowth, descriptor.StatGrowth + 1);
+            }
```
lodicolo, line 33: Add a method in `Randomization` with the signature `public static void NextValues(int[] buffer)` and it should fill the buffer with random integers.

In the `Item` implementation on the server it should fill StatModifiers using `Randomization.NextValues(properties.StatModifiers)` according to the equipment logic and the stat randomizer boolean.

Intersect.Server/Database/Item.cs
```diff
@@ -85,6 +67,17 @@ public string StatBuffsJson
         [NotMapped]
         public int[] StatBuffs { get; set; } = new int[(int) Stats.StatCount];
 
+        [NotMapped]
+        public ItemProperties ItemProperties { get; set; }
```
lodicolo, line 71: We're already in the Item class, the property should just be named `Properties`.

Intersect.Server/Database/Item.cs
```diff
@@ -85,6 +67,17 @@ public string StatBuffsJson
...
+
+        [Column("ItemProperties")]
+        [JsonIgnore]
+        public string ItemPropertiesJson
+        {
+            get => JsonConvert.SerializeObject(ItemProperties);
+            set => ItemProperties = JsonConvert.DeserializeObject<ItemProperties>(value ?? "") ?? new ItemProperties();
```
lodicolo, line 78: Use `string.Empty` instead of `""`

Intersect.Server/Entities/Player.cs
```diff
@@ -1550,7 +1550,13 @@ public override int CalculateAttackTime()
                 var descriptor = equippedItem.Descriptor;
                 if (descriptor != null)
                 {
-                    flatStats += descriptor.StatsGiven[(int)statType] + equippedItem.StatBuffs[(int)statType];
+                    flatStats += descriptor.StatsGiven[(int)statType];
+
+                    if(equippedItem.ItemProperties.RandomStats != null)
```
lodicolo, line 1555: nit: Should be a space between `if` and `(`

Intersect.Server/Entities/Entity.cs
```diff
@@ -2729,7 +2729,7 @@ protected virtual bool ShouldDropItem(Entity killer, ItemBase itemDescriptor, It
             }
 
             // Set the attributes for this item.
-            item.Set(new Item(item.ItemId, item.Quantity, true));
+            item.Set(new Item(item.ItemId, item.Quantity));
```
lodicolo, line 2732: This constructor call didn't need to change

Intersect.Server/Database/Item.cs
lodicolo: I would re-create the old constructor with the boolean renamed to `randomizeStatModifiers` and this constructor would be a _new_ constructor and `properties` should _not_ have a default-null value.

Intersect.Server/Database/Item.cs
```diff
@@ -20,44 +21,24 @@ public Item()
...
-            {
-                return;
-            }
-
-            for (var i = 0; i < (int) Stats.StatCount; i++)
-            {
-                // TODO: What the fuck?
-                StatBuffs[i] = Randomization.Next(-descriptor.StatGrowth, descriptor.StatGrowth + 1);
-            }
+            ItemProperties = itemProperties != null ? itemProperties : new ItemProperties(ItemId);
```
lodicolo, line 36: This constructor is also _not_ functionally equivalent anymore. ItemProperties are now defined with random stats 100% of the time and there's no way to turn that off.

Add `properties` between bag and includeStatBuffs, and rename includeStatBuffs to be "randomizeStatModifiers".

Stat randomization logic should remain in this constructor.

Intersect.Server/Database/Item.cs
```diff
@@ -97,10 +90,7 @@ public virtual void Set(Item item)
             Quantity = item.Quantity;
             BagId = item.BagId;
             Bag = item.Bag;
-            for (var i = 0; i < (int) Stats.StatCount; i++)
-            {
-                StatBuffs[i] = item.StatBuffs[i];
-            }
+            ItemProperties = item.ItemProperties;
```
lodicolo, line 93: This should do `new ItemProperties(item.Properties)`

lodicolo (changes requested):
I haven't really reviewed the client changes, the server changes are extensive enough that once they are changed I will re-review everything.

Intersect.Server/Database/Item.cs
WeylonSantana: image

WeylonSantana: And an item that uses the old and the new
image

Intersect (Core)/Network/Packets/Server/ItemProperties.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (other == default)
+            {
+                throw new ArgumentNullException(nameof(other));
+            }
+
+            Array.Copy(other.StatModifiers, StatModifiers, (int)Stats.StatCount);
+        }
+
+        [Key(1)]
```
lodicolo, line 24: Shouldn't this be `[Key(0)]`?

Intersect.Client/Interface/Game/Crafting/RecipeItem.cs
```diff
@@ -121,8 +121,7 @@ void pnl_HoverEnter(Base sender, EventArgs arguments)
             if (mIngredient != null && ItemBase.Get(mIngredient.ItemId) != null)
             {
                 DescWindow = new ItemDescriptionWindow(
-                    ItemBase.Get(mIngredient.ItemId), mIngredient.Quantity, mCraftingWindow.X, mCraftingWindow.Y,
-                    new int[(int) Stats.StatCount]
+                    ItemBase.Get(mIngredient.ItemId), mIngredient.Quantity, mCraftingWindow.X, mCraftingWindow.Y, null
```
lodicolo, line 124: Change to `if (mIngredient != null && ItemBase.TryGet(mIngredient.ItemId, out var itemDescriptor))`

and pass `itemDescriptor` instead of doing a second `Get()` on this line.

Intersect.Client/Interface/Game/DescriptionWindows/ItemDescriptionWindow.cs
```diff
@@ -323,11 +324,11 @@ protected void SetupEquipmentInfo()
             }
 
             // Stats
-            if (mStatBuffs != null)
+            if (mItemProperties != null && mItemProperties.StatModifiers != null)
```
lodicolo, line 327: `mItemProperties?.StatModifiers != default`

Intersect.Client/Interface/Game/DescriptionWindows/ItemDescriptionWindow.cs
```diff
@@ -244,21 +245,21 @@ protected void SetupEquipmentInfo()
...
                             speed = (int) Math.Round(speed / ((100 + weapon.PercentageStatsGiven[(int)Stats.Speed]) / 100f));
                             speed -= weapon.StatsGiven[(int)Stats.Speed];
-                            speed -= statBuffs[(int)Stats.Speed];
+                            speed -= randomStats[(int)Stats.Speed];
                         }
                     }
 
                     // Add current item's speed stats!
-                    if (mStatBuffs != null)
+                    if (mItemProperties != null)
```
lodicolo, line 259: `mItemProperties?.StatModifiers != default`

Intersect.Client/Items/Item.cs
```diff
@@ -37,7 +38,7 @@ public IItem Clone()
 
             for (var i = 0; i < (int)Stats.StatCount; i++)
             {
-                newItem.StatBuffs[i] = StatBuffs[i];
+                newItem.ItemProperties.StatModifiers[i] = ItemProperties.StatModifiers[i];
```
lodicolo, line 41: You never set `newItem.ItemProperties` to something not-null, this will NRE.

lodicolo (changes requested):

lodicolo (approved):

lodicolo (approved):

Arufonsu (approved):

PR #1645, opened 2022-10-28
feat: overhead hp bar settings

### LINKED ISSUE #969: feature: game settings for in-game visibility of HP bars
### **Is your feature request related to a problem? Please describe.**
Currently, HP bars only show up when entities are damaged. This behaviour is fine... But **then**, we have the other side of the spectrum, some users may enjoy the idea of HP bars being shown all the time for all entities, or perhaps, we might want to know entities HP bars by quickly hovering the cursor over them as well.

### **Describe the solution you'd like**
A sub-category for "HP Bars" in game settings with the following ( **concept** ):
imagen

- Whenever an specific option is toggled on, HP bars will always show up in the game view for the selected entity.
- When toggled off (hidden - default), HP Bars will show up only when entities are damaged (original behaviour) 
and/or when hovered with the cursor.

Arufonsu:
* Adds feature: visibility settings for entity overhead hp bars.
* When toggled off (default) : 
original behaviour (damaged entity) will show up hp bars on entities ;
along with the ability to see their hp bars by hovering the cursor on them.
* When toggled on, the selected type of entity will always show up their hp bars in the game view.
* Adds player-exclusive override for the boolean ShouldDrawHpBar.
* Fixes a crash due to null player being parsed over to the player boolean IsInMyParty.
* Fixed a couple of localized string and comment typos.
* Should resolve #969

[**Assets for this pull request**](https://github.com/AscensionGameDev/Intersect-Assets/pull/28)

### Preview:

image

image image image

image image

### DIFF
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1413,7 +1413,7 @@ namespace Intersect.Client.Entities
                     return true;
                 }
 
-                return GetType() == typeof(Entity) && (Globals.Database.NpcOverheadHpBar);
+                return GetType() == typeof(Entity) && Globals.Database.NpcOverheadHpBar;
             }
         }
 
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1404,7 +1404,7 @@ namespace Intersect.Client.Entities
             return shieldSize;
         }
 
-        public virtual bool ShouldDrawHpBar
+        protected virtual bool ShouldDrawHpBar
         {
             get
             {
@@ -1417,7 +1417,7 @@ namespace Intersect.Client.Entities
             }
         }
 
-        public bool ShouldDrawEntityHpBar
+        protected bool ShouldDrawEntityHpBar
         {
             get
             {
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1408,16 +1408,16 @@ namespace Intersect.Client.Entities
         {
             get
             {
-                if (ShouldNotDrawHpBar)
+                if (ShouldDrawEntityHpBar)
                 {
                     return true;
                 }
 
-                return (GetType() == typeof(Entity) && (Globals.Database.NpcOverheadHpBar || IsHovered));
+                return (GetType() == typeof(Entity) && (Globals.Database.NpcOverheadHpBar));
             }
         }
 
-        public bool ShouldNotDrawHpBar
+        public bool ShouldDrawEntityHpBar
         {
             get
             {
@@ -1430,7 +1430,12 @@ namespace Intersect.Client.Entities
                     maxHealth = MaxVital[(int)Vitals.Health],
                     shieldSize = GetShieldSize();
 
-                return shieldSize > 0 || health != maxHealth;
+                if (health < 1)
+                {
+                    return false;
+                }
+
+                return IsHovered || health != maxHealth || shieldSize > 0;
             }
         }
 
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -214,6 +214,8 @@ namespace Intersect.Client.Entities
         public int WalkFrame { get; set; }
 
         public FloatRect WorldPos { get; set; } = new FloatRect();
+        
+        public bool IsHovered { get; set; }
 
         //Location Info
         public byte X { get; set; }
@@ -1402,7 +1404,7 @@ namespace Intersect.Client.Entities
             return shieldSize;
         }
 
-        public bool ShouldDrawHpBar
+        public virtual bool ShouldDrawHpBar
         {
             get
             {
@@ -1416,8 +1418,6 @@ namespace Intersect.Client.Entities
                     return false;
                 }
 
-                //return true;
-
                 var health = Vital[(int)Vitals.Health];
                 if (health < 1)
                 {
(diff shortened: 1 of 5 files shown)


Intersect.Client/Entities/Entity.cs
```diff
@@ -1426,7 +1426,10 @@ public bool ShouldDrawHpBar
 
                 var maxHealth = MaxVital[(int)Vitals.Health];
                 var shieldSize = GetShieldSize();
-                return shieldSize > 0 || health != maxHealth;
+
+                return (shieldSize > 0 || health != maxHealth) ||
+                       (!(this is Player || this is Projectile || this is Resource || this is Event) &&
+                        (Globals.Database.NpcOverheadHpBar || IsHovered));
```
lodicolo, line 1432: checking for NPC is just this: `GetType() == typeof(Entity)`

Intersect.Client/Entities/Player.cs
```diff
@@ -2372,6 +2376,48 @@ public void DrawOverheadInfoOnHover()
...
+                if (!ShouldDraw)
+                {
+                    return false;
+                }
+
+                var health = Vital[(int)Vitals.Health];
+                if (health < 1)
+                {
+                    return false;
+                }
```
lodicolo, line 2397: Since this is duplicated, create a `protected virtual bool ShouldNotDrawHpBar` that does these computations, and then in both `ShouldDrawHpBar` check if ShouldNotDrawHpBar return false

Intersect.Client/Entities/Player.cs
```diff
@@ -2372,6 +2376,48 @@ public void DrawOverheadInfoOnHover()
...
+
+                var player = this;
+                bool isMe = player.Id == Globals.Me.Id,
+                    isFriend = !isMe && Globals.Me.IsFriend(player),
+                    isPartyMate = !isMe && Globals.Me.IsInMyParty(player),
+                    isGuildMate = !isMe && Globals.Me.IsGuildMate(player),
+                    friendSettingMatch = isFriend && Globals.Database.FriendOverheadHpBar,
+                    guildSettingMatch = isGuildMate && Globals.Database.GuildMemberOverheadHpBar;
+
+                return (shieldSize > 0 || health != maxHealth) ||
```
lodicolo, line 2410: Since this part is duplicated, after the short-circuit with `ShouldNotDrawHpBar`, do a if base.ShouldDrawHpBar return true

Intersect.Client/Entities/Player.cs
```diff
@@ -2372,6 +2376,48 @@ public void DrawOverheadInfoOnHover()
...
+                       (isMe && Globals.Database.MyOverheadHpBar) ||
+                       (!isMe && !isFriend && !isGuildMate && !isPartyMate && Globals.Database.PlayerOverheadHpBar) ||
+                       (isPartyMate && Globals.Database.PartyMemberOverheadHpBar) ||
+                       (friendSettingMatch && guildSettingMatch && !isPartyMate) ||
+                       (friendSettingMatch && !guildSettingMatch && !isPartyMate) ||
+                       (guildSettingMatch && !friendSettingMatch && !isPartyMate) ||
+                       IsHovered;
```
lodicolo, line 2417: The rest of this is wildly inefficient (it makes _every_ expensive check possible) and very illegible (it's difficult to separate anything).

Just break each of the other options into separate if statements, prioritizing the cheapest calls.

```csharp

if (IsHovered) {
    return true;
}

if (player == Globals.Me)
{
    return Globals.Database.MyOverheadHpBar;
}

if (<show party overhead> && <is in party>)
{
    return true;
}

if (<show friend overhead> && <is friend>)
{
    return true;
}

if (<show guild overhead> && <is in guild>)
{
    return true;
}

return general player overhead;
```

lodicolo (changes requested):

Intersect.Client/Entities/Entity.cs
```diff
@@ -1404,32 +1404,38 @@ public int GetShieldSize()
...
+        protected virtual bool ShouldDrawHpBar
         {
             get
             {
-                if (LatestMap == default)
+                if (ShouldDrawEntityHpBar)
                 {
-                    return false;
+                    return true;
                 }
```
lodicolo, line 1414: This is not what I said to add and is the exact opposite case.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
```diff
@@ -2372,6 +2376,39 @@ public void DrawOverheadInfoOnHover()
...
+        protected override bool ShouldDrawHpBar
+        {
+            get
+            {
+                if (!ShouldNotDrawHpBar || IsHovered)
+                {
+                    return true;
+                }
+
+                if (this.Id == Globals.Me.Id)
```
lodicolo, line 2388: `this.Id` -> `Id`

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: Hmm but, don't we have to check against "my ID" in order to tell filter "my character" from other players (?)

`MyOverheadHpBar =/= PlayerOverheadHpBar`

lodicolo (approved):

Intersect.Client/Entities/Player.cs
```diff
@@ -2372,6 +2376,39 @@ public void DrawOverheadInfoOnHover()
...
         }
 
+        protected override bool ShouldDrawHpBar
+        {
+            get
+            {
+                if (!ShouldNotDrawHpBar || IsHovered)
+                {
+                    return true;
+                }
```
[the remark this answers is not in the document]
Arufonsu: tested this out again today, and returning false within this method always leads to unwanted issues.
right now this disables the ability to show player's hp bars on entities with full HP

i'll open a small PR that reverts this back to this, which is giving as outcome the desired behavior:

```
                if (!ShouldNotDrawHpBar)
                {
                    return true;
                }
```

PR #1646, opened 2022-10-29
chore: configurable listbox even rows

Arufonsu:
Adds a client config to set whether the highlighted even rows of list elements should be marked differently or not.

### When false (default):

image
image
image

### When true (previously hard-coded to be the default behavior):

image
image
image

### DIFF
--- a/Intersect (Core)/Configuration/ClientConfiguration.cs
+++ b/Intersect (Core)/Configuration/ClientConfiguration.cs
@@ -165,7 +165,7 @@ namespace Intersect.Configuration
         /// <summary>
         /// Configures whether the highlighted even rows of list elements should be marked differently or not.
         /// </summary>
-        public bool MarkListEvenRows { get; set; } = true;
+        public bool EnableZebraStripedRows { get; set; } = true;
 
         /// <summary>
         /// Configures the name of the skin or skin texture (must end in .png) to use.
--- a/Intersect (Core)/Configuration/ClientConfiguration.cs
+++ b/Intersect (Core)/Configuration/ClientConfiguration.cs
@@ -165,7 +165,7 @@ namespace Intersect.Configuration
         /// <summary>
         /// Configures whether the highlighted even rows of list elements should be marked differently or not.
         /// </summary>
-        public bool MarkListEvenRows { get; set; } = false;
+        public bool MarkListEvenRows { get; set; } = true;
 
         /// <summary>
         /// Configures the name of the skin or skin texture (must end in .png) to use.
--- a/Intersect (Core)/Configuration/ClientConfiguration.cs
+++ b/Intersect (Core)/Configuration/ClientConfiguration.cs
@@ -163,6 +163,11 @@ namespace Intersect.Configuration
         public bool EnableContextMenus { get; set; } = true;
 
         /// <summary>
+        /// Configures whether the highlighted even rows of list elements should be marked differently or not.
+        /// </summary>
+        public bool MarkListEvenRows { get; set; } = false;
+
+        /// <summary>
         /// Configures the name of the skin or skin texture (must end in .png) to use.
         /// </summary>
         public string UiSkin { get; set; } = "Intersect2021";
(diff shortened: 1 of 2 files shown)


Arufonsu:
Notice: it's been discussed among collaborators, that despite this being configurable, _there's still a bug that needs to addressed._

lodicolo:
I am converting this to a draft PR because while I'm not opposed to introducing this as a feature, I am not approving or merging a feature that is just bypassing and not fixing the bug.

The bug needs to be fixed in this PR, or in a separate PR before this PR is merged in, though it is probably easier to just do it in this PR.

Arufonsu:
> I am converting this to a draft PR because while I'm not opposed to introducing this as a feature, I am not approving or merging a feature that is just bypassing and not fixing the bug.
> 
> The bug needs to be fixed in this PR, or in a separate PR before this PR is merged in, though it is probably easier to just do it in this PR.

Well, this is why it was meant to be a chore PR _instead_ of a bug fix PR for now, i don't see why this couldn't be merged in but i guess it's okay to wait for someone to fix the background issue with even rows on lists before adding the ability to disable/enable them.
Also I do agree with the idea of fixing the issue by opening it's own issue report with a PR that fixes it separately from this chore PR.

Arufonsu:
#1695 should resolve the issues so, re-opened this chore PR which is simply meant to make this configurable.

Intersect (Core)/Configuration/ClientConfiguration.cs
```diff
@@ -162,6 +162,11 @@ public void Validate()
         /// </summary>
         public bool EnableContextMenus { get; set; } = true;
 
+        /// <summary>
+        /// Configures whether the highlighted even rows of list elements should be marked differently or not.
+        /// </summary>
+        public bool MarkListEvenRows { get; set; } = false;
```
lodicolo, line 168: This accessibility feature should be default _on_.

Intersect.Client.Framework/Gwen/Control/Layout/Table.cs
```diff
@@ -393,8 +393,16 @@ protected override void Layout(Skin.Base skin)
...
+                if (ClientConfiguration.Instance.MarkListEvenRows)
+                {
+                    row.EvenRow = even;
+                    even = !even;
+                }
+                else
+                {
+                    row.EvenRow = false;
+                }
+
```
lodicolo, line 405: Change this to 

```csharp

row.EvenRow = even;
even = ClientConfiguration.Instance.MarkListEvenRows && !even;

```

lodicolo (changes requested):
If this is why there is no difference between the rows in the preview for #1695 the preview should be redone with the accessibility feature on.

Intersect (Core)/Configuration/ClientConfiguration.cs
Arufonsu: this feature is not included in that PR's preview.
[i've just explained here why they are _not that different - ( but they are _different_ ! )_ ](https://github.com/AscensionGameDev/Intersect-Engine/pull/1695#issuecomment-1369255731)

Arufonsu: that being said, i will change it to enabled by default within this PR 👍🏽

Intersect (Core)/Configuration/ClientConfiguration.cs
```diff
@@ -162,6 +162,11 @@ public void Validate()
         /// </summary>
         public bool EnableContextMenus { get; set; } = true;
 
+        /// <summary>
+        /// Configures whether the highlighted even rows of list elements should be marked differently or not.
+        /// </summary>
+        public bool MarkListEvenRows { get; set; } = true;
```
lodicolo, line 168: It took me a while to remember what this is actually called, but change the name to `EnableZebraStripedRows` (the name of the actual feature, but also made slightly more generic for use with tables).

lodicolo (changes requested):

lodicolo (approved):

PR #1650, opened 2022-11-01
enhancement: bind description windows to cursor position

### LINKED ISSUE #1647: enhancement: ability to bind description windows to cursor position
### Description

Description windows are currently bound to the top-left corner of their parent window, which makes them seem pretty static on their behavior around inventory, loot window, trade window, banks, bags, hot bar (currently centered under them), spells window, crafting window, etc... you probably get the idea. 

image

I don't _personally_ like the default behavior, thus, i would like to have an interface toggle-able checkbox that allows it to dynamically bind description windows to the cursor position, in a way that description windows just overlap their parent windows whenever we hover the cursor over items/spells.

image

### Intended Use-Case

This would allow us to have a way more comfortable view of items and spells in many situations where currently, the legacy behavior isn't fully capable to provide us with right now.

### Duplicate Check

Arufonsu:
* adds a client QoL interface setting (enabled by default) that allows to dynamically bind item/spell description windows to the current cursor position.
* this provides a more flexible and comfortable view in a way that description windows overlap their parent windows whenever we hover the cursor over items/spells.
* this also adjusts the description window without issues on the screen edges, while still based on the current cursor position.
* the legacy behavior is maintained untouched, which can be used back by disabling the enhancement checkbox.
* should resolve #1647 

[**Assets for this pull request**](https://github.com/AscensionGameDev/Intersect-Assets/pull/30)

### Preview

https://user-images.githubusercontent.com/17498701/199294300-db62adff-8a64-4015-9ebc-549616ad3963.mp4

### DIFF
--- a/Intersect.Client/Interface/Game/DescriptionWindows/DescriptionWindowBase.cs
+++ b/Intersect.Client/Interface/Game/DescriptionWindows/DescriptionWindowBase.cs
@@ -1,6 +1,6 @@
 using System.Collections.Generic;
-
 using Intersect.Client.Framework.Gwen.Control;
+using Intersect.Client.General;
 using Intersect.Client.Interface.Game.DescriptionWindows.Components;
 
 namespace Intersect.Client.Interface.Game.DescriptionWindows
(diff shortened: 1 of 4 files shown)


Intersect.Client/Interface/Game/DescriptionWindows/DescriptionWindowBase.cs
```diff
@@ -137,32 +137,64 @@ protected void FinalizeWindow()
...
+                // Note: we no longer need to check after   (newX < 0)   nor    (newY < 0)   in this case.
+                if (newX > mContainer.Canvas.Width - mContainer.Width)
+                {
+                    newX = (int)Globals.InputManager.GetMousePosition().X - mContainer.Width - posAdjust;
+                }
+
+                if (newY > mContainer.Canvas.Height - mContainer.Height)
+                {
+                    newY = (int)Globals.InputManager.GetMousePosition().Y - mContainer.Height - posAdjust;
+                }
```
lodicolo, line 197: I'd like a video/screenshots showing the behavior when it's not locked to the cursor, and ordinarily would be forced off the edge of the screen.

The current video only shows the window-locked variant when it's in the middle of the screen and doesn't show this edge case.

I am also not entirely sure that it should be bound to the cursor, it might make more sense to be bound to the thing the cursor is hovering.

I feel like having a better understanding of the full range of use cases will give us a better solution that involves less code (in particular eliminating the branch -- this code isn't an exact duplicate right now but it feels very duplicate-y, which seems to me like there may be a design flaw in particular with the edge cases).

lodicolo (changes requested):

Intersect.Client/Interface/Game/DescriptionWindows/DescriptionWindowBase.cs
Arufonsu: > The current video only shows the window-locked variant when it's in the middle of the screen and doesn't show this edge case.

Yeah, the default behaviour is kinda glitchy around the left corners sadly, i didn't meant to touch this in order to investigate a proper fix for that in a separate issue report with it's own PR later. As for the proposed windows bound to cursor enhancement, it doesn't glitches up in any corner / side of the game screen.


> I am also not entirely sure that it should be bound to the cursor, it might make more sense to be bound to the thing the cursor is hovering.

Yup, that would be way better to be fair, i just don't really know how to parse over the item/spell 's image panel positions into this, but if we were able to do so, it would be way better than binding these to the cursor.


> will give us a better solution that involves less code (in particular eliminating the branch -- this code isn't an exact duplicate right now but it feels very duplicate-y, which seems to me like there may be a design flaw in particular with the edge cases)

Yeah, it's kinda duplicate-y but we are constantly changing the same two variables in both of these cases (newX, newY), i also meant to keep it this way in order to avoid touching the previous behaviour that's been coded around here, which is, like we previously mentioned, glitchy in some scenarios.

Intersect.Client/Localization/Strings.cs
```diff
@@ -1704,6 +1704,9 @@ public partial struct Settings
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public static LocalizedString AutoCloseWindows = @"Auto-close Windows";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public static LocalizedString BindDescriptionWindowToCursor = @"Bind description windows to cursor";
```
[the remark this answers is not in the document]
Arufonsu: Sounds like we might end up with a combo box to pick the preferred "description tooltips behaviour" 🤔

Intersect.Client/Interface/Game/DescriptionWindows/DescriptionWindowBase.cs
Arufonsu: > I am also not entirely sure that it should be bound to the cursor, it might make more sense to be bound to the thing the cursor is hovering.

i honestly need to find out where to get the X and Y pos from them and this PR could be retaken :monocle_face:

Arufonsu:
> I am also not entirely sure that it should be bound to the cursor, it might make more sense to be bound to the thing the cursor is hovering.

i honestly need to find out where to get the X and Y pos from them and this PR could be retaken :monocle_face:

Arufonsu:
Self reminder -> Intersect.Client/Interface/Game/Inventory/InventoryItem.cs L:190

imagen

Arufonsu:
> > I am also not entirely sure that it should be bound to the cursor, it might make more sense to be bound to the thing the cursor is hovering.
> 
> i honestly need to find out where to get the X and Y pos from them and this PR could be retaken 🧐

Finally found out a viable way to do this while working with https://github.com/AscensionGameDev/Intersect-Engine/pull/2277
```csharp
Container.LocalPosToCanvas(new Point(0, 0)).X;
Container.LocalPosToCanvas(new Point(0, 0)).Y;
```
Will update and reopen this PR with panda's review soon along with this

PR #1653, opened 2022-11-02
fix: close context menu when closing window

### LINKED ISSUE #1651: bug: Context menu does not close after window closes
### Description

title

### Steps to Reproduce

1 - open the inventory
2 - open context menu (secondary on some item)
3 - press ESC
4 - the window closes but the menu does not

ps: if you click on the function (like drop) it will still work even with closed inventory

### Version with bug

0.7.1.81

Unknown

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
* Fixes an unwanted behavior (bug) where context menus would remain opened upon closing the window from where they were initially prompted (by closing them along with their parent window).
* Should resolve #1651

### DIFF
--- a/Intersect.Client/Interface/Game/Bank/BankWindow.cs
+++ b/Intersect.Client/Interface/Game/Bank/BankWindow.cs
@@ -125,8 +125,10 @@ namespace Intersect.Client.Interface.Game.Bank
 
         public void Update()
         {
-            if (mBankWindow.IsHidden == true)
+            if (mBankWindow.IsHidden)
             {
+                mContextMenu.Close();
+
                 if (mOpen)
                 {
                     Interface.GameUi.NotifyCloseBank();
--- a/Intersect.Client/Interface/Game/Bank/BankWindow.cs
+++ b/Intersect.Client/Interface/Game/Bank/BankWindow.cs
@@ -126,7 +126,7 @@ namespace Intersect.Client.Interface.Game.Bank
 
         public void Update()
         {
-            if (mBankWindow.IsHidden == true)
+            if (mBankWindow.IsHidden)
             {
                 if (mOpen)
                 {
--- a/Intersect.Client/Interface/Game/Bank/BankWindow.cs
+++ b/Intersect.Client/Interface/Game/Bank/BankWindow.cs
@@ -95,6 +95,7 @@ namespace Intersect.Client.Interface.Game.Bank
         {
             mBankWindow.IsHidden = true;
             mOpen = false;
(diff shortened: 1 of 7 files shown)


Intersect.Client/Interface/Game/Bank/BankWindow.cs
```diff
@@ -125,8 +125,10 @@ public bool IsVisible()
 
         public void Update()
         {
-            if (mBankWindow.IsHidden == true)
+            if (mBankWindow.IsHidden)
```
lodicolo, line 128: This is fine

Intersect.Client/Interface/Game/Bank/BankWindow.cs
```diff
@@ -125,8 +125,10 @@ public bool IsVisible()
...
             {
+                mContextMenu.Close();
```
lodicolo, line 130: This should be in the Close() method on L94

Intersect.Client/Interface/Game/Inventory/InventoryWindow.cs
```diff
@@ -283,15 +283,15 @@ public bool IsVisible()
             return !mInventoryWindow.IsHidden;
         }
 
-        public bool Hide()
+        public void Hide()
         {
             if (!Globals.CanCloseInventory)
             {
-                return false;
```
lodicolo, line 290: Nothing uses the return value?

lodicolo (changes requested):

Intersect.Client/Interface/Game/Inventory/InventoryWindow.cs
Arufonsu: > Nothing uses the return value?

imagen


Hide() was acting as a boolean but nothing used to require / currently requires it's previously returned values, so i turned the method into a void one since it doesn't really requires to return a value where it's used

Intersect.Client/Interface/Game/Bank/BankWindow.cs
```diff
@@ -125,10 +126,8 @@ public bool IsVisible()
 
         public void Update()
         {
-            if (mBankWindow.IsHidden)
+            if (mBankWindow.IsHidden == true)
```
lodicolo, line 129: The `== true` should still be removed from here because it is redundant.

lodicolo (changes requested):

lodicolo (approved):

PR #1654, opened 2022-11-03
workaround: fog taking too long to render on map

### LINKED ISSUE #1652: bug: fog takes too long to render
### Description

About the fog bug that happens when you come to a new map and the fog waits 2min to appear in the screen, it should be 1sec or less, i mean, this feature from the engine makes total difference to some projects, makes part of the beauty of the game. The time for the fog appear seems to change from computer to computer eng. In previous versions of intersect this do not happened here, have started on this one.

### Steps to Reproduce

1. come to a new map
2. wait for the fog

### Version with bug

0.7.1.63

0.7.1.24

### Affected platforms

Windows

No

### Relevant log output

```shell
Don't have
```

### Duplicate Bug Check

Arufonsu:
* Should resolve #1652 
* Also adjusted the UpdateTimer for overlay images and panorama images.

### Preview

https://user-images.githubusercontent.com/17498701/199637035-850de0cd-d586-4135-9fdd-6a33b56fb949.mp4

### DIFF
--- a/Intersect.Client/Maps/MapInstance.cs
+++ b/Intersect.Client/Maps/MapInstance.cs
@@ -93,7 +93,7 @@ namespace Intersect.Client.Maps
         protected float mFogCurrentY;
 
         //Fog Variables
-        protected long mFogUpdateTime = -1;
+        protected long mFogUpdateTime = Timing.Global.Milliseconds;
 
         //Update Timer
         private long mLastUpdateTime;
@@ -101,12 +101,12 @@ namespace Intersect.Client.Maps
         protected float mOverlayIntensity;
 
         //Overlay Image Variables
-        protected long mOverlayUpdateTime = -1;
+        protected long mOverlayUpdateTime = Timing.Global.Milliseconds;
 
         protected float mPanoramaIntensity;
 
         //Panorama Variables
-        protected long mPanoramaUpdateTime = -1;
+        protected long mPanoramaUpdateTime = Timing.Global.Milliseconds;
 
         private bool mTexturesFound = false;
 


WeylonSantana (approved):

lodicolo (changes requested):
I would like to see a "before" video from your machine as well without the changes.

I would also like to understand why this change is even necessary, because it seems like it's going to be a bandaid instead of solving the real issue.

Arufonsu:
> I would like to see a "before" video from your machine as well without the changes.
> 
> I would also like to understand why this change is even necessary, because it seems like it's going to be a bandaid instead of solving the real issue.

I literally followed the same logic that the editor uses for initializing the fog timers on this one, not really sure how the timing and millisecond have been changed to the point that things like these are now delayed (or were? can't reproduce this issue today anymore for some odd reason... uh, but i do recall the fog issue being a thing that pop ups every now and then).

So yeah, take it as a _workaround atm_ since i'm not entirely sure how/when/why this delay is happening at times, and this is by far what _seems to be helping those with this issue_ as two of our collaborators already confirmed few days ago.

Arufonsu:
_Note from today's collaborator chat:_
It seems like the delay becomes more evident **the longer the server stays on**

Arufonsu:
> I would like to see a "before" video from your machine as well without the changes.
> 
> I would also like to understand why this change is even necessary, because it seems like it's going to be a bandaid instead of solving the real issue.

I've followed the same logic that the editor uses for initial values of the fog timer, not really sure how the timing and millisecond have been changed to the point that things like these are now delayed.

Here are the requested videos.

### **Before** changing the client initial fog timer:
_[By leaving the server on for some extended period of time (12 hrs-ish uptime) ]_


https://user-images.githubusercontent.com/17498701/201734643-eea23e00-dbf6-4e86-9e0e-9c8080e6f309.mp4



### **After** changing the client initial fog timer by following the logic used within the editor:
_[Same server instance as the one used in the first video (12 hrs-ish uptime) ; re-compiled the game client with the changes]_



https://user-images.githubusercontent.com/17498701/201734675-1ab4947a-8df0-49a6-9dfc-c23540ee6433.mp4

lodicolo (changes requested):
I don't know off the top of my head what the editor is doing, but this change is 100% a hack, and if this is what the editor is doing that most certainly does not make it the right fix (if anything, it more likely means that it is definitely the wrong fix).

The logic needs to be fixed, not the initial state value.

Arufonsu:
> I don't know off the top of my head what the editor is doing, but this change is 100% a hack, and if this is what the editor is doing that most certainly does not make it the right fix (if anything, it more likely means that it is definitely the wrong fix).
> 
> The logic needs to be fixed, not the initial state value.

My bad if this doesn't really fix the fog logic in the best possible way, anyways, here's what i meant with the logic used by editor:

https://github.com/AscensionGameDev/Intersect-Engine/blob/d416695aefe79d7de7b6d88e6dcdf02dbb8916e5/Intersect.Editor/Core/Graphics.cs#L72

And here's the method used by the editor to draw the fog, which is literally the same method used by the game client:

https://github.com/AscensionGameDev/Intersect-Engine/blob/d416695aefe79d7de7b6d88e6dcdf02dbb8916e5/Intersect.Editor/Core/Graphics.cs#L1812-L1868

Arufonsu:
closed this and opened the following instead: #1694

PR #1667, opened 2022-11-20
enhancement: show CharacterWindow even with 1 max character

### LINKED ISSUE #1666: enhancement: CharacterWindow should show up even with 1 max character
### Description

To give those games that have (or will have) only one maximum character the ability to delete and create characters, and not log out and log in mandatory.

### Intended Use-Case

In addition to the case mentioned above, it helps in the debugging of those who make custom modifications without necessarily having to increase the number of characters to be able to delete and create characters.

### Duplicate Check

WeylonSantana:
Resolves #1666

### DIFF
--- a/Intersect.Server/Networking/PacketHandler.cs
+++ b/Intersect.Server/Networking/PacketHandler.cs
@@ -605,16 +605,9 @@ namespace Intersect.Server.Networking
             }
 
             //Otherwise proceed with login normally...
-            if (Options.MaxCharacters > 1)
+            if (client.Characters?.Count > 0)
             {
                 PacketSender.SendPlayerCharacters(client);
-            }
-            else if (client.Characters?.Count > 0)
-            {
-                client.LoadCharacter(client.Characters.First());
-                client.Entity.SetOnline();
-
-                PacketSender.SendJoinGame(client);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


Arufonsu (changes requested):
I don't personally like the idea of removing an existing feature (which i personally like, need and use in my project that's based on the intersect engine) as an enhancement.
Instead (and in my opinion): the ability to display/skip the character selection window upon loggin while there's one character within the game account should be a configurable server option (boolean).

lodicolo:
I agree with this being a configurable option rather than just changing the functionality completely.

That said, I think that we may want to change the default behavior (so by default do not skip) because I feel like we've gotten more questions about us only supporting 1 character _because_ it skips.

WeylonSantana:
> I agree with this being a configurable option rather than just changing the functionality completely.
> 
> That said, I think that we may want to change the default behavior (so by default do not skip) because I feel like we've gotten more questions about us only supporting 1 character _because_ it skips.

OK, but it's done that way in #1668, so we can close this one and revise that, so it defaults to always show character window.

lodicolo:
> > I agree with this being a configurable option rather than just changing the functionality completely.
> > That said, I think that we may want to change the default behavior (so by default do not skip) because I feel like we've gotten more questions about us only supporting 1 character _because_ it skips.
> 
> OK, but it's done that way in #1668, so we can close this one and revise that, so it defaults to always show character window. What your think?

Agreed, I hadn't actually seen that there was another PR addressing the same thing until after writing my comment. Closing.

PR #1668, opened 2022-11-20
enhancement: configurable character select skip

### LINKED ISSUE #1666: enhancement: CharacterWindow should show up even with 1 max character

Arufonsu:
* A configurable alternative to resolve #1666 
* Adds the ability to configure whether or not the game client skips the character select window upon login or going back to characters when the max number of characters allowed per account is one.
* QoL change: By logic, newly created accounts have 0 characters on them, so the code logic behind this has been changed to skip the character select window upon creating a new account by directly going thru the character creation window instead.
* QoL change: Already created accounts with 0 characters will directly go thru the character creation window now.

### DIFF
--- a/Intersect (Core)/Config/PlayerOptions.cs
+++ b/Intersect (Core)/Config/PlayerOptions.cs
@@ -91,6 +91,12 @@
         public bool ShowLevelByName { get; set; } = false;
 
         /// <summary>
+        /// Configures whether or not the game client skips the character select window upon login or going back
+        /// to characters when the max number of characters allowed per account is one.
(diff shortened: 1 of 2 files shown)


lodicolo (changes requested):
I apparently forgot to submit this review.

PR #1672, opened 2022-11-25
Removed useless instance

lucas-tardivo:
Removed this unnecessary instance at client player Load() method.

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -310,17 +310,15 @@ namespace Intersect.Client.Entities
             Guild = pkt.Guild;
             Rank = pkt.GuildRank;
 
-            var playerPacket = (PlayerEntityPacket)packet;
-
-            if (playerPacket.Equipment != null)
+            if (pkt.Equipment != null)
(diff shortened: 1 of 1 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -310,17 +310,15 @@ public override void Load(EntityPacket packet)
             Guild = pkt.Guild;
             Rank = pkt.GuildRank;
 
-            var playerPacket = (PlayerEntityPacket)packet;
-
-            if (playerPacket.Equipment != null)
+            if (pkt.Equipment != null)
```
lodicolo, line 313: Rather than using the pkt rename, change `pkt` (which is a bad variable name) to `playerPacket` -- this will change L306-L311 and leave L313+ untouched.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
lucas-tardivo: Changed :)

lodicolo (approved):

PR #1694, opened 2022-12-30
fix: fog timing, intensity and consistency

### LINKED ISSUE #1652: bug: fog takes too long to render

Arufonsu:
* Should resolve #1652 
* Chore: refactors client's `DrawFog` method in order to improve: performance, readability and stability.
* Fix: Solves an issue where the value of `elapsedTime` could potentially be very large if there is a significant delay between calls to the `DrawFog` method. This could cause the `mCurFogIntensity` to increase or decrease at an unexpectedly fast rate, which could cause the fog to appear and disappear abruptly or delayed.
* Removed the unnecessary check for `mCurFogIntensity` being less than 1 in the first conditional block.
* Used Math.Min and Math.Max instead of the if statements to clamp the values of `mCurFogIntensity`
* In order to handle cases where the fog texture's position goes out of bounds: converted `mFogCurrentX`/`mFogCurrentY` to compound assignments.

(30-12-22)
* refactor: client's `CompareEffects`
* refactor: editor's `DrawFog` method in order to improve: performance, readability and stability.
* fixed [inconsistent / mismatching / out of sync] fog texture's position based on its speed and elapsed time on both ends: client and editor.

### Preview (Before this PR)

https://user-images.githubusercontent.com/17498701/201734643-eea23e00-dbf6-4e86-9e0e-9c8080e6f309.mp4

### Preview (After this PR)

https://user-images.githubusercontent.com/17498701/201734675-1ab4947a-8df0-49a6-9dfc-c23540ee6433.mp4

https://user-images.githubusercontent.com/17498701/199637035-850de0cd-d586-4135-9fdd-6a33b56fb949.mp4

### DIFF
--- a/Intersect.Client/Maps/MapInstance.cs
+++ b/Intersect.Client/Maps/MapInstance.cs
@@ -1036,7 +1036,7 @@ namespace Intersect.Client.Maps
             var yCount = Options.MapHeight * Options.TileHeight * 3 / fogTex.GetHeight();
 
             // Update the fog texture's position based on its speed and elapsed time.
-            mFogCurrentX -= elapsedTime / 1000f * FogXSpeed * -6;
+            mFogCurrentX += elapsedTime / 1000f * FogXSpeed * 2;
             mFogCurrentY += elapsedTime / 1000f * FogYSpeed * 2;
 
             // Handle cases where the fog texture's position goes out of bounds.
--- a/Intersect.Client/Maps/MapInstance.cs
+++ b/Intersect.Client/Maps/MapInstance.cs
@@ -1008,13 +1008,13 @@ namespace Intersect.Client.Maps
         /// </summary>
         public void DrawFog()
         {
-            // Exit early if the player or map data is not available, or if there is no fog texture
+            // Exit early if the player or map data is not available, or if there is no fog texture.
             if (Globals.Me == null || Lookup.Get(Globals.Me.MapId) == null || Fog == null || Fog.Length <= 0)
             {
                 return;
             }
 
-            // Get fog texture and exit early if it is not available
+            // Get fog texture and exit early if it is not available.
             var fogTex = Globals.ContentManager.GetTexture(Framework.Content.TextureType.Fog, Fog);
             if (fogTex == null)
             {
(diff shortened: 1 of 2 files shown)


Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1003,104 +1003,63 @@ private GameTileBuffer[][] DrawMapLayer(string layerName, float xoffset = 0, flo
...
 
-        //Fogs/Panorama/Overlay
+        /// <summary>
+        /// Draws the fog over the game view.
+        /// </summary>
         public void DrawFog()
         {
-            if (Globals.Me == null || Lookup.Get(Globals.Me.MapId) == null)
+            // Exit early if the player or map data is not available, or if there is no fog texture.
+            if (Globals.Me == null || Lookup.Get(Globals.Me.MapId) == null || Fog == null || Fog.Length <= 0)
```
lodicolo, line 1012: `string.IsNullOrWhiteSpace(Fog)` instead of `Fog == null || Fog.Length <= 0`

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1003,104 +1003,63 @@ private GameTileBuffer[][] DrawMapLayer(string layerName, float xoffset = 0, flo
...
+                ? Math.Min(1, mCurFogIntensity + elapsedTime / 2000f)
+                : Math.Max(0, mCurFogIntensity - elapsedTime / 2000f);
 
-                    if (mFogCurrentX < fogTex.GetWidth())
-                    {
-                        mFogCurrentX += fogTex.GetWidth();
-                    }
+            // Calculate the number of times the fog texture needs to be drawn to cover the map area.
+            var xCount = Options.MapWidth * Options.TileWidth * 3 / fogTex.GetWidth();
+            var yCount = Options.MapHeight * Options.TileHeight * 3 / fogTex.GetHeight();
```
lodicolo, line 1036: `Width`/`Height` instead of `GetWidth()`/`GetHeight()`

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1003,104 +1003,63 @@ private GameTileBuffer[][] DrawMapLayer(string layerName, float xoffset = 0, flo
...
+            mFogCurrentX += elapsedTime / 1000f * FogXSpeed * 2;
+            mFogCurrentY += elapsedTime / 1000f * FogYSpeed * 2;
 
-                    if (mFogCurrentY > fogTex.GetHeight())
-                    {
-                        mFogCurrentY -= fogTex.GetHeight();
-                    }
+            // Handle cases where the fog texture's position goes out of bounds.
+            mFogCurrentX %= fogTex.GetWidth();
+            mFogCurrentY %= fogTex.GetHeight();
```
lodicolo, line 1044: `Width`/`Height`

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1003,104 +1003,63 @@ private GameTileBuffer[][] DrawMapLayer(string layerName, float xoffset = 0, flo
...
-                                    GetX() - Options.MapWidth * Options.TileWidth * 1f + x * fogW + drawX,
-                                    GetY() - Options.MapHeight * Options.TileHeight * 1f + y * fogH + drawY, fogW, fogH
-                                ), new Intersect.Color((byte)(FogTransparency * mCurFogIntensity), 255, 255, 255),
-                                null, GameBlendModes.None
-                            );
-                        }
-                    }
+            for (var x = -1; x < xCount; x++)
+            {
+                for (var y = -1; y < yCount; y++)
```
lodicolo, line 1052: Why are these initialized to -1 when they are still `< Count`? What are the effects of changing it to start with `0` or end with `<= Count`?

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1003,104 +1003,63 @@ private GameTileBuffer[][] DrawMapLayer(string layerName, float xoffset = 0, flo
...
+                {
+                    var fogW = fogTex.GetWidth();
+                    var fogH = fogTex.GetHeight();
+                    Graphics.DrawGameTexture(
+                        fogTex, new FloatRect(0, 0, fogW, fogH),
+                        new FloatRect(
+                            GetX() - Options.MapWidth * Options.TileWidth * 1f + x * fogW + drawX,
+                            GetY() - Options.MapHeight * Options.TileHeight * 1f + y * fogH + drawY, fogW, fogH
+                        ), new Color((byte)(FogTransparency * mCurFogIntensity), 255, 255, 255)
+                    );
```
lodicolo, line 1062: Use properties when available

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1236,52 +1195,60 @@ public void DrawOverlayGraphic()
...
+
+                // Calculate displacement of current map compared to old map.
+                float dx = GetX() - oldMap.X;
+                float dy = GetY() - oldMap.Y;
+
+                // Update fog position based on displacement.
+                float fogWidth = fogTex.GetWidth();
+                float fogHeight = fogTex.GetHeight();
+                mFogCurrentX -= dx * Options.TileWidth * Options.MapWidth % fogWidth;
+                mFogCurrentY -= dy * Options.TileHeight * Options.MapHeight % fogHeight;
```
lodicolo, line 1229: 1. Use properties
2. Do not create local variable copies, just inline them

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1236,52 +1195,60 @@ public void DrawOverlayGraphic()
...
+                // Copy over fog values.
+                mFogUpdateTime = tempMap.mFogUpdateTime;
+                var ratio = (float)tempMap.FogTransparency / FogTransparency;
+                mCurFogIntensity = ratio * tempMap.mCurFogIntensity;
+                mFogCurrentX = tempMap.mFogCurrentX;
+                mFogCurrentY = tempMap.mFogCurrentY;
+
+                // Calculate displacement of current map compared to old map.
+                float dx = GetX() - oldMap.X;
+                float dy = GetY() - oldMap.Y;
```
lodicolo, line 1223: If the properties are available for X and Y use them

lodicolo (changes requested):

Intersect.Client/Maps/MapInstance.cs
Arufonsu: it seems to have the same outcome in game 🤔 i tho that X and y are initialized to -1 because the loop will run one extra time for each of them due to the < operator in the loop condition. This means that the inner loop should run an extra time for each outer loop iteration so it effectively drawing an extra row and column of fog texture. but in the practice, it seems to have no apparent change to set it to 0.

Should i change this and the one present within the editor to 0 then (?)

Arufonsu: changed them to the following on both ends:

imagen

**but within the editor:** there's some unwanted outcome :

imagen

imagen

i will leave them as they were originally for now.

Arufonsu: just figured out ~

lodicolo (approved):

lodicolo:
There are some issues I still have with this (namely <= instead of <) but I think this fix needs to go out

Arufonsu:
> There are some issues I still have with this (namely <= instead of <) but I think this fix needs to go out

I probably misunderstood [this one](https://github.com/AscensionGameDev/Intersect-Engine/pull/1694#discussion_r1059523823), let's chat about it tomorrow, gn !

PR #1695, opened 2022-12-31
fix: glitched even rows and hovered rows within listboxes

### LINKED ISSUE #1648: bug: glitchy even rows on list elements
### Description

There's an issue happening with Gwen's even rows on list elements. Apparently: even rows are supposed to be highlighted differently from non-even rows, but  it seems that this behavior isn't exactly working as intended.

image

image

### Steps to Reproduce

1. Check out any GUI window that contains a list within them.
2. Click over even rows.
3. See the issue.

### Version with bug

v0.7.1.81

Unknown

### Affected platforms

Windows, Linux

Setting this value here to false, will forcefully disable even rows from appearing on lists:

image

image

image

image

image

This took me over to open the following PR: #1646 in order to make this small behavior a configurable one instead of an enabled one by default ; although this is a good enhancement to have within the engine and serves as a workaround atm, **this isn't the real fix to the issue, which needs to be addressed eventually**.

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
* Fixes an issue where list box elements that are selected and hovered, are drawn with the selected texture instead of the hovered texture.
* Realigned listbox rows.
* Fixes glitchy and misaligned even rows.
* Should resolve #1648
* [Reworked assets with better looking listbox rows](https://github.com/AscensionGameDev/Intersect-Assets/pull/35)

### Before:

imagen

imagen

### Preview (_After_ this PR):

https://user-images.githubusercontent.com/17498701/211175865-f10e6a99-f4bf-4e28-a716-4310eee43fdb.mp4

### DIFF
--- a/Intersect.Client.Framework/Gwen/Skin/TexturedBase.cs
+++ b/Intersect.Client.Framework/Gwen/Skin/TexturedBase.cs
@@ -645,12 +645,12 @@ namespace Intersect.Client.Framework.Gwen.Skin
                 );
             }
 
-            mTextures.Input.ListBox.Background = new Bordered(mTexture, 256, 256, 63, 127, Margin.Eight);
-            mTextures.Input.ListBox.Hovered = new Bordered(mTexture, 320, 320, 31, 31, Margin.Eight);
-            mTextures.Input.ListBox.EvenLine = new Bordered(mTexture, 352, 256, 31, 31, Margin.Eight);
-            mTextures.Input.ListBox.OddLine = new Bordered(mTexture, 352, 288, 31, 31, Margin.Eight);
-            mTextures.Input.ListBox.EvenLineSelected = new Bordered(mTexture, 320, 270, 31, 31, Margin.Eight);
-            mTextures.Input.ListBox.OddLineSelected = new Bordered(mTexture, 320, 288, 31, 31, Margin.Eight);
+            mTextures.Input.ListBox.Background = new Bordered(mTexture, 256, 256, 63, 127, Margin.Six);
+            mTextures.Input.ListBox.Hovered = new Bordered(mTexture, 320, 320, 32, 32, Margin.Six);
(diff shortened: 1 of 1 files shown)


lodicolo (changes requested):
I see no difference between the even and odd rows in the preview? I would have expected it to based on the bug.

Arufonsu:
> I see no difference between the even and odd rows in the preview? I would have expected it to based on the bug.

Updated the PR information with:

### Before:

imagen

imagen

Arufonsu:
More graphic examples:

imagen

imagen

imagen

imagen


imagen

imagen

lodicolo:
@Arufonsu So my expectation is that _non-selected rows_ have zebra-striping, not vertically inverting the graphic for the selected row (that doesn't help with accessibility in the slightest).

There are essentially 3 colors for rows:
- background (unselected odd row)
- semi-inverted background (unselected even row) -- semi-inverted meaning it doesn't need to be a perfect inverse, it can be something like a 25% brightness if the background is black, or 75% brightness if background is white
- different color (selected row, doesn't matter if the row is odd or even)

[Zebra striping example (with link to article on its usefulness)
image](https://alistapart.com/article/zebrastripingdoesithelp/)

[Example table with me highlighting (aka "selection") a given row while maintaining zebra striping -- the three colors being gray, white, blue (not necessarily the colors we would be using, just an example)
image](https://www.w3schools.com/howto/howto_css_table_zebra.asp)

Arufonsu:
> zebra-striping

i see, okay, i can work around the assets to achieve this effect, but as for the code changes, this should be all we need to fix the initial misaligned / wrong sprite position from which Gwen was reading the rows sprites for even rows.

I'll work a bit with a skin to achieve what you just described tomorrow.

Arufonsu:
Just opened the following https://github.com/AscensionGameDev/Intersect-Assets/pull/37

https://user-images.githubusercontent.com/17498701/211175865-f10e6a99-f4bf-4e28-a716-4310eee43fdb.mp4

lodicolo (approved):

PR #1697, opened 2022-12-31
fix: overhead hp bar issues

### LINKED ISSUE #1696: bug: overhead hp bar issues
### Description

The player boolean override `ShouldDrawHpBar` is not properly allowing to render player's hp bars:

```
                if (ShouldNotDrawHpBar)
                {
                    return false;
                }
```

If we return false within this method, **we aren't allowing to loop thru the rest of conditions ahead during the itineration.** 

### Steps to Reproduce

1. Have any player with their HP at 100%
2. Enable or disable my/players overhead hp bar settings.
3. See hp bar not drawing in both of the case scenarios (enabled/full hp, disabled/hovered).

### Version with bug

0.7.1.92

Never did, this is a recently added feature

### Affected platforms

Windows, Linux

The player boolean override `ShouldDrawHpBar` should initially follow the same pattern as it's virtual boolean `ShouldDrawHpBar` in order to have the desired outcome

imagen

```
                if (!ShouldNotDrawHpBar || IsHovered)
                {
                    return true;
                }
```

or for better readability:

```

                if (!ShouldNotDrawHpBar)
                {
                    return true;
                }

                if (IsHovered)
                {
                    return true;
                }
```

### Relevant log output

```shell
No logs are required for this report.
```

### Duplicate Bug Check

Arufonsu:
Should resolve #1696

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -2378,12 +2378,7 @@ namespace Intersect.Client.Entities
         {
             get
             {
-                if (ShouldNotDrawHpBar)
-                {
-                    return false;
-                }
-
-                if (IsHovered)
+                if (!ShouldNotDrawHpBar || IsHovered)
                 {
                     return true;
                 }
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -2378,32 +2378,13 @@ namespace Intersect.Client.Entities
         {
             get
             {
-                if (!ShouldNotDrawHpBar || IsHovered)
-                {
-                    return true;
-                }
-
-                if (Id == Globals.Me.Id)
-                {
-                    return Globals.Database.MyOverheadHpBar;
-                }
-
-                if (Globals.Database.PartyMemberOverheadHpBar && Globals.Me.IsInMyParty(this))
-                {
-                    return true;
-                }
-
-                if (Globals.Database.FriendOverheadHpBar && Globals.Me.IsFriend(this))
-                {
-                    return true;
-                }
-
-                if (Globals.Database.GuildMemberOverheadHpBar && Globals.Me.IsGuildMate(this))
-                {
-                    return true;
-                }
-
-                return Globals.Database.PlayerOverheadHpBar;
+                var me = Globals.Me;
+                bool userPreference = (Globals.Database.MyOverheadHpBar && Id == me.Id) ||
+                                      (Globals.Database.PlayerOverheadHpBar && Id != me.Id) ||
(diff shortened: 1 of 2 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -2378,12 +2378,7 @@ protected override bool ShouldDrawHpBar
         {
             get
             {
-                if (ShouldNotDrawHpBar)
-                {
-                    return false;
-                }
-
-                if (IsHovered)
+                if (!ShouldNotDrawHpBar || IsHovered)
```
lodicolo, line 2381: This is 100% not a bug fix, it creates a bug. The `ShouldNotDrawHpBar` logic is not invertible like this.

lodicolo (changes requested):
If you really consider this a bug you need to provide a before and after as well as show all of the cases that ShouldNotDrawHpBar covers not breaking with this change.

Intersect.Client/Entities/Player.cs
Arufonsu: this was previously working and i changed it to according to the PR's review, my fault on this since i didn't fully tested the possible outcomes of changing it in the way that was previously suggested. Currently, the player boolean override `ShouldDrawHpBar` should initially follow the same pattern as it's entity virtual boolean `ShouldDrawHpBar` in order to have the desired outcome

imagen

if we return false within this method, we won't be able to show up the hp bars when players are damaged nor hovered, since we are returning false before the loop can even check after the rest of conditions ahead during the itineration.

lodicolo: https://github.com/AscensionGameDev/Intersect-Engine/blob/f9349c8d949c34c2586a621d42006153e35c9cfb/Intersect.Client/Entities/Entity.cs#L1423-L1437

But ShouldNotDrawHpBar is only true (resulting in a short-circuit to ShouldDrawHpBar to false) if HP is 0, HP is max, or if there is a shield -- which is actually where the problem lies.

image

I did not catch this in my review, but you maintained `|| shield > 0` when it needed to be `(<other conditions>) && shield < 1` in ShouldNotDrawHpBar.

Arufonsu: it should always been shield < 1 back then when the method was named "should draw"
the hp bar isn't supposed to be drawn if the shield bar is up ('ve never personally seen it in action but from what i understand this is how the logic flows). thus, shield > 0 is what we need in order to tell that the HP bar should not be drawn since the shield bar is up instead.

Arufonsu: just doubled checked all cases, not sure what i was into that day, couldn't really repro the issues i was experiencing today -facepalm- so i'm closing this one.

PR #1701, opened 2023-01-02
chore: overhead name rendering

Arufonsu:
~~* Reduced nesting and loops by eliminating unnecessary checks and combining conditions.~~
~~* Added caching of values within the DrawOverheadInfoOnHover method to improve performance.~~
~~* Simplified the DrawOverheadInfoOnHover method to improve readability and maintainability.~~
~~* Added caching of values within the DrawOverheadInfo method to improve performance.~~
~~* Simplified the DrawOverheadInfo method to improve readability and maintainability.~~

* Nukes a bunch of unnecessary and duplicated code, implementation feels cleaner and makes more sense !
* Removed entire method from graphics (w** is wrong with me)
* Removed entire duplicated / mirror method from players class.
* Entity DrawName loops thru the new boolean check `DrawNameRequirements()` with every other entity class in mind, so no overrides are really needed with this.
* `entity.HideName` is already checked by !ShouldDrawName
* `entity.MapInstance.InView()` is in theory already checked prior calling `DrawName()`  (Graphics.cs L458 ; L476 ) ; so we don't really need to filter with another list nor another loop at all.
* Removed unnecessary `MapInstance == null` check: is already checked within `DrawNameRequirements`.
* Checking for `Interface.Interface.MouseHitGui()` in the way were doing it makes no sense now, so it's part of the `IsHovered` check instead: this was previously leading to odd behaviour. ie. all entities names would hide when hovering any GUI element.


Intersect.Client/Core/Graphics.cs
```diff
@@ -522,9 +522,21 @@ public static void DrawInGame(TimeSpan deltaTime)
         // Draw Entities Overhead Information.
         private static void DrawOverheadInfo(Entity entity)
         {
-            // Who's who.
+            // Cache values.
```
lodicolo, line 525: This is fine

Intersect.Client/Core/Graphics.cs
```diff
@@ -541,53 +553,13 @@ private static void DrawOverheadInfo(Entity entity)
...
-            }
-
-            // If FriendOverheadInfo is toggled on, always draw friend's info.
-            else if (Globals.Database.FriendOverheadInfo && isFriend && !isPartyMate)
-            {
-                entity.DrawName(null);
-            }
-
-            // If GuildMemberOverheadInfo is toggled on, always draw guild members info.
-            else if (Globals.Database.GuildMemberOverheadInfo && isGuildMate && !isPartyMate)
```
lodicolo, line 590: This is not.

Use the old code and only compute the booleans as close to their using code as needed.

Use return statements as needed to break up the else-if chain into just ifs and comparisons as close the the consuming if statements as possible.

Intersect.Client/Core/Graphics.cs
```diff
@@ -522,9 +522,21 @@ public static void DrawInGame(TimeSpan deltaTime)
...
+                isOtherPlayer = false,
+                isFriend = false,
+                isGuildMate = false,
+                isPartyMate = false,
+                myOverheadPref = Globals.Database.MyOverheadInfo,
+                npcOverheadPref = Globals.Database.NpcOverheadInfo,
+                playerOverheadPref = Globals.Database.PlayerOverheadInfo,
+                friendOverheadPref = Globals.Database.FriendOverheadInfo,
+                guildMemberOverheadPref = Globals.Database.GuildMemberOverheadInfo,
+                partyMemberOverheadPref = Globals.Database.PartyMemberOverheadInfo;
```
lodicolo, line 538: This is just horrible. The code was bad before, this is introducing more lines that only front-load possibly unnecessary computation.

Intersect.Client/Core/Graphics.cs
```diff
@@ -541,53 +553,13 @@ private static void DrawOverheadInfo(Entity entity)
...
-
-            // If GuildMemberOverheadInfo is toggled on, always draw guild members info.
-            else if (Globals.Database.GuildMemberOverheadInfo && isGuildMate && !isPartyMate)
+            if (isEvent ||
+                (myOverheadPref && isMe) ||
+                (npcOverheadPref && isNpc) ||
+                (playerOverheadPref && isOtherPlayer && !isFriend && !isGuildMate && !isPartyMate) ||
+                (partyMemberOverheadPref && isPartyMate) ||
+                (friendOverheadPref && isFriend && !isPartyMate && !(isGuildMate && guildMemberOverheadPref)) ||
+                (guildMemberOverheadPref && isGuildMate && !isPartyMate && !(isFriend && friendOverheadPref)))
```
lodicolo, line 562: :vomiting_face:

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
         {
-            if (Interface.Interface.MouseHitGui())
+            var map = this.MapInstance;
+            if (map == null || Interface.Interface.MouseHitGui())
             {
                 return;
             }
 
-            var mousePos = Graphics.ConvertToWorldPoint(Globals.InputManager.GetMousePosition());
-            foreach (MapInstance map in Maps.MapInstance.Lookup.Values)
```
lodicolo, line 2264: This looks functionally different from what you changed it to.

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
+            // Cache values available out of the loop.
+            Type[] ignoreEntityTypes = { typeof(Projectile), typeof(Resource), typeof(Event) };
```
lodicolo, line 2265: This caching is unnecessary.

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
-
-                            if (en.Value.MapId == mapId &&
-                                !en.Value.HideName &&
-                                (!en.Value.IsStealthed || isPlayer && Globals.Me.IsInMyParty(player)) &&
-                                en.Value.WorldPos.Contains(mousePos.X, mousePos.Y))
-                            {
-                                // We don't want to deal with these entities when hovering the cursor by them.
-                                var ignoreEntities = !(en.Value is Projectile || en.Value is Resource || en.Value is Event);
-
-                                if (ignoreEntities)
```
lodicolo, line 2292: This if statement just needed to be inverted into a continue, specifically

```csharp
if (en.Value is Projectile || en.Value is Resource || en.Value is Event)
{
    continue;
}
```

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
-                                {
-                                    // Who's who.
-                                    var isMe = isPlayer && player.Id == Globals.Me.Id;
-                                    var isOtherPlayer = isPlayer && !isMe;
-                                    var isFriend = isOtherPlayer && Globals.Me.IsFriend(player);
-                                    var isGuildMate = isOtherPlayer && Globals.Me.IsGuildMate(player);
-                                    var isPartyMate = isOtherPlayer && Globals.Me.IsInMyParty(player);
```
lodicolo, line 2299: This caching is horrible and unnecessary

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
-                                    }
+                // Cache values that are only available within the loop.
+                var player = en.Value as Player;
+                bool isPlayer = en.Value is Player,
+                    isNpc = !isPlayer,
+                    isMe = isPlayer && player.Id == Globals.Me.Id,
+                    isOtherPlayer = isPlayer && !isMe,
+                    isFriend = isOtherPlayer && Globals.Me.IsFriend(player),
+                    isGuildMate = isOtherPlayer && Globals.Me.IsGuildMate(player),
+                    isPartyMate = isOtherPlayer && Globals.Me.IsInMyParty(player);
```
lodicolo, line 2300: But don't replace that garbage caching with this either.

Intersect.Client/Entities/Player.cs
```diff
@@ -2252,124 +2252,71 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
 
-                        break;
-                    }
+                // Draw Overhead Information.
+                if ((isNpc && !npcOverheadPref) ||
+                    (isMe && !myOverheadPref) ||
+                    (isOtherPlayer && !isFriend && !isGuildMate && !isPartyMate && !playerOverheadPref) ||
+                    (isFriend && !friendOverheadPref && !isPartyMate && !(isGuildMate && guildMemberOverheadPref)) ||
+                    (isGuildMate && !guildMemberOverheadPref && !isPartyMate && !(isFriend && friendOverheadPref)) ||
+                    (isPartyMate && !partyMemberOverheadPref))
```
lodicolo, line 2317: And don't reduce a bunch of if statements into this monstrosity.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: we already exited early if this.MapInstance is null, we are also exiting early if the entities map id doesn't belongs to that map instance, thus, we don't really need all these checks and loops anymore:

```
foreach (MapInstance map in Maps.MapInstance.Lookup.Values)
            {
                if (mousePos.X >= map.GetX() && mousePos.X <= map.GetX() + Options.MapWidth * Options.TileWidth)
                {
                    if (mousePos.Y >= map.GetY() && mousePos.Y <= map.GetY() + Options.MapHeight * Options.TileHeight)
                    {
```

but instead, with less nested code and much more simple and less resource consuming, we do the following:

`en.Value.WorldPos.Contains(mousePos.X, mousePos.Y)`

from which we have already filtered out the entities we don't want to check after.

Intersect.Client/Entities/Player.cs
Arufonsu: removed `isNpc` and `isNpc` since they were used only once. kept the rest since they are used 2 to 3 + times.

Arufonsu: got rid of the variables that were used only once and took another approach by moving these out so we don't iterate though these booleans on every foreach loop.

Intersect.Client/Entities/Player.cs
lodicolo: But the previous loop checked to see if an entity belonged to the map instance that contains the world coordinate the mouse points to -- that is not necessarily `this.MapInstance`.

If there is an syntax change that needs to happen it would be to refactor the bounds checks into a method and then invert the checks into `continue;` statements to skip additional maps.

If there is a performance optimization that needs to happen, I would think it would be just looping through the entities of the given map, not looping through the global entities (why are we even doing this?).

Intersect.Client/Entities/Player.cs
```diff
@@ -2251,125 +2251,124 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
+                else if (!Globals.Database.GuildMemberOverheadInfo && isGuildMate && !isPartyMate)
+                {
+                    // Skip if guild mate is a friend.
+                    if (Globals.Database.FriendOverheadInfo && isFriend)
+                    {
+                        continue;
                     }
+
+                    en.Value.DrawName(null);
                 }
```
lodicolo, line 2372: I think all of this logic needs to be looked at with a fine-toothed comb and simplified, as well as detailed explanations if things are double-checked.

Intersect.Client/Entities/Player.cs
```diff
@@ -2251,125 +2251,124 @@ public virtual void DrawGuildName(Color textColor, Color borderColor = null, Col
...
+                // If NpcOverheadInfo is toggled off, draw NPCs overhead information.
+                else if (!Globals.Database.NpcOverheadInfo && isNpc)
+                {
+                    en.Value.DrawName(null);
+                }
 
-                                        continue;
-                                    }
+                // If PlayerOverheadInfo is toggled off, draw other player's overhead information.
+                else if (!Globals.Database.PlayerOverheadInfo && isOtherPlayer &&
```
lodicolo, line 2331: In particular this if statement looks like it might be an else condition, but I am not sure because the if-blocks below it are too tangled and I can't tell just by eyeballing it.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: excluding the entities from other maps lead to the issue where we can't hover/see entities info outside the current map.

Now, how about !en.Value.MapInstance.InView() ? 

imagen

imagen

seems to behave as intended but i'm not sure if it's really allowing me to exclude computing within the loop all of the entities

Arufonsu: > If there is a performance optimization that needs to happen, I would think it would be just looping through the entities of the given map, not looping through the global entities (why are we even doing this?).

This approach is possible with the drawback i mentioned in the previous comment (inability to hover over entities outside the current map boundaries - which i _personally_ don't mind since i use unconnected maps, but it may be annoying to users doing so)

Intersect.Client/Entities/Entity.cs
```diff
@@ -1323,6 +1317,41 @@ public virtual void DrawName(Color textColor, Color borderColor = null, Color ba
             );
         }
 
+        private bool DrawNameRequirements()
```
lodicolo, line 1320: Almost all of this logic should already be in ShouldDrawName or ShouldNotDrawName if it exists. I don't see how this method is useful, it looks like it is rechecking things unnecessarily.

lodicolo (changes requested):

Intersect.Client/Entities/Entity.cs
Arufonsu: there wasn't nor there is a `ShouldNotDrawName` method for the logic behind this so far

moved over the logic to ShouldDrawName

Intersect.Client/Entities/Entity.cs
```diff
@@ -856,7 +856,71 @@ public virtual bool IsStealthed
...
+                        else if (Globals.Database.FriendOverheadInfo && me.IsFriend(player))
+                        {
+                            meetsRequirements = true;
+                        }
+                        else if (Globals.Database.GuildMemberOverheadInfo && me.IsGuildMate(player))
+                        {
+                            meetsRequirements = true;
+                        }
+
+                        return meetsRequirements;
```
lodicolo, line 913: Instead of the variable and `else`s, just do returns

Also add a blank line between all of the returns and the case after them, it makes the cases blend together without the vertical white space

Intersect.Client/Entities/Entity.cs
```diff
@@ -856,7 +856,71 @@ public virtual bool IsStealthed
...
+                    default:
+                        if (IsHovered)
+                        {
+                            return true;
+                        }
+
+                        return Globals.Database.NpcOverheadInfo;
```
lodicolo, line 920: This can be `return IsHovered || Globals.Database.NpcOverheadInfo;`

Intersect.Client/Entities/Entity.cs
```diff
@@ -1240,8 +1304,8 @@ public virtual float GetTop(int overrideHeight = -1)
 
         public virtual void DrawName(Color textColor, Color borderColor = null, Color backgroundColor = null)
         {
-            // Are we supposed to hide this name?
-            if (!ShouldDrawName || string.IsNullOrWhiteSpace(Name))
+            // Are we really supposed to draw this name?
+            if (!ShouldDrawName)
```
lodicolo, line 1308: Why remove the IsNullOrWhiteSpace check?

lodicolo (changes requested):

Intersect.Client/Entities/Entity.cs
Arufonsu: because it's already happening within `ShouldDrawName`, which will be false in the case of `string.IsNullOrWhiteSpace(Name)`

lodicolo (approved):

PR #1702, opened 2023-01-02
chore: performance and memory

Arufonsu:
* Adds comments and reorganizes the BankWindow class (Close() should be placed by the end of initialization).
* Improves the performance and readability of Gwen's Position method.
* Replaces a Pointf struct with separate float variables and uses ref variables within the Rotate method of the DrawTexture class to improve performance and save memory.
* Simplified the LoadFont method within the MonoRenderer class.
* Uses a for loop that iterates through children in reverse order and removes unnecessary variables within the GetControlAt and DoCacheRender methods of the Graphics and Gwen classes to improve performance.

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/Base.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Base.cs
@@ -2366,15 +2366,14 @@ namespace Intersect.Client.Framework.Gwen.Control
         /// <returns>Control or null if not found.</returns>
         public virtual Base GetControlAt(int x, int y)
         {
-            // Return null if control is hidden or coordinates are outside the control's bounds
+            // Return null if control is hidden or coordinates are outside the control's bounds.
             if (IsHidden || x < 0 || y < 0 || x >= Width || y >= Height)
             {
                 return null;
             }
 
-            // Check children in reverse order (last added first)
-            var count = mChildren.Count;
-            for (int i = count - 1; i >= 0; i--)
+            // Check children in reverse order (last added first).
+            for (int i = mChildren.Count - 1; i >= 0; i--)
             {
                 var child = mChildren[i];
                 var found = child.GetControlAt(x - child.X, y - child.Y);
@@ -2384,7 +2383,7 @@ namespace Intersect.Client.Framework.Gwen.Control
                 }
             }
 
-            // Return control if it is mouse input enabled, otherwise return null
+            // Return control if it is mouse input enabled, otherwise return null.
             return MouseInputEnabled ? this : null;
         }
 
--- a/Intersect.Client.Framework/Gwen/Control/Base.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Base.cs
@@ -11,7 +11,6 @@ using Intersect.Client.Framework.Gwen.Control.EventArguments;
 using Intersect.Client.Framework.Gwen.ControlInternal;
 using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Gwen.Input;
-using Intersect.Client.Framework.Audio;
 #if DEBUG || DIAGNOSTIC
 #endif
 using Newtonsoft.Json;
@@ -1751,49 +1750,48 @@ namespace Intersect.Client.Framework.Gwen.Control
         }
 
         /// <summary>
-        ///     Positions the control inside its parent.
+        /// Positions the control inside its parent.
         /// </summary>
         /// <param name="pos">Target position.</param>
-        /// <param name="xpadding">X padding.</param>
-        /// <param name="ypadding">Y padding.</param>
-        public virtual void Position(Pos pos, int xpadding = 0, int ypadding = 0) // todo: a bit ambiguous name
+        /// <param name="xPadding">X padding.</param>
+        /// <param name="yPadding">Y padding.</param>
+        public virtual void Position(Pos pos, int xPadding = 0, int yPadding = 0)
         {
-            var w = Parent.Width;
-            var h = Parent.Height;
-            var padding = Parent.Padding;
+            // Cache frequently used values.
+            int parentWidth = Parent.Width;
+            int parentHeight = Parent.Height;
+            Padding padding = Parent.Padding;
 
-            var x = X;
-            var y = Y;
-            if (0 != (pos & Pos.Left))
+            // Calculate the new X and Y positions using bitwise operations.
+            int x = X;
+            int y = Y;
+            if ((pos & Pos.Left) != 0)
             {
-                x = padding.Left + xpadding;
+                x = padding.Left + xPadding;
+            }
+            else if ((pos & Pos.Right) != 0)
+            {
+                x = parentWidth - Width - padding.Right - xPadding;
+            }
+            else if ((pos & Pos.CenterH) != 0)
+            {
+                x = (int)(padding.Left + xPadding + (parentWidth - Width - padding.Left - padding.Right) * 0.5f);
             }
 
-            if (0 != (pos & Pos.Right))
+            if ((pos & Pos.Top) != 0)
             {
-                x = w - Width - padding.Right - xpadding;
+                y = padding.Top + yPadding;
+            }
+            else if ((pos & Pos.Bottom) != 0)
+            {
+                y = parentHeight - Height - padding.Bottom - yPadding;
+            }
+            else if ((pos & Pos.CenterV) != 0)
+            {
+                y = (int)(padding.Top + yPadding + (parentHeight - Height - padding.Bottom - padding.Top) * 0.5f);
             }
 
-            if (0 != (pos & Pos.CenterH))
-            {
-                x = (int) (padding.Left + xpadding + (w - Width - padding.Left - padding.Right) * 0.5f);
-            }
-
-            if (0 != (pos & Pos.Top))
-            {
-                y = padding.Top + ypadding;
-            }
-
-            if (0 != (pos & Pos.Bottom))
-            {
-                y = h - Height - padding.Bottom - ypadding;
-            }
-
-            if (0 != (pos & Pos.CenterV))
-            {
-                y = (int) (padding.Top + ypadding + (h - Height - padding.Bottom - padding.Top) * 0.5f);
-            }
-
+            // Set the new position.
             SetPosition(x, y);
         }
 
@@ -1853,61 +1851,56 @@ namespace Intersect.Client.Framework.Gwen.Control
         /// <param name="master">Root parent.</param>
         protected virtual void DoCacheRender(Skin.Base skin, Base master)
         {
-            var render = skin.Renderer;
-            var cache = render.Ctt;
+            var renderer = skin?.Renderer;
+            var cache = renderer?.Ctt;
 
+            // Return early if the cache is not available
             if (cache == null)
             {
                 return;
             }
 
-            var oldRenderOffset = render.RenderOffset;
-            var oldRegion = render.ClipRegion;
+            // Save current render offset and clip region
+            var oldRenderOffset = renderer.RenderOffset;
+            var oldRegion = renderer.ClipRegion;
 
+            // Set render offset and clip region based on whether this control is the root parent
             if (this != master)
             {
-                render.AddRenderOffset(Bounds);
-                render.AddClipRegion(Bounds);
+                renderer.AddRenderOffset(Bounds);
+                renderer.AddClipRegion(Bounds);
             }
             else
             {
-                render.RenderOffset = Point.Empty;
-                render.ClipRegion = new Rectangle(0, 0, Width, Height);
+                renderer.RenderOffset = Point.Empty;
+                renderer.ClipRegion = new Rectangle(0, 0, Width, Height);
             }
 
-            if (mCacheTextureDirty && render.ClipRegionVisible)
+            // Render the control and its children if the cache is dirty and the clip region is visible
+            if (mCacheTextureDirty && renderer.ClipRegionVisible)
             {
-                render.StartClip();
+                renderer.StartClip();
 
+                // Setup the cache texture if necessary
                 if (ShouldCacheToTexture)
                 {
                     cache.SetupCacheTexture(this);
                 }
 
-                //Render myself first
-                //var old = render.ClipRegion;
-                //render.ClipRegion = Bounds;
-                //var old = render.RenderOffset;
-                //render.RenderOffset = new Point(Bounds.X, Bounds.Y);
+                // Render the control
                 Render(skin);
 
-                //render.RenderOffset = old;
-                //render.ClipRegion = old;
-
-                if (mChildren.Count > 0)
+                // Render the children (Reverse).
+                for (int i = 0; i < mChildren.Count; i++)
                 {
-                    //Now render my kids
-                    for (int i = 0; i < mChildren.Count; i++)
+                    var child = mChildren[i];
+                    if (!child.IsHidden)
                     {
-                        if (mChildren[i].IsHidden)
-                        {
-                            continue;
-                        }
-
-                        mChildren[i].DoCacheRender(skin, master);
+                        child.DoCacheRender(skin, master);
                     }
                 }
 
+                // Finish the cache texture if necessary
                 if (ShouldCacheToTexture)
                 {
                     cache.FinishCacheTexture(this);
@@ -1915,10 +1908,12 @@ namespace Intersect.Client.Framework.Gwen.Control
                 }
             }
 
-            render.ClipRegion = oldRegion;
-            render.StartClip();
-            render.RenderOffset = oldRenderOffset;
+            // Restore the original clip region and render offset
+            renderer.ClipRegion = oldRegion;
+            renderer.StartClip();
+            renderer.RenderOffset = oldRenderOffset;
 
+            // Draw the cached control texture if necessary
             if (ShouldCacheToTexture)
             {
                 cache.DrawCachedControlTexture(this);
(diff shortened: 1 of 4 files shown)


Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2580,24 +2566,26 @@ public bool IsChild(Base child)
         /// <returns>Canvas coordinates.</returns>
         public virtual Point LocalPosToCanvas(Point pnt)
         {
-            if (mParent != null)
+            if (mParent == null)
```
lodicolo, line 2569: Inverting this was a good decision

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2580,24 +2566,26 @@ public bool IsChild(Base child)
...
-                    x += mParent.mInnerPanel.X;
-                    y += mParent.mInnerPanel.Y;
-                }
+            var x = pnt.X + X;
+            var y = pnt.Y + Y;
 
+            // If our parent has an innerpanel and we're a child of it
+            // add its offset onto us.
+            //
+            if (mParent.mInnerPanel == null || !mParent.mInnerPanel.IsChild(this))
```
lodicolo, line 2580: Inverting this was not

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2580,24 +2566,26 @@ public bool IsChild(Base child)
...
+            }
 
-                // If our parent has an innerpanel and we're a child of it
-                // add its offset onto us.
-                //
-                if (mParent.mInnerPanel != null && mParent.mInnerPanel.IsChild(this))
-                {
-                    x += mParent.mInnerPanel.X;
-                    y += mParent.mInnerPanel.Y;
-                }
```
lodicolo, line 2595: Should keep this

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2607,24 +2595,26 @@ public virtual Point LocalPosToCanvas(Point pnt)
...
+            if (mParent.mInnerPanel == null || !mParent.mInnerPanel.IsChild(this))
+            {
                 return mParent.CanvasPosToLocal(new Point(x, y));
             }
 
-            return pnt;
+            x -= mParent.mInnerPanel.X;
+            y -= mParent.mInnerPanel.Y;
+
+            return mParent.CanvasPosToLocal(new Point(x, y));
```
lodicolo, line 2617: Do the -= inside of the if statement like the += in the method above

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2607,24 +2595,26 @@ public virtual Point LocalPosToCanvas(Point pnt)
         /// <returns>Local coordinates.</returns>
         public virtual Point CanvasPosToLocal(Point pnt)
         {
-            if (mParent != null)
+            if (mParent == null)
             {
-                var x = pnt.X - X;
-                var y = pnt.Y - Y;
+                return pnt;
+            }
```
lodicolo, line 2601: This is fine

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -615,33 +610,45 @@ public override GameTileBuffer CreateTileBuffer()
...
-                var pnt1 = new Pointf((float) tw, 0);
-                var pnt2 = new Pointf(0, (float) th);
-                var cntr = new Pointf((float) tw / 2, (float) th / 2);
-
-                var pntMod = Rotate(pnt, cntr, rotationDegrees);
-                var pntMod2 = Rotate(pnt1, cntr, rotationDegrees);
-                var pntMod3 = Rotate(pnt2, cntr, rotationDegrees);
-
-                var width = (int) Math.Round(GetDistance(pntMod.X, pntMod.Y, pntMod2.X, pntMod2.Y));
-                var height = (int) Math.Round(GetDistance(pntMod.X, pntMod.Y, pntMod3.X, pntMod3.Y));
```
lodicolo, line 632: This old code is better (more legible, less confusing), change it back.

The new code has no real performance benefit.

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -615,33 +610,45 @@ public override GameTileBuffer CreateTileBuffer()
...
+
+                var width = (int)Math.Round(GetDistance(pntX, pntY, pnt1X, pnt1Y));
+                var height = (int)Math.Round(GetDistance(pntX, pntY, pnt2X, pnt2Y));
 
                 if (packRotated)
                 {
-                    var z = width;
-                    width = height;
-                    height = z;
+                    (width, height) = (height, width);
```
lodicolo, line 633: This is fine

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -682,12 +674,20 @@ private static double GetDistance(double x1, double y1, double x2, double y2)
...
+            float c = (float)Math.Cos(angle);
+
+            x -= centerX;
+            y -= centerY;
+
+            float newX = x * c - y * s;
+            float newY = x * s + y * c;
+
+            x = newX + centerX;
+            y = newY + centerY;
```
lodicolo, line 690: Change it back

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -739,7 +739,7 @@ public override List<string> GetValidVideoModes()
 
             var allowedResolutions = new[]
             {
-                new Resolution(800, 600),
+                new Resolution(800),
```
lodicolo, line 742: Don't remove the `, 600`

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -246,16 +244,16 @@ public override bool Begin()
                 UpdateGraphicsState(mScreenWidth, mScreenHeight);
             }
 
-            StartSpritebatch(mCurrentView, GameBlendModes.None, null, null, true, null);
+            StartSpritebatch(mCurrentView, GameBlendModes.None, null, null, true);
```
lodicolo, line 247: Don't remove the `, null`

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -830,31 +830,25 @@ public void Init(GraphicsDevice graphicsDevice)
...
-                }
+            // Check if the font size can be extracted
+            if (parts.Length < 1 || !int.TryParse(parts[parts.Length - 1], out var size))
+            {
+                return null;
             }
 
-            return null;
+            // Concatenate the parts of the name except the last one to get the full name
+            name = string.Join("_", parts.Take(parts.Length - 1));
```
lodicolo, line 848: `'_'` instead of the string

Intersect.Client/MonoGame/IntersectGame.cs
```diff
@@ -80,9 +80,7 @@ private IntersectGame(IClientContext context, Action postStartupAction)
 
             mGraphics = new GraphicsDeviceManager(this)
             {
-                PreferredBackBufferWidth = 800,
-                PreferredBackBufferHeight = 480,
-                PreferHalfPixelOffset = true
+                PreferredBackBufferWidth = 800, PreferredBackBufferHeight = 480, PreferHalfPixelOffset = true
```
lodicolo, line 83: Change this back

Intersect.Client/MonoGame/IntersectGame.cs
```diff
@@ -117,19 +114,29 @@ private IntersectGame(IClientContext context, Action postStartupAction)
             Interface.Interface.GwenInput = new IntersectInput();
             Controls.Init();
 
-            Window.Position = new Microsoft.Xna.Framework.Point(-20, -2000);
+            // Reuse Point object instead of creating a new one each time.
+            var hiddenWindowPosition = new Microsoft.Xna.Framework.Point(-20, -2000);
+            Window.Position = hiddenWindowPosition;
```
lodicolo, line 119: 1. Point is not an object
2. you are not reusing it anywhere

Change this back

lodicolo (changes requested):

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: whops 😮 🤣

Arufonsu: oh, this is why:

imagen

i will restore it back anyways.

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: imagen

Reason ^

Restoring this back anyways.

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: Replaced the Pointf struct with separate float variables for the x and y coordinates becayse it saves memory (Pointf is a struct and requires more memory than two separate float variables). Also modified the Rotate method to accept the coordinates as ref variables rather than as a Pointf struct. This allows the method to directly modify the original variables **rather than creating new ones**, which _can_ save memory and improve performance. As for cntr, it was unnecessary to create it as Pointf variable under this scenario so i replaced it with two separate float variables for the x and y coordinates. This also _should_ save some memory. i essentially tried to solve the todo from the comment within the code here.

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: see https://github.com/AscensionGameDev/Intersect-Engine/pull/1702/files#r1060247028

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
lodicolo: > Pointf is a struct and requires more memory than two separate float variables

Pointf _is_ a struct. That is exactly why it _**does not**_ require more memory than two separate float variables.

All value types (structs, enums, non-string primitives) have _no_ additional overhead.

_Classes_ have additional memory consumption because of the overhead of inheritance (i.e. pointers to vtables for overloads, etc).

Arufonsu: I'll change it back then 👍🏽

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: [same](https://github.com/AscensionGameDev/Intersect-Engine/pull/1702#discussion_r1060807951) for this

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -2371,35 +2366,26 @@ protected virtual void OnChildTouched(Base control)
...
-                return null;
-            }
-
-            // todo: convert to linq FindLast
-            var rev = ((IList<Base>) mChildren)
-                .Reverse(); // IList.Reverse creates new list, List.Reverse works in place.. go figure
-
-            foreach (var child in rev)
+            // Check children in reverse order (last added first)
+            var count = mChildren.Count;
```
lodicolo, line 2376: This is not necessary, should be inlined.

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -426,18 +428,13 @@ public override GameRenderTexture CreateRenderTexture(int width, int height)
...
-            }
-
-            var font = (SpriteFont) gameFont.GetFont();
+            var font = (SpriteFont)gameFont?.GetFont();
             if (font == null)
             {
                 return;
             }
 
-            StartSpritebatch(mCurrentView, GameBlendModes.None, null, renderTexture, false, null);
```
lodicolo, line 440: Add back the `, false, null`

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -615,33 +610,43 @@ public override GameTileBuffer CreateTileBuffer()
...
+            // Cache the result of ConvertColor(renderColor) in a temporary variable.
+            var color = ConvertColor(renderColor);
+
+            // Use a single Draw method to avoid duplicating code.
+            void Draw(FloatRect currentView, GameRenderTexture targetObject)
+            {
+                StartSpritebatch(currentView, blendMode, shader, targetObject, false, null, drawImmediate);
+                mSpriteBatch.Draw((Texture2D)texture, new Vector2(tx, ty),
+                    new XNARectangle((int)sx, (int)sy, (int)sw, (int)sh), color, rotationDegrees, origin,
+                    new Vector2(tw / sw, th / sh), SpriteEffects.None, 0);
```
lodicolo, line 647: Break each argument of Draw() into a separate line to make it more readable please, also the starting ( and terminating ); should be on separate lines from the first/last argument

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -830,31 +820,25 @@ public void Init(GraphicsDevice graphicsDevice)
...
-                }
+            // Check if the font size can be extracted
+            if (parts.Length < 1 || !int.TryParse(parts[parts.Length - 1], out var size))
+            {
+                return null;
             }
 
-            return null;
+            // Concatenate the parts of the name except the last one to get the full name
+            name = string.Join($"{'_'}", parts.Take(parts.Length - 1));
```
lodicolo, line 838: Uh this should just be `'_'`, do not use an interpolated string.

lodicolo (changes requested):

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: can't use it as char directly for string.Join

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: imagen

imagen


( reverting them back )

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: imagen

other way around:
imagen

Arufonsu: actually, `'_'` is used twice, i'll make a variable for it

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -826,7 +827,8 @@ public override GameFont LoadFont(string filename)
                 .TrimStart(Path.DirectorySeparatorChar);
 
             // Split the name into parts
-            var parts = name.Split('_');
+            const char separator = '_';
+            var parts = name.Split(separator);
```
lodicolo, line 831: Storing the separator in variable is not needed, just leave the literal

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
```diff
@@ -835,7 +837,7 @@ public override GameFont LoadFont(string filename)
             }
 
             // Concatenate the parts of the name except the last one to get the full name
-            name = string.Join($"{'_'}", parts.Take(parts.Length - 1));
+            name = string.Join(separator.ToString(), parts.Take(parts.Length - 1));
```
lodicolo, line 840: I keep trying to stop you from casting the char into a string but you keep finding different ways to convert it to a string.

You should be using [Join\<T\>(String, IEnumerable\<T\>)](https://learn.microsoft.com/en-us/dotnet/api/system.string.join?view=net-7.0#system-string-join-1(system-string-system-collections-generic-ienumerable((-0))))

lodicolo (changes requested):

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
lodicolo: Apparently this method doesn't actually exist in .NET Framework, so use

`string.Join('_', parts, 0, parts.Length - 1)`

lodicolo (changes requested):

Intersect.Client/MonoGame/Graphics/MonoRenderer.cs
Arufonsu: > Cannot resolve method 'Join(char, System.Collections.Generic.IEnumerable<string>)', candidates are: 
> string Join(string, params object[]) (in class String) 
> string Join(string, params string[]) (in class String) 
> string Join(string, System.Collections.Generic.IEnumerable<string>) (in class String)
> string Join<string>(string, System.Collections.Generic.IEnumerable<string>) (in class String)

Arufonsu: The following :

`name = string.Join("'_'", 0, parts.Length - 1);`

nor the following work

`name = string.Join('_', 0, parts.Length - 1);`

we can't use chars directly here, it has to be a char converted to it or the string right way, previously it was using a string like this:

imagen

which takes me to the following (and works to load the fonts too)

imagen

lodicolo (approved):

PR #1707, opened 2023-01-20
chore: overhead hp bar

Arufonsu:
* Simplified code, tested and covered all scenarios.

### DIFF
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1416,8 +1416,22 @@ namespace Intersect.Client.Entities
                     return false;
                 }
 
-                return (GetType() == typeof(Entity) && Globals.Database.NpcOverheadHpBar) ||
-                       Vital[(int)Vitals.Health] != MaxVital[(int)Vitals.Health] || GetShieldSize() > 0 || IsHovered;
+                if (IsHovered)
+                {
+                    return true;
+                }
+
+                if (GetShieldSize() > 0)
+                {
+                    return true;
+                }
+
+                if (Vital[(int)Vitals.Health] != MaxVital[(int)Vitals.Health])
+                {
+                    return true;
+                }
+
+                return GetType() == typeof(Entity) && Globals.Database.NpcOverheadHpBar;
             }
         }
 
--- a/Intersect.Client/Entities/Entity.cs
+++ b/Intersect.Client/Entities/Entity.cs
@@ -1411,12 +1411,13 @@ namespace Intersect.Client.Entities
         {
             get
             {
-                if (!ShouldNotDrawHpBar || IsHovered)
+                if (ShouldNotDrawHpBar)
                 {
-                    return true;
+                    return false;
                 }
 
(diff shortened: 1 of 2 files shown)


WeylonSantana (approved):

Arufonsu:
For the record: the following was untested until now 🧪 


https://user-images.githubusercontent.com/17498701/214177794-c7473780-2074-4add-985a-7edd53dc50ae.mp4

Intersect.Client/Entities/Entity.cs
```diff
@@ -1411,29 +1411,21 @@ protected virtual bool ShouldDrawHpBar
...
             {
-                if (!ShouldNotDrawHpBar || IsHovered)
+                if (ShouldNotDrawHpBar)
                 {
-                    return true;
+                    return false;
                 }
 
-                return GetType() == typeof(Entity) && Globals.Database.NpcOverheadHpBar;
+                return (GetType() == typeof(Entity) && Globals.Database.NpcOverheadHpBar) ||
```
lodicolo, line 1419: Check IsHovered first, then shield, then HP, and only then should you check the entity type.

This should also be broken up into multiple if statements so it isn't such a long line.

Intersect.Client/Entities/Player.cs
```diff
@@ -2383,32 +2383,13 @@ protected override bool ShouldDrawHpBar
...
-                }
-
-                return Globals.Database.PlayerOverheadHpBar;
+                var me = Globals.Me;
+                bool userPreference = (Globals.Database.MyOverheadHpBar && Id == me.Id) ||
+                                      (Globals.Database.PlayerOverheadHpBar && Id != me.Id) ||
+                                      (Globals.Database.PartyMemberOverheadHpBar && me.IsInMyParty(this)) ||
+                                      (Globals.Database.FriendOverheadHpBar && me.IsFriend(this)) ||
+                                      (Globals.Database.GuildMemberOverheadHpBar && me.IsGuildMate(this));
+                return userPreference || IsHovered;
```
lodicolo, line 2392: Revert this. This is not more readable, and forces expensive operations to happen before IsHovered which is a much less costly check.

The original structure and order was very intentional.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
```diff
@@ -2388,27 +2388,31 @@ protected override bool ShouldDrawHpBar
...
-                    return true;
+                    userPreference = true;
+                }
+                else if (Globals.Database.GuildMemberOverheadHpBar && me.IsGuildMate(this))
+                {
+                    userPreference = true;
                 }
 
-                return Globals.Database.PlayerOverheadHpBar;
+                return userPreference;
```
lodicolo, line 2415: Get rid of this extra variable, and all of the `else` keywords.

Just return true (and return <the last if statement>)

lodicolo (changes requested):

lodicolo (approved):

PR #1710, opened 2023-01-24
enhancement: Threaded`1.cs TODO

Arufonsu:
- Created a utility method that checks the type of each element in the args array against the expected type TArgument. If any element in the array is not of type TArgument, the utility method should throw an exception indicating that the argument type is incorrect.
- Solves a TODO

### DIFF
--- a/Intersect (Core)/Threading/ArgumentTypeChecker.cs
+++ b/Intersect (Core)/Threading/ArgumentTypeChecker.cs
@@ -1,0 +1,33 @@
+﻿using System;
+
+namespace Intersect.Threading
+{
+    /// <summary>
+    /// Utility class for checking the type of arguments passed to a new thread.
+    /// </summary>
+    /// <typeparam name="TArgument">Type of the argument that passes to a new thread.</typeparam>
+    public static class ArgumentTypeChecker<TArgument>
+    {
+        /// <summary>
+        /// Checks the type of each element within `args` array against the expected type `TArgument`. 
+        /// If any element in the array is not of type `TArgument` -> An exception is thrown.
+        /// </summary>
+        /// <param name="args">Array of arguments to be checked.</param>
+        public static void CheckArgumentTypes(object[] args)
+        {
+            if (args == null || args.Length < 1)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(args), $@"Expected one argument of type {typeof(TArgument).FullName}."
+                );
+            }
+
+            if (!(args[0] is TArgument))
+            {
+                throw new ArgumentException(
+                    $"Expected argument of type {typeof(TArgument).FullName}, but got {args[0].GetType().FullName}."
+                );
+            }
+        }
+    }
+}
--- a/Intersect (Core)/Threading/Threaded`1.cs
+++ b/Intersect (Core)/Threading/Threaded`1.cs
@@ -12,10 +12,8 @@ namespace Intersect.Threading
         {
         }
 
-        /// <inheritdoc />
-        protected override void ThreadStart(params object[] args)
+        private static void CheckArgumentTypes(object[] args)
         {
-            // TODO: Generic utility that checks arg types against expected types
             if (args == null || args.Length < 1)
             {
                 throw new ArgumentOutOfRangeException(
@@ -23,7 +21,19 @@ namespace Intersect.Threading
                 );
             }
 
-            ThreadStart((TArgument) args[0]);
+            if (!(args[0] is TArgument))
+            {
+                throw new ArgumentException(
+                    $"Expected argument of type {typeof(TArgument).FullName}, but got {args[0].GetType().FullName}."
+                );
+            }
+        }
+
+        /// <inheritdoc />
+        protected override void ThreadStart(params object[] args)
+        {
+            CheckArgumentTypes(args);
+            ThreadStart((TArgument)args[0]);
         }
 
         protected abstract void ThreadStart(TArgument argument);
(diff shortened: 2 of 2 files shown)


lodicolo (changes requested):
This is not introducing a generic utility (would be an entirely separate class) that checks arguments.

Intersect (Core)/Threading/ArgumentTypeChecker.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+
+namespace Intersect.Threading
+{
+    /// <summary>
+    /// Utility class for checking the type of arguments passed to a new thread.
+    /// </summary>
+    /// <typeparam name="TArgument">Type of the argument that passes to a new thread.</typeparam>
+    public static class ArgumentTypeChecker<TArgument>
```
lodicolo, line 9: The type shouldn't be on the class for this.

Intersect (Core)/Threading/ArgumentTypeChecker.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            if (!(args[0] is TArgument))
+            {
+                throw new ArgumentException(
+                    $"Expected argument of type {typeof(TArgument).FullName}, but got {args[0].GetType().FullName}."
+                );
+            }
+        }
+    }
```
lodicolo, line 32: I have two suggestions, one is significantly more work.

One is via generics, but that would need more than 1 method, 1 per X number of arguments.

The other one method where the consumer has to pass in an array of types.
- If either array is null, throw ArgumentNullException
- If the number of arguments is not exactly equal to the number of types, throw ArgumentException stating length mismatch, saying what the expected length is and what the actual length is
- Loop through all of the arguments matching them against the respective type for that argument index, if there is a type mismatch (using `is`, not `GetType()` just to be clear), throw stating what index it occurred at, what the expected type was, and what the argument type actually is

Intersect (Core)/Threading/Threaded`1.cs
```diff
@@ -1,33 +1,32 @@
...
+        /// </summary>
+        /// <param name="name">Name of new thread.</param>
         protected Threaded(string name = null) : base(name)
         {
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Overrides the <see cref="Threaded.ThreadStart"/> method.
+        /// </summary>
```
lodicolo, line 19: This comment change provides zero information to the user and removes all of the original information inherited. Revert to inheritdoc.

Intersect (Core)/Threading/Threaded`1.cs
```diff
@@ -1,33 +1,32 @@
-﻿using System;
-
-namespace Intersect.Threading
+﻿namespace Intersect.Threading
 {
-
-    /// <inheritdoc />
+    /// <summary>
+    /// Represents a generic base class for creating and managing new threads.
+    /// </summary>
```
lodicolo, line 5: This part of the documentation should be moved to the base class, and inheritdoc should remain here.

Intersect (Core)/Threading/Threaded`1.cs
```diff
@@ -1,33 +1,32 @@
...
+    /// <typeparam name="TArgument">Type of argument that will be passed to a new thread.</typeparam>
```
lodicolo, line 6: This line can stay, but with different wording. It should say
> The type of the startup argument that is expected to be passed into <cref to Thread.Start>.

Also please change TArgument to TStartupArgument.

(I may have gotten the cref wrong, but it should be pointing to `ThreadStart(params object[] args)`).

Intersect (Core)/Threading/Threaded`1.cs
lodicolo: If you want to document it, the documentation should be in the base class.

Intersect (Core)/Threading/Threaded`1.cs
```diff
@@ -1,33 +1,32 @@
...
-
-            ThreadStart((TArgument) args[0]);
+            ArgumentTypeChecker<TArgument>.CheckArgumentTypes(args);
+            ThreadStart((TArgument)args[0]);
         }
 
+        /// <summary>
+        /// Main entry point for a new thread, where the argument of type <typeparamref name="TArgument"/> is passed.
+        /// </summary>
+        /// <param name="argument">Argument of type <typeparamref name="TArgument"/> that will be passed to a new thread.</param>
```
lodicolo, line 29: This does not provide any meaningful information to the reader, it is just restating what the code clearly does.

Intersect (Core)/Threading/Threaded`1.cs
```diff
@@ -1,33 +1,32 @@
...
-            }
-
-            ThreadStart((TArgument) args[0]);
+            ArgumentTypeChecker<TArgument>.CheckArgumentTypes(args);
+            ThreadStart((TArgument)args[0]);
         }
 
+        /// <summary>
+        /// Main entry point for a new thread, where the argument of type <typeparamref name="TArgument"/> is passed.
+        /// </summary>
```
lodicolo, line 28: This summary does not really explain what the method is for.

> Entry point for a threaded system that expects a single startup argument of type <typeparamref name="TArgument" /> and is called by <cref pointing to the params object[] args version of the method>.

lodicolo (changes requested):
Quite frankly my original implementation is bad and the arguments are actually better checked in Threaded.cs before ever actually calling Thread.Start().

PR #1711, opened 2023-01-24
enhancement: TabControl.cs TODO

Arufonsu:
- This uses margin.Left instead for scrolling. Also, it is using Util.Approach function for animation which is frame rate independent.
- Solves a TODO


lodicolo:
@Arufonsu I need a video clip of what this looks like before and after

Arufonsu:
> @Arufonsu I need a video clip of what this looks like before and after

Will do as soon i'm back home 👍🏽

PR #1719, opened 2023-01-31
enhancement: directly compare with TimeSystem conditions

WeylonSantana:
Before it was necessary to define a variable for the value "Time system" so that we could compare with our variables that use the time system, this modification allows to compare directly.

https://user-images.githubusercontent.com/63019821/215845980-fc58147b-0e5a-474a-a2d1-b097e22af4f3.mp4

### DIFF
--- a/Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
+++ b/Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
@@ -700,6 +700,11 @@ namespace Intersect.Editor.Forms.Editors.Events.Event_Commands
             UpdateNumericVariableElements();
         }
 
+        private void rdoTimeSystem_CheckedChanged(object sender, EventArgs e)
+        {
+            UpdateNumericVariableElements();
+        }
+
         private void UpdateNumericVariableElements()
         {
             nudVariableValue.Enabled = rdoVarCompareStaticValue.Checked;
@@ -787,7 +792,7 @@ namespace Intersect.Editor.Forms.Editors.Events.Event_Commands
             }
         }
 
-        private void TryLoadVariableBooleanComparison(VariableCompaison comparison)
+        private void TryLoadVariableBooleanComparison(VariableComparison comparison)
         {
             if (!(comparison is BooleanVariableComparison booleanComparison))
             {
@@ -829,7 +834,7 @@ namespace Intersect.Editor.Forms.Editors.Events.Event_Commands
             }
         }
 
-        private void TryLoadVariableIntegerComparison(VariableCompaison comparison)
+        private void TryLoadVariableIntegerComparison(VariableComparison comparison)
         {
             if (!(comparison is IntegerVariableComparison integerComparison))
             {
@@ -866,6 +871,10 @@ namespace Intersect.Editor.Forms.Editors.Events.Event_Commands
                     cmbCompareUserVar.SelectedIndex = UserVariableBase.ListIndex(integerComparison.CompareVariableId);
                 }
             }
+            else if(integerComparison.TimeSystem == true)
+            {
+                rdoTimeSystem.Checked = true;
+            }
             else
             {
                 rdoVarCompareStaticValue.Checked = true;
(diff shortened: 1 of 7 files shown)


Arufonsu:
Could you please show how it looks like when you add/remove this variable comparison/condition (?) also, can you use the former english editor for this (?) ( despite the fact i know spanish, this is kinda confusing to me tbh ). 

Anyways, seeing that this adds up a new way to compare with the system time directly **but** this is indirectly a feature that already exists if certain steps are followed, i'll tag it as _minor feature addition or quality of life change_ (Enhancement).

lodicolo:
> Could you please show how it looks like when you add/remove this variable comparison/condition (?)

I would also like to see what this actually looks like when adding/removing the condition.

WeylonSantana:
Idk if you want this screenshoot precisely, but

before:
image

after:
image
image

lodicolo:
> (@lodicolo, @Arufonsu) Idk if you want this screenshoot precisely, but

Yes, and I am considering that we begin requiring screenshots and/or video clips to make doc writing easier.

Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
```diff
@@ -866,6 +871,10 @@ private void TryLoadVariableIntegerComparison(VariableCompaison comparison)
                     cmbCompareUserVar.SelectedIndex = UserVariableBase.ListIndex(integerComparison.CompareVariableId);
                 }
             }
+            else if(integerComparison.TimeSystem == true)
```
lodicolo, line 874: Drop ` == true`, this isn't a nullable

Intersect.Editor/Localization/Strings.cs
```diff
@@ -419,6 +419,11 @@ public static string GetVariableComparisonString(IntegerVariableComparison compa
             if (comparison.CompareVariableId == Guid.Empty)
             {
                 value = comparison.Value.ToString();
+
+                if(comparison.TimeSystem == true)
```
lodicolo, line 423: `if (`, and drop ` == true`

Intersect.Editor/Forms/Editors/Events/Event Commands/EventCommand_ConditionalBranch.cs
lodicolo: Also `if (` (add a space)

Intersect.Editor/Localization/Strings.cs
```diff
@@ -2661,6 +2666,8 @@ public partial struct EventConditionDesc
 
             public static LocalizedString UserVariable = @"{00}: {01} {02}";
 
+            public static LocalizedString SystemTime = @"System Time (ms)";
```
lodicolo, line 2669: Can you move this up to L2659 to sort it?

Intersect.Server/Entities/Events/Conditions.cs
```diff
@@ -614,6 +614,11 @@ Event instance
                     compValue = player.User.GetVariableValue(comparison.CompareVariableId);
                 }
             }
+            else if(comparison.TimeSystem == true)
```
lodicolo, line 617: `if (`, drop ` == true`

Intersect.Server/Entities/Events/Conditions.cs
```diff
@@ -614,6 +614,11 @@ Event instance
...
+            {
+                compValue = (long)(DateTime.UtcNow - new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc))
```
lodicolo, line 619: Is there any way we can just not have to subtract Jan 1 1970?

lodicolo (changes requested):

Intersect.Server/Entities/Events/Conditions.cs
WeylonSantana: https://dirask.com/posts/C-NET-get-current-time-in-milliseconds-gp5BRj

lodicolo (approved):

PR #1720, opened 2023-02-03
fix: bank deposit behavior

### LINKED ISSUE #1681: bug: single stackable item from inventory to the bank
### Description

Hi, it's my first time for report a bug and I'm not english (I'm sorry for my disaster language) ; 

I found this bug randomly : 

I don't know what's going on, I just want to move my stackable item (potion) to the same item in my bank and some of those same items disappeared suddently
Link (1) : https://youtu.be/knqO2je_GAU

And now, if entierly of my items stacked disapeared in the bank, the slot in my bank is broken and I can't use this slot anymore : 
Link Video (2) :  https://youtu.be/3SLylQVr8y0

### Steps to Reproduce

Step 1 : Drag a stackable item to the bank like a potion or other item (Stackable only) (First Video)
Step 2 : Drag the same item through the inventory into the same slot of the same item in the bank (First Video)
Step 3 : You have the bug and item disapear randomly or/and create randomly too a broken slot in the bank and make it unusable (Second Video)

### Version with bug

v0.7.1.088

Unknow

### Affected platforms

Windows

You just shouldn't drag an item from your inventory onto a bank slot with a stackable item, otherwise it would create this bug and probably broke your bank slot. 

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
### Fixes:
- Adjusts quantity and maxQuantity for stackable item's slider.
- Checks if item is stackable when required so we don't unnecessarily show up the deposit slider on single equipment items.
- Context menu bank deposit d/c fix: when destination slot is < 0 (context menu deposit) -> FindsOpenSlot and deposit.
- Non-stackable items fix: we weren't checking stackable items properly, which translated into non-stackable items being placed over each other, leading to glitched bank slots.
- Single stackable items (ie, an arrow) won't overwrite existing item(s) in slot anymore and they stack along with the rest of the same stackable items.

### More fixes along with improvements:
 - We can now easily fill the bank when depositing stackable items, just drag the slider to the right! the server will make sure to adjust the amount and fill the required slots.
 - If the bank slot that an item is dropped on (slot is used as "hint") does not stack or is full of stackable items but there is still available slots in the bank, the server makes sure to find a new slot for non stackable items and fills the non-fully-stacked items.
 - If the bank slot used as "hint" is empty, the server fills the non-fully-stacked items and deposits the rest within the "hint" slot.
 - If there's no initial "hint" (ie, context menu deposit), then finds an open slot to deposit the rest of the stackable items.
 - Updates and simplifies the "CanStore" method to properly work along with this commit changes.
 - Creates the "FindAvailableSpaceForStackableItem" method which finds the amount of available space in the bank for a stackable item.
 - Added in-game system messages to be sent to players when successfully depositing and withdrawing items from the bank.

### Previews

~
(April 15th, 2023)

https://user-images.githubusercontent.com/17498701/232185444-0c5532f0-4cf6-4aa9-80af-fc331f4f9fe6.mp4

~

https://user-images.githubusercontent.com/17498701/231619732-28ca3eca-e3a2-4520-b7b0-a137d926ede7.mp4

~

https://user-images.githubusercontent.com/17498701/231619932-64a53f73-4fc6-4637-8161-4da5d7079e12.mp4

~

https://user-images.githubusercontent.com/17498701/231617217-dd9c7808-d68d-4f12-9525-3600e9001d10.mp4

### DIFF
--- a/Intersect.Server/Entities/BankInterface.cs
+++ b/Intersect.Server/Entities/BankInterface.cs
@@ -291,16 +291,18 @@ namespace Intersect.Server.Entities
                 return amount;
             }
 
+            int maxBankStack = ItemBase.Get(itemId).MaxBankStack;
+
             for (int i = 0; i < mMaxSlots; i++)
             {
-                if (mBank[i]?.ItemId == itemId)
+                var bankSlot = mBank[i];
+                if (bankSlot != null && bankSlot.ItemId == itemId)
                 {
-                    int quantityLeft = mBank[i].Descriptor.MaxBankStack - mBank[i].Quantity;
-                    spaceLeft += quantityLeft;
+                    spaceLeft += maxBankStack - bankSlot.Quantity;
                 }
-                else if (mBank[i]?.ItemId == Guid.Empty)
+                else if (bankSlot == null || bankSlot.ItemId == Guid.Empty)
                 {
-                    spaceLeft += ItemBase.Get(itemId).MaxBankStack;
+                    spaceLeft += maxBankStack;
                 }
             }
 
--- a/Intersect.Server/Entities/BankInterface.cs
+++ b/Intersect.Server/Entities/BankInterface.cs
@@ -194,55 +194,26 @@ namespace Intersect.Server.Entities
         /// </summary>
         /// <param name="item">The <see cref="Item"/> to check for.</param>
         /// <returns>Returns whether or not the bank instance can store this item.</returns>
-        public bool CanStoreItem(Item item)
+        private bool CanStoreItem(Item item)
         {
-            if (item.Descriptor != null)
+            // If this item isn't stackable we just need to find an open slot.
+            if (!item.Descriptor.Stackable)
             {
-                // Is the item stackable?
-                if (item.Descriptor.IsStackable)
-                {
-                    // Does the bank have this item already?
-                    var slot = FindItemSlot(item.ItemId);
-                    if (slot >= 0)
-                    {
-                        var existingSlot = mBank[slot];
+                return FindOpenSlot() != -1;
+            }
 
-                        // Can we blindly add more to this stack?
-                        var untilFull = item.Descriptor.MaxBankStack - existingSlot.Quantity;
-                        if (untilFull >= item.Quantity)
-                        {
-                            return true;
-                        }
-
-                        // Check to see if we have the inventory spaces required to hand these items out AFTER filling up the existing slot!
-                        var toGive = item.Quantity - untilFull;
-                        if (Math.Ceiling((double)toGive / item.Descriptor.MaxBankStack) <= FindOpenSlots().Count)
-                        {
-                            return true;
-                        }
-                    }
-                    // User doesn't have this item yet.
-                    else
-                    {
-                        // Does the user have enough free space for these stacks?
-                        if (Math.Ceiling((double)item.Quantity / item.Descriptor.MaxBankStack) <= FindOpenSlots().Count)
-                        {
-                            return true;
-                        }
-                    }
-                }
-                // Not a stacking item, so can we contain the amount we want to give them?
-                else
+            // If it is stackable we need to find an open slot or an existing slot that contains this item type.
+            for (var i = 0; i < mMaxSlots; i++)
+            {
+                var bankSlot = mBank[i];
+                if (bankSlot.ItemId == item.ItemId &&
+                    bankSlot.Quantity + item.Quantity <= bankSlot.Descriptor.MaxBankStack)
                 {
-                    if (FindOpenSlots().Count >= item.Quantity)
-                    {
-                        return true;
-                    }
+                    return true;
                 }
             }
 
-            // Nothing matches in here, give up!
-            return false;
+            return FindOpenSlot() != -1;
         }
 
         /// <summary>
@@ -299,6 +270,36 @@ namespace Intersect.Server.Entities
         }
 
         /// <summary>
+        /// Finds the amount of available space in the bank for a stackable item.
+        /// This method calculates the space available in the bank by checking
+        /// the number of slots that have the same item ID as the provided item, and
+        /// adding up the remaining quantity space in each of those slots. It also
+        /// checks for any empty slots in the bank, and adds the maximum stack size
+        /// of the provided item to the available space for each empty slot found.
+        /// </summary>
+        /// <param name="itemId">The ID of the stackable item to find available space for.</param>
+        /// <param name="amount">The requested amount of the item to deposit.</param>
+        /// <returns>The maximum amount of the stackable item that can be deposited in the bank.</returns>
+        private int FindAvailableSpaceForStackableItem(Guid itemId, int amount)
+        {
+            int spaceLeft = 0;
+            for (int i = 0; i < mMaxSlots; i++)
+            {
+                if (mBank[i]?.ItemId == itemId)
+                {
+                    int quantityLeft = mBank[i].Descriptor.MaxBankStack - mBank[i].Quantity;
+                    spaceLeft += quantityLeft;
+                }
+                else if (mBank[i]?.ItemId == Guid.Empty)
+                {
+                    spaceLeft += ItemBase.Get(itemId).MaxBankStack;
+                }
+            }
+
+            return Math.Min(amount, spaceLeft);
+        }
+
+        /// <summary>
         /// Checks whether or not this bank instance has enough items to accept a take operation.
         /// </summary>
         /// <param name="itemId">The ItemId to see if it can be taken away.</param>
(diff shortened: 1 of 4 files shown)


lodicolo:
@Arufonsu Both of the videos look like the broken behavior, where is a video of the correct behavior with this fix?

lodicolo:
From Discord:
> If the bank slot does not stack but there is space in the bank still, the move should not fail
> the slot that an item is dropped on should be a "hint" not an absolute "this is where it goes"

That should be addressed too in this PR.

Arufonsu:
Updated PR, code and video previews, so far this manages to fix the issues and behaves as expected, discussed and suggested 👍🏽

Arufonsu:
Updated PR (_again_, sorry for the bunch of force pushes, won't happen after the first review - unless is required).
Fixed some logic issues i've encountered, also added some nice additions, preview:

https://user-images.githubusercontent.com/17498701/232185444-0c5532f0-4cf6-4aa9-80af-fc331f4f9fe6.mp4

Hopefully this yields the expected behavior, PR feels ready for review as well 👍🏽

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -194,55 +194,26 @@ public List<int> FindItemSlots(Guid itemId, int quantity = 1)
         /// </summary>
         /// <param name="item">The <see cref="Item"/> to check for.</param>
         /// <returns>Returns whether or not the bank instance can store this item.</returns>
-        public bool CanStoreItem(Item item)
+        private bool CanStoreItem(Item item)
```
lodicolo, line 197: Don't change this to private.

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -194,55 +194,26 @@ public List<int> FindItemSlots(Guid itemId, int quantity = 1)
...
         {
-            if (item.Descriptor != null)
+            // If this item isn't stackable we just need to find an open slot.
+            if (!item.Descriptor.Stackable)
```
lodicolo, line 200: Why did you remove the null check...?

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -194,55 +194,26 @@ public List<int> FindItemSlots(Guid itemId, int quantity = 1)
...
+                var bankSlot = mBank[i];
+                if (bankSlot.ItemId == item.ItemId &&
+                    bankSlot.Quantity + item.Quantity <= bankSlot.Descriptor.MaxBankStack)
                 {
-                    if (FindOpenSlots().Count >= item.Quantity)
-                    {
-                        return true;
-                    }
+                    return true;
                 }
```
lodicolo, line 213: You removed partial stack filling? If you are trying to put in 100 stackable items, the limit per stack is 100, and you have no totally empty slots but you have two slots that only have 50 which it would fit in, your code rejects it.

lodicolo (changes requested):

Intersect.Server/Entities/BankInterface.cs
Arufonsu: you are right, this happens if there's no slot left that's empty but for example, two slots that aren't completly filled, i've just added the following after the for loop and it seems to handle things well now:

imagen

preview:
https://user-images.githubusercontent.com/17498701/233818612-3a07865c-5033-4316-a31e-2e1f7b47d7dc.mp4

Intersect.Server/Entities/BankInterface.cs
Arufonsu: uuhh ;whoops; ( restored )

Intersect.Server/Entities/BankInterface.cs
Arufonsu: Also: with this, we don't really need the for loop that's before this one condition anymore.

Intersect.Server/Entities/BankInterface.cs
Arufonsu: restored

Arufonsu:
Self reminder: 
- While playing around with this few weeks ago, i found out a way to also automatically adjust the max quantity sliders in relation with the max stackable items either in bank or in inventory.
- Were we able to drag in many non stackable items with the slider before? if so, i should restore this lol

Arufonsu:
self-reminder: restored the ability to deposit multiple non-stackable items

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/f7ab5f5b-f12b-4d52-9231-1d4673736284

PR is ready for review and being built for ease of testing ~

Arufonsu:
Introduces the 'FindAvailableBankSpaceForItem' method to calculate the total available space within the bank for specific items. This enhancement streamlines the bank deposit process by automatically determining the appropriate values for the 'quantity' and 'maxQuantity' parameters in the deposit slider, improving user experience and accuracy when depositing items in the bank.


https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/a55a39e5-6a4e-4f48-9de2-a943ff874059

Arufonsu:
client enhancement: NoSpace bank system message

- Instead of blindly opening the item deposit slider when there's no space left in the bank, we now inform the player "There is no space left in your bank for that item!" before the slider is up with a max quantity of 0 (which is pointless to do).

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/fcae93ee-9a2a-435f-9805-3aaae51dd66a

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -298,6 +271,36 @@ public int FindItemQuantity(Guid itemId)
...
+        /// </summary>
+        /// <param name="itemId">The ID of the item to find available space for.</param>
+        /// <param name="amount">The requested amount of the item to deposit.</param>
+        /// <returns>The maximum amount of the item that can be deposited in the bank.</returns>
+        private int FindAvailableSpaceForItem(Guid itemId, int amount)
+        {
+            int spaceLeft = 0;
+            for (int i = 0; i < mMaxSlots; i++)
+            {
+                if (mBank[i]?.ItemId == itemId)
```
lodicolo, line 290: Before the if statement add this as a short-circuit to skip costly searches

```
if (spaceLeft >= amount)
{
    return amount;
}
```

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -196,53 +196,26 @@ public List<int> FindItemSlots(Guid itemId, int quantity = 1)
...
-                        if (Math.Ceiling((double)item.Quantity / item.Descriptor.MaxBankStack) <= FindOpenSlots().Count)
-                        {
-                            return true;
-                        }
-                    }
-                }
-                // Not a stacking item, so can we contain the amount we want to give them?
-                else
-                {
-                    if (FindOpenSlots().Count >= item.Quantity)
```
lodicolo, line 237: This case appears to not be covered in the new code?

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -319,79 +322,106 @@ public bool TryDepositItem(int slot, int amount, int destSlot = -1, bool sendUpd
...
+                PacketSender.SendChatMsg(mPlayer, Strings.Items.nobank, ChatMessageType.Bank, CustomColors.Items.Bound);
+                return false;
+            }
 
-                        if (CanStoreItem(new Item(itemBase.Id, amount), destSlot))
+            // Calculate the maximum amount that can be deposited, based on quantity and bank space.
+            var depositMultipleItems = mPlayer.FindInventoryItemQuantity(inventoryItemItemId) > 1;
+            var maxAmount = depositMultipleItems ? Math.Min(mPlayer.FindInventoryItemQuantity(inventoryItemItemId),
+                    FindAvailableSpaceForItem(inventoryItemItemId, amount)) : 1;
+            amount = Math.Min(amount, maxAmount);
```
lodicolo, line 345: `depositMultipleItems` should be based on this line instead?

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -319,79 +322,106 @@ public bool TryDepositItem(int slot, int amount, int destSlot = -1, bool sendUpd
...
+            {
+                PacketSender.SendChatMsg(mPlayer, Strings.Items.nobank, ChatMessageType.Bank, CustomColors.Items.Bound);
+                return false;
+            }
 
-                        if (CanStoreItem(new Item(itemBase.Id, amount), destSlot))
+            // Calculate the maximum amount that can be deposited, based on quantity and bank space.
+            var depositMultipleItems = mPlayer.FindInventoryItemQuantity(inventoryItemItemId) > 1;
+            var maxAmount = depositMultipleItems ? Math.Min(mPlayer.FindInventoryItemQuantity(inventoryItemItemId),
+                    FindAvailableSpaceForItem(inventoryItemItemId, amount)) : 1;
```
lodicolo, line 344: `mPlayer.FindInventoryItemQuantity(inventoryItemItemId)` is repeated in both the condition and the branch, the branch is totally unnecessary here.

lodicolo: This could be changed to
```csharp
var inventoryQuantity = mPlayer.FindInventoryItemQuantity(inventoryItemItemId);
amount = Math.Min(amount, inventoryQuantity); // This is to optimize the next line
var availableStorageSpace = FindAvailableSpaceForItem(inventoryItemItemId, amount));
var maxAmount = Math.Min(inventoryQuantity, availableStorageSpace);
```

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -319,79 +322,106 @@ public bool TryDepositItem(int slot, int amount, int destSlot = -1, bool sendUpd
...
+                    destSlot = FindOpenSlot();
+                }
+
+                if (destSlot < 0)
+                {
+                    PacketSender.SendChatMsg(mPlayer, Strings.Banks.banknospace, ChatMessageType.Bank,
+                        CustomColors.Alerts.Error);
+                    return false;
                 }
             }
```
lodicolo, line 385: I'm wondering if generic and simpler logic is better than introducing an "if 1" condition

Intersect.Server/Entities/BankInterface.cs
```diff
@@ -319,79 +322,106 @@ public bool TryDepositItem(int slot, int amount, int destSlot = -1, bool sendUpd
...
+            }
 
-                        if (CanStoreItem(new Item(itemBase.Id, amount), destSlot))
+            // Calculate the maximum amount that can be deposited, based on quantity and bank space.
+            var depositMultipleItems = mPlayer.FindInventoryItemQuantity(inventoryItemItemId) > 1;
+            var maxAmount = depositMultipleItems ? Math.Min(mPlayer.FindInventoryItemQuantity(inventoryItemItemId),
+                    FindAvailableSpaceForItem(inventoryItemItemId, amount)) : 1;
+            amount = Math.Min(amount, maxAmount);
+
+            if (amount <= 0)
```
lodicolo, line 347: amount < 1

lodicolo (changes requested):
This isn't an exhaustive review but there is a missing thing and a few over-complicated things that I think can be simplified

Intersect.Server/Entities/BankInterface.cs
Arufonsu: This should be covering it when it comes to non stackable items:

https://github.com/AscensionGameDev/Intersect-Engine/blob/0765628603c5e32890fc7757e6dc79f2394fa100/Intersect.Server/Entities/BankInterface.cs#L205-L209

Intersect.Server/Entities/BankInterface.cs
Arufonsu: maxAmount < 1 after the suggested changes (?)

lodicolo: No, it would still be `amount < 1` because you would not be getting rid of the last `amount = Math.Min(amount, maxAmount);`

Intersect.Server/Entities/BankInterface.cs
lodicolo: Yeah but the code specifically said `>= item.Quantity` not `1` -- it makes me think there's a case where something can be not stackable with a quantity > 1...

Arufonsu: that would be weird (?) but just to make sure, i can add this check back, although i would have to see how it behaves with this... which tbh feels weird, right now we are just looking away for an available slot whenever it comes to stackables, since they aren't supposed to ... stack together in a single a slot (?)

PR #1744, opened 2023-03-07
fix: (Day) Fix serialization issues with typewriter options

AlexVild:


lodicolo (changes requested):
Instead of casting, try changing LoadPreference to
```csharp
        public T LoadPreference<T>(string key, T defaultValue)
        {
            var value = LoadPreference(key);
            if (string.IsNullOrEmpty(value))
            {
                return defaultValue;
            }

            var type = typeof(T);
            if (type.IsEnum)
            {
                try
                {
                    var enumValue = Enum.Parse(type, value);
                    return (T)enumValue;
                }
                catch
                {
                    return defaultValue;
                }
            }

            return (T)Convert.ChangeType(value, typeof(T));
        }
```

lodicolo (approved):

PR #1747, opened 2023-03-08
chore: correct enum names

lodicolo:
I need someone to compile and test the editor (already tested client/server). I don't think it is broken, but I don't trust the IDE while it can't compile it.

With the exception of the "split into multiple files" commit, each commit had two actions performed:
- renamed the enum
- in the enum file and only the enum file I formatted the file


Intersect (Core)/Enums/VariableType.cs
```diff
@@ -0,0 +1,19 @@
+namespace Intersect.Enums
```
Arufonsu, line 1: 👍🏽  !

PR #1755, opened 2023-03-08
fix: don't auto-turn when self targeting

### LINKED ISSUE #1754: bug: self target turns player to the right
### Description

The feature auto-turn to targets is making players to turn over to the right whenever they target themselves.

### Steps to Reproduce

1. Enable auto-turn to target in both: server configuration and in game settings.
1. As seen in the following video, simply target yourself to trigger the unwanted behavior:

https://user-images.githubusercontent.com/17498701/223771636-f91fd004-053d-4821-a0a1-5d757e71def6.mp4

### Version with bug

0.7.1.114

never did, this situation wasn't considered uppon the feat release.

### Affected platforms

Windows

Yes, i'll open a PR with the fix:

```csharp
            if (en == this)
            {
                return;
            }
```

being "this" the player and "en" the actual target in the scope of the method.

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
- fixes an issue where auto-turn to target made the player turn over to the right whenever they target themselves.
- should resolve #1754 

### Preview:

https://user-images.githubusercontent.com/17498701/223774184-0aa71dde-174e-4478-adaf-486c11096eb9.mp4

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -1454,8 +1454,22 @@ namespace Intersect.Client.Entities
 
         private void AutoTurnToTarget(Entity en)
         {
-            if (!Globals.Database.AutoTurnToTarget || !Options.Instance.PlayerOpts.EnableAutoTurnToTarget ||
-                Controls.KeyDown(Control.TurnAround))
+            if (!Options.Instance.PlayerOpts.EnableAutoTurnToTarget)
+            {
+                return;
+            }
+
+            if (!Globals.Database.AutoTurnToTarget)
+            {
+                return;
+            }
+
+            if (Controls.KeyDown(Control.TurnAround))
+            {
+                return;
+            }
+
+            if (en == this)
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -1454,8 +1454,22 @@ private void SetTargetBox(Entity en)
...
+            {
+                return;
+            }
+
+            if (Controls.KeyDown(Control.TurnAround))
+            {
+                return;
+            }
+
+            if (en == this)
```
lodicolo, line 1472: Re-order these checks:

1. `en`, because it's the closest value to check since it's a parameter
2. `Globals.Database.AutoTurnToTarget` (1-2 levels of pointer indirection depending on how the compiler handles statics like Database`)
3. `Options.Instance.PlayerOpts.EnableAutoTurnToTarget` (2-3 levels of pointer indirection)
4. `Controls.KeyDown(Control.TurnAround)` (this is a method call, on top of pointer indirection)

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: Done 👍🏽

lodicolo (approved):

PR #1773, opened 2023-03-20
chore: MovesTo method with map attributes

### LINKED ISSUE #1100: chore: refactor Entity#CanMove()'s return types

Arufonsu:
- Renames "CanMove" to "MovesTo" in order to make more sense (the method wasn't a boolean).
- Adds additional values to MapAttribute (OutOfBounds, Player, Event) and sorts them all alphabetically.
- MovesTo now returns updated MapAttribute values from the core instead of confusing integers, which allows us to have more readable and simplified code.
- frmMapLayers.cs: removes Method 'GetAttributeFromEditor' which contained unnecessary int casts to map attributes plus, is never used.
- Updates Dash's CalculateRange with the new values and a simplified switch.
- Should resolve #1100

### DIFF
--- a/Intersect (Core)/Enums/MapAttribute.cs
+++ b/Intersect (Core)/Enums/MapAttribute.cs
@@ -1,29 +1,38 @@
 namespace Intersect.Enums
 {
-    public enum MapAttribute : byte
+    /// <summary>
+    /// This enumeration type represents the attributes and/or entities within in-game tiles.
+    /// </summary>
+    public enum MapAttribute
     {
-        Walkable = 0,
+        Animation,
 
         Blocked,
 
-        Item,
+        Critter,
 
-        ZDimension,
-
-        NpcAvoid,
-
-        Warp,
-
-        Sound,
-
-        Resource,
-
-        Animation,
+        Event,
 
         GrappleStone,
 
+        Item,
+
+        NpcAvoid,
+
+        OutOfBounds,
+
+        Player,
+
+        Resource,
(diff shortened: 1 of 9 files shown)


Intersect (Core)/Enums/MapAttribute.cs
lodicolo, line 1: This change corrupts all existing maps :slightly_smiling_face:

lodicolo (changes requested):
This PR is sufficiently flawed that I am actually just going to close it.

You are welcome to open a new PR that actually cleans up code, but corrupting all maps and introducing invalid attributes is totally off the table.

(A player is not an attribute you can set on a tile? Why on earth was this even included?)

Intersect (Core)/Enums/MapAttribute.cs
```diff
@@ -1,29 +1,38 @@
...
 
-        Item,
+        Critter,
+
+        Event,
+
+        GrappleStone,
 
-        ZDimension,
+        Item,
```
AlexVild, line 18: You can't change the order of an enum like this, as by default each enumerated value is assigned a numeric value based off of its order within the enum, i.e `(int)Walkable == 0`, `(int)Blocked == 1, etc`.

If you really wanted to sort these, there needs to be a corresponding migration that re-maps attributes to their new values for any existing map

PR #1774, opened 2023-03-23
chore: inheritance between players and NPCs

lucas-tardivo:
- Applied some SOLID patterns to remove casting between Player -> Npc


Intersect.Server/Entities/Entity.cs
```diff
@@ -2724,8 +2724,17 @@ public bool InRangeOf(Entity target, int range)
             return false;
         }
 
-        public virtual void NotifySwarm(Entity attacker)
+        public virtual void NotifySwarm(Entity attacker) 
         {
+            if (MapController.TryGetInstanceFromMap(MapId, MapInstanceId, out var instance))
```
lodicolo, line 2729: To avoid nesting, return if false

Intersect.Server/Entities/Entity.cs
```diff
@@ -2724,8 +2724,17 @@ public bool InRangeOf(Entity target, int range)
...
+            {
+                instance.GetEntities(true).ForEach(
```
lodicolo, line 2731: for or foreach loop (not .ForEach()) is preferred because it doesn't create a closure.

Intersect.Server/Entities/Npc.cs
```diff
@@ -1633,6 +1633,24 @@ public override EntityPacket EntityPacket(EntityPacket packet = null, Player for
...
+            {
+                RemoveFromDamageMap(en);
+            }
+        }
+
+        public override void OnNearEntityAttacked(Entity victim, Entity attacker) 
+        {
+            if (Target == null &&
+                IsAllyOf(victim) &&
+                InRangeOf(this, Base.SightRange) &&
```
lodicolo, line 1648: This InRangeOf call seems wrong.

Before, it was calling InRangeOf on the Player, and passing in the NPC. Now it's InRangeOf on the NPC, and passing in the NPC? The NPC will always be in its own sight range though.

Intersect.Server/Entities/Npc.cs
```diff
@@ -1633,6 +1633,24 @@ public override EntityPacket EntityPacket(EntityPacket packet = null, Player for
...
+                victim.GetEntityType() == EntityType.Player)
```
lodicolo, line 1649: Also to avoid line breaks in the middle of an if statement, invert the conditions and return

lodicolo (changes requested):

Intersect.Server/Entities/Entity.cs
```diff
@@ -2724,8 +2724,19 @@ public bool InRangeOf(Entity target, int range)
             return false;
         }
 
-        public virtual void NotifySwarm(Entity attacker)
+        public virtual void NotifySwarm(Entity attacker)
```
Arufonsu, line 2727: get rid of the whitespace that's by the end of the line here 👀 (unnecesary)   ~

Arufonsu (changes requested):

Intersect.Server/Entities/Npc.cs
```diff
@@ -1633,6 +1633,38 @@ public override EntityPacket EntityPacket(EntityPacket packet = null, Player for
...
+
+            if (!InRangeOf(attacker, Base.SightRange))
+            {
+                return;
+            }
+
+            if (victim.GetEntityType() != EntityType.Player)
+            {
+                return;
+            }
```
lodicolo, line 1664: This logic does not make sense.

You refactored Player.NotifySwarm(), but you didn't refactor the Npc.NotifySwarm()? They are essentially the same, the difference being what qualifies as an "ally".

I also think some of the current code is wrong.

Intersect.Server/Entities/Entity.cs
```diff
@@ -3037,6 +3053,10 @@ public StatusPacket[] StatusPackets()
             return statusPackets;
         }
 
+        public virtual void OnNearEntityDies(Entity en) { }
```
lodicolo, line 3056: Rename to `OnNearbyEntityDeath`, also fully spell out `entity`.

Intersect.Server/Entities/Entity.cs
```diff
@@ -3037,6 +3053,10 @@ public StatusPacket[] StatusPackets()
...
+
+        public virtual void OnNearEntityAttacked(Entity victim, Entity attacker)  { }
```
lodicolo, line 3058: Rename to `OnSwarmAttacked`. Parameters can be left in for mergeability for those who need the extra information.

Intersect.Server/Entities/Npc.cs
lodicolo: The `Target` check stays here.

I suggest refactoring out the `IsAllyOf` and `InRangeOf` checks into the base class.

The entity type check is already part of Npc's `IsAllyOf` method, and so this is redundant.

Intersect.Server/Entities/Entity.cs
```diff
@@ -2726,6 +2726,17 @@ public bool InRangeOf(Entity target, int range)
...
+            {
+                return;
+            }
+
+            var entities = instance.GetEntities(true);
+
+            foreach (var entity in entities)
+            {
+                entity.OnNearEntityAttacked(this, attacker);
+            }
```
lodicolo, line 2739: Add to the for-loop before the OnSwarmAttacked call:

```csharp
            if (!CanSwarm)
            {
                continue;
            }
            
            if (!IsAllyOf(victim))
            {
                continue;
            }

            if (!InRangeOf(attacker, SightRange))
            {
                continue;
            }
```

lodicolo: This requires that
- we add a `CanSwarm` virtual property to Entity, and the default value that it returns should be `false`
  - `Npc` needs an override that returns `Base.Swarm`
  - `Npc`'s `IsAllyOf` should have the `case Npc otherNpc` arm changed from `return Base == otherNpc.Base` to `return Base.Swarm && Base == otherNpc.Base` because if an NPC can't swarm then it really isn't an ally per the current code in `Npc`'s override for `NotifySwarm`
  - `Npc`'s override for `NotifySwarm` swarm should be deleted because all of its logic is actually part of the base `NotifySwarm` and `IsAllyOf` methods
- we add a `SightRange` virtual property to `Entity,` and the default value that it returns should be `999` (1K - 1, there's another special value at `9999` [10K-1] already)

lodicolo (changes requested):

PR #1775, opened 2023-03-24
chore: CanMoveInDirection method

### LINKED ISSUE #1100: chore: refactor Entity#CanMove()'s return types

Arufonsu:
* Should resolve #1100

[chore: CanMoveInDirection method](https://github.com/AscensionGameDev/Intersect-Engine/pull/1775/commits/fd56dc2ca0be09747c48db41d95960c4c83a7abd)
* Renames "CanMove" to "CanMoveInDirection", which is now a boolean.
* Creates and uses enum type "MovementBlockerType".
* CanMoveInDirection takes in the direction and two out parameters of the MovementBlockerType and EntityType enum types. The method checks if the entity can move in the given direction by analyzing the MapController, MapAttributes, and nearby entities. If there is a blockage, the method sets the out parameters with the corresponding MovementBlockerType and EntityType, and returns false. Otherwise, it returns true.
* Updates Dash's CalculateRange with the new method and a simplified switch.

[review: 28.03.2023](https://github.com/AscensionGameDev/Intersect-Engine/pull/1775/commits/2acbb28862a64b156c627d38591ba35a291ae4b1)
- Renames "IsTileWalkable" -> "GetTileBlocker" which now returns Movement and Entity blocker types.
- (Editor) slideAttribute.Direction is a Direction value instead of byte now (tested and working).
- Restores equivalent checks commented in the review.
- Less code nesting.

### DIFF
--- a/Intersect.Server/Entities/Combat/Dash.cs
+++ b/Intersect.Server/Entities/Combat/Dash.cs
@@ -81,8 +81,14 @@ namespace Intersect.Server.Entities.Combat
                         case MovementBlockerType.Entity:
                             switch (entityType)
                             {
-                                case EntityType.Resource when activeResourcePass == false:
-                                case EntityType.Resource when deadResourcePass == false:
+                                case EntityType.Resource:
+                                    if (activeResourcePass || deadResourcePass)
+                                    {
+                                        break;
+                                    }
+
+                                    return;
+
                                 case EntityType.Event:
                                 case EntityType.Player:
                                 case EntityType.GlobalEntity:
@@ -90,8 +96,7 @@ namespace Intersect.Server.Entities.Combat
                                 case EntityType.Projectile:
                                     break;
                                 default:
-                                    throw new InvalidOperationException(
-                                        "Unreachable"); // TODO: Change to \UnreachableException` in .NET 7`
+                                    throw new InvalidOperationException("Unreachable"); // TODO: Change to UnreachableException in .NET 7
                             }
 
                             break;
@@ -101,9 +106,7 @@ namespace Intersect.Server.Entities.Combat
                             break;
 
                         default:
-                            throw
-                                new InvalidOperationException(
-                                    "Unreachable"); // TODO: Change to \UnreachableException` in .NET 7`
+                            throw new InvalidOperationException("Unreachable"); // TODO: Change to UnreachableException in .NET 7
                     }
                 }
 
--- a/Intersect.Server/Entities/Combat/Dash.cs
+++ b/Intersect.Server/Entities/Combat/Dash.cs
@@ -82,12 +82,12 @@ namespace Intersect.Server.Entities.Combat
                             switch (entityType)
                             {
                                 case EntityType.Resource:
-                                    if (activeResourcePass || deadResourcePass)
+                                    if (!activeResourcePass || !deadResourcePass)
                                     {
-                                        break;
+                                        return;
                                     }
 
-                                    return;
+                                    break;
 
                                 case EntityType.Event:
                                 case EntityType.Player:
--- a/Intersect.Server/Entities/Combat/Dash.cs
+++ b/Intersect.Server/Entities/Combat/Dash.cs
@@ -1,4 +1,5 @@
-﻿using Intersect.Enums;
+﻿using System;
+using Intersect.Enums;
 using Intersect.Server.Networking;
 using Intersect.Utilities;
 
@@ -50,40 +51,60 @@ namespace Intersect.Server.Entities.Combat
             bool zdimensionPass = false
         )
         {
-            var n = 0;
             en.MoveTimer = 0;
             Range = 0;
             for (var i = 1; i <= range; i++)
             {
-                n = en.CanMove(Direction);
-                if (n == -5) //Check for out of bounds
+                if (!en.CanMoveInDirection(Direction, out var blockerType, out var entityType))
                 {
-                    return;
-                } //Check for blocks
+                    switch (blockerType)
+                    {
+                        case MovementBlockerType.OutOfBounds:
+                            return;
 
-                if (n == -2 && blockPass == false)
-                {
-                    return;
-                } //Check for ZDimensionTiles
+                        case MovementBlockerType.MapAttribute:
+                            if (blockPass)
+                            {
+                                break;
+                            }
 
-                if (n == -3 && zdimensionPass == false)
-                {
-                    return;
-                } //Check for active resources
+                            return;
 
-                if (n == (int) EntityType.Resource && activeResourcePass == false)
-                {
-                    return;
-                } //Check for dead resources
+                        case MovementBlockerType.ZDimension:
+                            if (zdimensionPass)
+                            {
+                                break;
+                            }
 
-                if (n == (int) EntityType.Resource && deadResourcePass == false)
-                {
-                    return;
-                } //Check for players and solid events
+                            return;
 
-                if (n == (int) EntityType.Player || n == (int) EntityType.Event)
-                {
-                    return;
+                        case MovementBlockerType.Entity:
+                            switch (entityType)
+                            {
+                                case EntityType.Resource when activeResourcePass == false:
+                                case EntityType.Resource when deadResourcePass == false:
+                                case EntityType.Event:
+                                case EntityType.Player:
+                                case EntityType.GlobalEntity:
+                                    return;
+                                case EntityType.Projectile:
+                                    break;
+                                default:
+                                    throw new InvalidOperationException(
+                                        "Unreachable"); // TODO: Change to \UnreachableException` in .NET 7`
+                            }
+
+                            break;
+
+                        case MovementBlockerType.NotBlocked:
+                        case MovementBlockerType.Slide:
+                            break;
+
+                        default:
+                            throw
+                                new InvalidOperationException(
+                                    "Unreachable"); // TODO: Change to \UnreachableException` in .NET 7`
+                    }
                 }
 
                 en.Move(Direction, null, true);
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -381,189 +381,154 @@ namespace Intersect.Server.Entities
             }
         }
 
-        //Movement
-        /// <summary>
-        ///     Determines if this entity can move in the direction given.
-        ///     Returns -5 if the tile is completely out of bounds.
-        ///     Returns -3 if a tile is blocked because of a Z dimension tile
-        ///     Returns -2 if a tile is blocked by a map attribute.
-        ///     Returns -1 for clear.
-        ///     Returns the type of entity that is blocking the way (if one exists)
-        /// </summary>
-        /// <param name="moveDir"></param>
-        /// <returns></returns>
-        public virtual int CanMove(Direction moveDir)
-        {
-            var xOffset = 0;
-            var yOffset = 0;
+        public virtual bool CanMoveInDirection(Direction direction) => CanMoveInDirection(direction, out _, out _);
 
+        /// <summary>
+        /// Determines if this entity can move in the given direction and returns what is blocking the path.
+        /// </summary>
+        /// <param name="direction">The direction to check for movement.</param>
+        /// <param name="blockerType">Output parameter for the type of movement blocker, if any.</param>
+        /// <param name="entityType">Output parameter for the type of entity blocking movement, if any.</param>
+        /// <returns>Returns true if the entity can move in the given direction, or false if movement is blocked.</returns>
+        public virtual bool CanMoveInDirection(Direction direction,
+            out MovementBlockerType blockerType,
+            out EntityType entityType)
+        {
             // If this is an Npc that has the Static behaviour, it can NEVER move.
             if (this is Npc npc)
             {
                 if (npc.Base.Movement == (byte)NpcMovement.Static)
                 {
-                    return -2;
+                    blockerType = MovementBlockerType.OutOfBounds;
+                    entityType = default;
+                    return false;
                 }
             }
 
+            var xOffset = 0;
+            var yOffset = 0;
             var tile = new TileHelper(MapId, X, Y);
-            switch (moveDir)
+
+            switch (direction)
             {
                 case Direction.Up:
                     yOffset--;
-
                     break;
+
                 case Direction.Down:
                     yOffset++;
-
                     break;
+
                 case Direction.Left:
                     xOffset--;
-
                     break;
+
                 case Direction.Right:
                     xOffset++;
-
                     break;
+
                 case Direction.UpLeft:
                     yOffset--;
                     xOffset--;
-
                     break;
+
                 case Direction.UpRight:
                     yOffset--;
                     xOffset++;
-
                     break;
+
                 case Direction.DownRight:
                     yOffset++;
                     xOffset++;
-
                     break;
 
                 case Direction.DownLeft:
                     yOffset++;
                     xOffset--;
-
                     break;
             }
 
-            MapController mapController = null;
-            int tileX = 0;
-            int tileY = 0;
-
             if (tile.Translate(xOffset, yOffset))
             {
-                mapController = MapController.Get(tile.GetMapId());
-                tileX = tile.GetX();
-                tileY = tile.GetY();
-                var tileAttribute = mapController.Attributes[tileX, tileY];
-                if (tileAttribute != null)
-                {
-                    if (tileAttribute.Type == MapAttribute.Blocked || (tileAttribute.Type == MapAttribute.Animation && ((MapAnimationAttribute)tileAttribute).IsBlock))
-                    {
-                        return -2;
-                    }
-
-                    if (tileAttribute.Type == MapAttribute.NpcAvoid && (this is Npc || (this is EventPageInstance evtPage && !evtPage.MyPage.IgnoreNpcAvoids)))
-                    {
-                        return -2;
-                    }
-
-                    if (tileAttribute.Type == MapAttribute.ZDimension &&
-                        ((MapZDimensionAttribute)tileAttribute).BlockedLevel > 0 &&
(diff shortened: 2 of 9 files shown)


Intersect.Server/Entities/Player.cs
```diff
@@ -6470,52 +6470,64 @@ CommonEventTrigger trigger
...
+                    instance = instance.GlobalClone;
+                }
+
+                if (instance.Map == map &&
+                    instance.X == x &&
+                    instance.Y == y &&
+                    instance.Z == z &&
+                    !instance.Passable)
+                {
+                    return false;
```
lodicolo, line 6526: This is losing information. The method returned the blocker before, and this does not return a blocker.

Intersect.Server/Entities/Entity.cs
```diff
@@ -576,28 +590,29 @@ public virtual int CanMove(Direction moveDir)
...
             }
 
-            //Otherwise fine
-            return -1;
+            blockerType = MovementBlockerType.OutOfBounds;
+            entityType = EntityType.GlobalEntity;
+            return false;
         }
 
+        protected virtual bool IsTileWalkable(MapController map, int x, int y, int z) => map != null;
```
lodicolo, line 614: This should return a `MovementBlockerType` and should be renamed to `GetTileBlocker`.

TBH I think the X/Y out of bounds check should probably happen here in the default implementation.

lodicolo (changes requested):

Intersect.Server/Entities/Entity.cs
```diff
@@ -460,110 +460,124 @@ public virtual int CanMove(Direction moveDir)
...
+                            return false;
 
-                    if (tileAttribute.Type == MapAttribute.ZDimension &&
-                        ((MapZDimensionAttribute)tileAttribute).BlockedLevel > 0 &&
-                        ((MapZDimensionAttribute)tileAttribute).BlockedLevel - 1 == Z)
-                    {
-                        return -3;
-                    }
+                        case MapAttribute.Slide:
+                            switch (((MapSlideAttribute)tileAttribute).Direction)
```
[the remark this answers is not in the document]
Arufonsu: After checking this up and seeing the values we are getting from this, and the fact that all directions are supported and limited by the core when configured, i've realized that we can reduce this down to simply:

```csharp

case MapAttribute.Slide:
                    blockerType = MovementBlockerType.Slide;
                    entityType = EntityType.GlobalEntity;
                    return false;
```

tested with slide map attributes to be working as expected

**Also, a self reminder:** we should implement diagonal slide attributes since we can surely support them now like we do with knockbacks and dash.

imagen

Intersect.Server/Entities/Entity.cs
```diff
@@ -460,110 +460,124 @@ public virtual int CanMove(Direction moveDir)
                 var tileAttribute = mapController.Attributes[tileX, tileY];
                 if (tileAttribute != null)
                 {
-                    if (tileAttribute.Type == MapAttribute.Blocked || (tileAttribute.Type == MapAttribute.Animation && ((MapAnimationAttribute)tileAttribute).IsBlock))
+                    switch (tileAttribute.Type)
```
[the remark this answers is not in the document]
Arufonsu:
```csharp
            var mapController = MapController.Get(tile.GetMapId());
            int tileX = tile.GetX();
            int tileY = tile.GetY();
            var tileAttribute = mapController?.Attributes[tileX, tileY];

            switch (tileAttribute)
            {
                case MapAnimationAttribute animAttr when animAttr.IsBlock:
                case MapBlockedAttribute _:
                case MapNpcAvoidAttribute _
                    when (this is Npc || (this is EventPageInstance evtPage && !evtPage.MyPage.IgnoreNpcAvoids)):
                    blockerType = MovementBlockerType.MapAttribute;
                    entityType = EntityType.GlobalEntity;
                    return false;

                case MapZDimensionAttribute zAttr when zAttr.BlockedLevel > 0 && zAttr.BlockedLevel - 1 == Z:
                    blockerType = MovementBlockerType.ZDimension;
                    entityType = EntityType.GlobalEntity;
                    return false;

                case MapSlideAttribute _:
                    blockerType = MovementBlockerType.Slide;
                    entityType = EntityType.GlobalEntity;
                    return false;
            }
```

Intersect.Server/Entities/Entity.cs
```diff
@@ -576,26 +537,18 @@ public virtual int CanMove(Direction moveDir)
...
-            //Out of bounds if no map
-            if (map == null)
-            {
-                return -5;
-            }
-
-            //Otherwise fine
-            return -1;
+            blockerType = MovementBlockerType.NotBlocked;
+            entityType = EntityType.GlobalEntity;
```
lodicolo, line 550: `default`

Intersect.Server/Entities/Entity.cs
```diff
@@ -381,189 +381,150 @@ public virtual void Update(long timeMs)
...
             if (!Passable)
             {
-                var targetMap = mapController;
                 var mapEntities = new List<Entity>();
+                var zoneType = (int)mapController.ZoneType;
+
                 if (mapController.TryGetInstance(MapInstanceId, out var mapInstance))
                 {
-                    mapEntities.AddRange(mapInstance.GetCachedEntities());
+                    mapEntities.AddRange(mapInstance?.GetCachedEntities());
```
lodicolo, line 503: `mapInstance` can't be null here

Intersect.Server/Entities/Entity.cs
```diff
@@ -381,189 +381,150 @@ public virtual void Update(long timeMs)
...
-                        }
+                        continue;
+                    }
+
+                    //Set a target if a projectile
+                    CollisionIndex = en.Id;
+
+                    switch (en)
+                    {
+                        case Player _ when !Options.Instance.Passability.Passable[zoneType]:
```
lodicolo, line 518: This is not logically equivalent to the old code.

Intersect.Server/Entities/Entity.cs
```diff
@@ -381,189 +381,150 @@ public virtual void Update(long timeMs)
...
+                    blockerType = MovementBlockerType.Slide;
+                    entityType = EntityType.GlobalEntity;
+                    return false;
             }
 
             if (!Passable)
             {
-                var targetMap = mapController;
                 var mapEntities = new List<Entity>();
+                var zoneType = (int)mapController.ZoneType;
```
lodicolo, line 499: After this line put `var blockableByPlayer = !(this is Player) || !Options.Instance.Passability.Passable[zoneType]`

Intersect.Server/Entities/Entity.cs
lodicolo: Replace with `blockableByPlayer`

Intersect.Server/Entities/Events/EventPageInstance.cs
```diff
@@ -372,340 +373,339 @@ protected override bool ProcessMoveRoute(Player forPlayer, long timeMs)
...
+                    return true;
+                }
+
+                switch (MoveRoute.Actions[MoveRoute.ActionIndex].Type)
+                {
+                    case MoveRouteEnum.MoveTowardsPlayer:
+                        //Pathfinding required.. this will be weird.
+                        if (forPlayer != null && GlobalClone == null) //Local Event
+                        {
+                            if (mPathFinder.GetTarget() == null ||
```
lodicolo, line 388: `var target = mPathFinder.GetTarget();`

Intersect.Server/Entities/Events/EventPageInstance.cs
```diff
@@ -372,340 +373,339 @@ protected override bool ProcessMoveRoute(Player forPlayer, long timeMs)
...
-                                }
+                                mPathFinder.SetTarget(
+                                    new PathfinderTarget(forPlayer.MapId, forPlayer.X, forPlayer.Y, forPlayer.Z)
+                                );
+                            }
 
-                                //Todo check if next to or on top of player.. if so don't run pathfinder.
-                                if (mPathFinder.Update(timeMs) == PathfinderResult.Success)
+                            //Todo check if next to or on top of player.. if so don't run pathfinder.
+                            if (mPathFinder.Update(timeMs) == PathfinderResult.Success)
```
lodicolo, line 399: Invert and break

Intersect.Server/Entities/Events/EventPageInstance.cs
```diff
@@ -372,340 +373,339 @@ protected override bool ProcessMoveRoute(Player forPlayer, long timeMs)
...
-                                        if (CanMove(pathDir) == -1)
-                                        {
-                                            Move(pathDir, forPlayer);
-                                            moved = true;
-                                        }
-                                        else
-                                        {
-                                            mPathFinder.PathFailed(timeMs);
-                                        }
+                                        mPathFinder.PathFailed(timeMs);
```
lodicolo, line 411: This line literally can't be hit now. When `CanMoveInDirection()` returns true `blockerType` is always going to be `NotBlocked`. The `if (blockerType == MovementBlockerType.NotBlocked)` isn't needed and the else should be moved one level up.

lodicolo (changes requested):

Intersect.Server/Entities/Events/EventPageInstance.cs
Arufonsu: 👍🏽 

```csharp
                            var pathDir = mPathFinder.GetMove();
                            if (pathDir > Direction.None && CanMoveInDirection(pathDir, out _, out _))
                            {
                                Move(pathDir, forPlayer);
                                moved = true;
                            }
                            else
                            {
                                mPathFinder.PathFailed(timeMs);
                            }
```

lodicolo:
@Arufonsu can you figure out why the PR is listing changes already in main as "changed" in this PR?

Intersect.Server/Entities/Events/EventPageInstance.cs
lodicolo: This should be `pathDir != Direction.None` rather than `>`, there is no guarantee the tombstone value won't get moved.

lodicolo: You can also just do `CanMoveInDirection(pathDir)`? That overload exists so you don't have to do `out _, out _`

Arufonsu:
> @Arufonsu can you figure out why the PR is listing changes already in main as "changed" in this PR?

imagen

i can update the branch and squash the needed commits if that helps

Intersect.Server/Entities/Combat/Dash.cs
```diff
@@ -50,40 +51,60 @@ public partial class Dash
...
+                                break;
+                            }
+
+                            return;
+
+                        case MovementBlockerType.Entity:
+                            switch (entityType)
+                            {
+                                case EntityType.Resource when activeResourcePass == false:
+                                case EntityType.Resource when deadResourcePass == false:
```
lodicolo, line 85: Make these into fall-throughs with an if (pass) { break; }

1. using pattern matching kills the jump table when compiled (the entire reason to use a switch)
2. when *Pass is true, it's caught by default now instead, which throws an exception when it should not be

Intersect.Server/Entities/Entity.cs
```diff
@@ -381,189 +381,154 @@ public virtual void Update(long timeMs)
...
+                    entityType = default;
+                    return false;
 
-                                break;
-                            case 4:
-                                if (moveDir == Direction.Left)
-                                {
-                                    return -4;
-                                }
+                case MapZDimensionAttribute zAttr when zAttr.BlockedLevel > 0 && zAttr.BlockedLevel - 1 == Z:
```
lodicolo, line 488: For the reasons I mentioned in the other file, don't use this pattern matching

Intersect.Server/Entities/Entity.cs
```diff
@@ -381,189 +381,154 @@ public virtual void Update(long timeMs)
...
+
+                    switch (en)
+                    {
+                        case Player _ when blockableByPlayer:
+                        case Npc _:
+                            blockerType = MovementBlockerType.Entity;
+                            entityType = EntityType.Player;
+                            return false;
+
+                        case Resource resource when !resource.IsPassable():
```
lodicolo, line 528: More pattern matching to remove

lodicolo (changes requested):

Intersect.Server/Entities/Combat/Dash.cs
Arufonsu:
```csharp
                                    case EntityType.Resource:
                                    if (activeResourcePass || deadResourcePass)
                                    {
                                        break;
                                    }

                                    return;
```

Intersect.Server/Entities/Combat/Dash.cs
```diff
@@ -50,40 +51,63 @@ public partial class Dash
...
         )
         {
-            var n = 0;
             en.MoveTimer = 0;
             Range = 0;
             for (var i = 1; i <= range; i++)
             {
-                n = en.CanMove(Direction);
-                if (n == -5) //Check for out of bounds
+                if (!en.CanMoveInDirection(Direction, out var blockerType, out var entityType))
```
lodicolo, line 58: Invert, don't add nesting

Intersect.Server/Entities/Combat/Dash.cs
```diff
@@ -50,40 +51,63 @@ public partial class Dash
...
+                        case MovementBlockerType.Entity:
+                            switch (entityType)
+                            {
+                                case EntityType.Resource:
+                                    if (activeResourcePass || deadResourcePass)
+                                    {
+                                        break;
+                                    }
+
+                                    return;
```
lodicolo, line 90: This does _not_ match the original logic.

Intersect.Server/Entities/Player.cs
```diff
@@ -6470,52 +6470,72 @@ CommonEventTrigger trigger
...
-                            !instance.Passable)
-                        {
-                            return (int) EntityType.Event;
-                        }
-                    }
+                var instance = evt.Value.PageInstance;
+
+                if (instance.GlobalClone != null)
+                {
+                    instance = instance?.GlobalClone;
```
lodicolo, line 6519: Why is there a nullish operator right after checking that it isn't null...?

Also this reads from `GlobalClone` twice (potentially has a different value the second time), just change this to `instance = instance?.GlobalClone ?? instance;`

Intersect.Server/Entities/Player.cs
```diff
@@ -6470,52 +6470,72 @@ CommonEventTrigger trigger
...
+                    continue;
+                }
+
+                blockerType = MovementBlockerType.Entity;
+                entityType = EntityType.Event;
+                return false;
             }
 
-            return -1;
+            blockerType = MovementBlockerType.NotBlocked;
```
lodicolo, line 6536: `default`

lodicolo (changes requested):

Intersect.Server/Entities/Combat/Dash.cs
Arufonsu: uhm, but this is functionally equivalent to this:

```csharp
                if (n == (int) EntityType.Resource && activeResourcePass == false)
                {
                    return;
                } //Check for dead resources

                if (n == (int) EntityType.Resource && deadResourcePass == false)
                {
                    return;
                }
```

The original code uses a series of conditional statements to check if the entity type is a resource and if certain flags are set, and then returns early if the conditions are not met.

The new one uses a switch statement to check the entity type, and then checks the same flags using a single if statement with a logical OR operator. If the conditions are met, the code block breaks out of the switch statement, otherwise it returns early.

So, both implementations achieve the same result: if the entity type is a resource and the necessary flags are not set, the code exits the block (return) and does not execute any further instructions.

Intersect.Server/Entities/Combat/Dash.cs
Arufonsu: But you said: https://github.com/AscensionGameDev/Intersect-Engine/pull/1775#discussion_r1157976094

and if we invert this -> https://github.com/AscensionGameDev/Intersect-Engine/pull/1775#discussion_r1157939900

also if we invert this, where do we place this within the for chain? inverting would mean to use the following early:

```csharp
                en.Move(Direction, null, true);
                en.Dir = Facing;
                Range = i;
```

this would make the switches... useless in the logic chain (unless we duplicate these 3 lines at the very end of the for method then but this doesn't sounds efficient...), from what i'm seeing this can be either done by using the negated version of the boolean (like it is atm) or either don't use the boolean return type which is what you pointed out to be changed in the previous review.

Intersect.Server/Entities/Combat/Dash.cs
lodicolo: It is not.

Before
- if the blocking entity was a resource and
  - `activeResourcePass` was false and `deadResourcePass` was true, it would return
  - `deadResourcePass` was false and `activeResourcePass` was true, it would return
  
Your new code `break`s if either one is `true`, instead of `return`ing if either one is `false`. They are not functionally equivalent.

Intersect.Server/Entities/Combat/Dash.cs
lodicolo: Correct, it would duplicate the lines at the end. Disregard.

Intersect.Server/Entities/Combat/Dash.cs
Arufonsu: oh 🤔

would something like this do it?

```csharp
switch (blockerType)
{
    case EntityType.Resource:
        if (!activeResourcePass && deadResourcePass)
        {
            return;
        }
        break;
    // other cases omitted for brevity
}
```

lodicolo:
Honestly I decided around the last time I reviewed this PR that it's too complicated and I have found too many issues with it that I don't actually have confidence to approve it.

And if I am not going to approve it I won't merge it, so I'm just going to close it.

PR #1788, opened 2023-04-16
enhancement: configurable map items size

Arufonsu:
enhances spawned/dropped items in maps by:
 - Adding configurable size for their sprites.
 - Centering their sprites in relation to the tiles they occupy.

_This doesn't requires additional texture loading, this simply allows the users to configure the size the icon textures.
The implementation also makes sure to always center the item's texture in relation with the tile they occupy while considering the configured item render size, this way we are not limited by the boundaries of the tile size of the grid anymore._

Preview:

imagen

imagen

### DIFF
--- a/Intersect (Core)/Config/MapOptions.cs
+++ b/Intersect (Core)/Config/MapOptions.cs
@@ -39,6 +39,16 @@ namespace Intersect.Config
         public LayerOptions Layers { get; set; } = new LayerOptions();
 
         /// <summary>
+        /// The width of map items.
+        /// </summary>
+        public int MapItemHeight { get; set; } = 32;
+
+        /// <summary>
+        /// The height of map items.
+        /// </summary>
+        public int MapItemWidth { get; set; } = 32;
+
+        /// <summary>
         /// The height of the map in tiles.
         /// </summary>
         public int MapHeight { get; set; } = 26;
--- a/Intersect (Core)/Config/MapOptions.cs
+++ b/Intersect (Core)/Config/MapOptions.cs
@@ -41,12 +41,12 @@ namespace Intersect.Config
         /// <summary>
         /// The width of map items.
         /// </summary>
-        public int MapItemHeight { get; set; } = 32;
+        public uint MapItemHeight { get; set; }
 
         /// <summary>
         /// The height of map items.
         /// </summary>
-        public int MapItemWidth { get; set; } = 32;
+        public uint MapItemWidth { get; set; }
 
         /// <summary>
         /// The height of the map in tiles.
--- a/Intersect (Core)/Config/MapOptions.cs
+++ b/Intersect (Core)/Config/MapOptions.cs
@@ -119,6 +119,9 @@ namespace Intersect.Config
             {
                 throw new Exception("Config Error: Map size out of bounds! (All values should be > 10 and < 64)");
             }
+
+            MapItemWidth = MapItemWidth < 1 ? (uint)TileWidth : MapItemWidth;
+            MapItemHeight = MapItemHeight < 1 ? (uint)TileHeight : MapItemHeight;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Intersect (Core)/Config/Options.cs
```diff
@@ -146,6 +146,10 @@ public partial class Options
 
         public static bool ZDimensionVisible => Instance.MapOpts.ZDimensionVisible;
 
+        public static int MapItemHeight => Instance?.MapOpts?.MapItemHeight ?? 32;
+
+        public static int MapItemWidth => Instance?.MapOpts?.MapItemWidth ?? 32;
+
```
lodicolo, line 152: Stop adding more statics

Intersect (Core)/Config/MapOptions.cs
```diff
@@ -38,6 +38,16 @@ public bool EnableDiagonalMovement
...
+        /// <summary>
+        /// The width of map items.
+        /// </summary>
+        public int MapItemHeight { get; set; } = 32;
+
+        /// <summary>
+        /// The height of map items.
+        /// </summary>
+        public int MapItemWidth { get; set; } = 32;
+
```
lodicolo, line 50: > The <word> of the map items in pixels.

Also, remove the `= 32;`, and change these to `uint`s.

lodicolo (changes requested):

Intersect (Core)/Config/Options.cs
Arufonsu: offtopic and sort of related question, i had my doubts when doing this so, asking for later (a later chore PR perhaps): do we want to get rid of all these statics at Config/Options.cs eventually?

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -718,32 +719,37 @@ public void DestroyVBOs()
...
+                    var tileX = key % Options.MapWidth;
+                    var tileY = (int)Math.Floor(key / (float)Options.MapWidth);
+                    var x = X + tileX * Options.TileWidth;
+                    var y = Y + tileY * Options.TileHeight;
+                    var centerX = x + (Options.TileWidth / 2);
+                    var centerY = y + (Options.TileHeight / 2);
+                    var textureWidth = Options.MapItemWidth;
+                    var textureHeight = Options.MapItemHeight;
+                    var textureXPosition = centerX - (textureWidth / 2);
+                    var textureYPosition = centerY - (textureHeight / 2);
```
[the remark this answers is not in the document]
Arufonsu: should i move the variables out of the for loop where they were originally at (not all of them were) and only change the ones that were originally within the loop _only_ if there's valid configuration values -> else: proceed as the code did in the original logic (?)

Intersect.Client/Maps/MapInstance.cs
```diff
@@ -717,42 +718,56 @@ public void DestroyVBOs()
...
         {
-            // Draw map item icons.
-            foreach (var itemCollection in MapItems)
+            // Calculate tile and map item dimensions.
+            var tileWidth = (uint)Options.TileWidth;
+            var tileHeight = (uint)Options.TileHeight;
+            var preferredMapItemWidth = Options.Instance.MapOpts.MapItemWidth;
+            var preferredMapItemHeight = Options.Instance.MapOpts.MapItemHeight;
+            var mapItemWidth = preferredMapItemWidth <= 0 ? tileWidth : preferredMapItemWidth;
+            var mapItemHeight = preferredMapItemHeight <= 0 ? tileHeight : preferredMapItemHeight;
```
lodicolo, line 727: This should be getting done in `MapOptions.Validate()` in the other file, like I said this gets done when it is loaded.

Also switch to `< 1`

lodicolo (changes requested):

Arufonsu:
Updated PR for B8

lodicolo (approved):

Issue #1725, opened 2023-02-19
bug: Weather animations get stuck on transition between 2 maps


NikoCummings:
### Description

You can be walking and the raindrop animations are falling as intended. Then when transitioning between two maps some of the raindrop animations get stuck on the map and stop moving in the direction you set it. This lasts briefly and then vanishes.
Here's a video of what I mean: (https://s3.us-east-2.amazonaws.com/ascensiongamedev/filehost/a047e4ca9ed40859e835b3394b6aab05.mp4)

### Steps to Reproduce

- Login to said version of the editor 
- add weather 
- make multiple maps
- walk through the many maps with the same weather added to them

### Version with bug

0.7.1.92

### Last version that worked well

unkown

### Affected platforms

Windows

### Did you find any workaround?

No.

### Duplicate Bug Check

- [X] This bug report is not a duplicate to the best of my knowledge.

lodicolo:
@Arufonsu Is this fixed by #2005 and #2007 ?

Arufonsu:
> @Arufonsu Is this fixed by #2005 and #2007 ?

this is quite late for a reply, but this seems to be fixed with current [v0.8.0-beta.191](https://github.com/AscensionGameDev/Intersect-Engine/releases/tag/v0.8.0-beta.191) (I'm unable to reproduce this)

PR #1792, opened 2023-04-24
fix: client-side timing synchronization

Arufonsu:
### Potentially fixes the following:
- If the server is started while the client is already open while in the main menu, the animated background (if there's any) abruptly stops as the client detects the server's status changing from offline to online due to time synchronization issues, this can rely on MillisecondsUtc, which avoid this unintended behavior.
- Similar issue happens with the day/time/hue color cycles in game, we don't need to sync the clients to Milliseconds (sColorUpdate), MillisecondsUtc is the way.
- Hopefully fixes a reported issue where weather suddenly "stops" falling in some maps [(?) (need more testing - unable to reproduce yet).](https://github.com/AscensionGameDev/Intersect-Engine/issues/1725)

### As overall, the following can rely on MillisecondsUtc (client's local timing)
Without having to worry about server timing synchronization (+ potentially avoids issues): 
- Animations (intro/world/auto.tiles/entities)
- Audio
- Critters
- Basic interface elements such as click timers, context menus, typewriter, inventory, settings, chat box, etc.
- Graphic effects such as fading, fog, weather and rendering in general.

### Timing.Global.Milliseconds is still used for things that require timing synchronization:
- ServerTime timing update.
- Player input.
- Mobility.
- Attack, cast, shield and projectile handlers.
- Networking handlers.

From discord:

imagen

To consider (not part of this PR but is quite do-able - to be discussed (?) ):

imagen

### DIFF
--- a/Intersect.Client/Interface/Game/AnnouncementWindow.cs
+++ b/Intersect.Client/Interface/Game/AnnouncementWindow.cs
@@ -1,6 +1,5 @@
 ﻿using Intersect.Client.Core;
 using Intersect.Client.Framework.File_Management;
-using Intersect.Client.Framework.Graphics;
 using Intersect.Client.Framework.Gwen.Control;
 
 using Intersect.Utilities;
@@ -57,7 +56,7 @@ namespace Intersect.Client.Interface.Game
                 mLabel.Text = mLabelText;
 
                 // Are we still supposed to be visible?
-                if (Timing.Global.Milliseconds > mDisplayUntil)
+                if (Timing.Global.MillisecondsUtc > mDisplayUntil)
                 {
                     Hide();
                 }
@@ -72,7 +71,7 @@ namespace Intersect.Client.Interface.Game
         public void ShowAnnouncement(string announcementText, long displayTime)
         {
             mLabelText = announcementText;
-            mDisplayUntil = Timing.Global.Milliseconds + displayTime;
+            mDisplayUntil = Timing.Global.MillisecondsUtc + displayTime;
             Show();
         }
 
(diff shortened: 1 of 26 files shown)


Intersect.Client/Maps/MapInstance.cs
```diff
@@ -1260,7 +1260,7 @@ public void DrawActionMsgs()
                     ActionMessages[n].Y * Options.TileHeight -
                     Options.TileHeight *
                     2 *
-                    (1000 - (ActionMessages[n].TransmissionTimer - Timing.Global.Milliseconds)) /
+                    (1000 - (ActionMessages[n].TransmissionTimer - Timing.Global.MillisecondsUtc)) /
```
lodicolo, line 1263: I am not sure about this being not-synchronized

Intersect.Client/Interface/Game/AnnouncementWindow.cs
lodicolo, line 1: I am not sure about this being not-synchronized

lodicolo (changes requested):

Intersect.Client/Maps/MapInstance.cs
Arufonsu: this timer is used to **_graphically_ render the strings of action messages** on screen going up and fading out, it doesn't matter if the rendered position of these aren't synced in time, just like animations, it should be good as long as the user can see the action messages on their side.

Preview (using MillisecondsUtc):

https://user-images.githubusercontent.com/17498701/234128115-d59dfff7-29da-456c-a854-1968f46c8c84.mp4

Intersect.Client/Interface/Game/AnnouncementWindow.cs
Arufonsu: not sure about this one either, hmm.. if we want these to be up X amount of synced time, then yeah.. i'll change this back.

lodicolo (approved):

PR #1802, opened 2023-05-05
Development: Make it a real server

Zebra000:
Hello, I'm stuck with the server engine !

I thinked, if my schools and my past, that a server software must be autonomous, without require a keyboard or gui environment (include gui).
A server should start alone, without human to start.
A server don't require an Internet Connection with a human at the other side on the planet to start the server in a datacenter, event with a reboot of the hardware.
A server don't stop just because no input or client is disconnected (the admin) with no luck on the meteorology outside of his house, which has no common point with the server in a datacenter.....

So, can I have help to avoid me to recode your piece of code into a real server software (on Microsoft it's called a service, on linux or derivated it's a daemon) who run in autonomy, alone, without a needed human on the internet to launch a client and stay on it juste to server run please ?

I don't care the server is coded in dot net, it can run with mono, I do that.
But why the software shutdown if and only if I redirect all input from /dev/null to run it like a ..... real server software ?

In thirty years in numeric, I never see a server software used a software desktop type application, I don't understand your choices on that point, unbelievable.

 Can you help or correct your code please ? Or shoud I tell to my client to use another engine more profeessional and recode his entire game else my client should never start in this domain ?

Thank you, and yeah, I know, I'm angry, but if you read that entirely, respect and I maybe calm down myself.


Arufonsu:
Not sure what your intentions were with this, but feel free to use https://github.com/AscensionGameDev/Intersect-Engine/discussions in order to do what you just did here.... opening a PR as a rant makes no sense at all if you want to contribute to the work flow of the engine.

PR #1813, opened 2023-05-11
fix: cleanse behavior

### LINKED ISSUE #1812: bug: cleanse behavior
### Description

- Cleanse is not properly dispelling currently active "unfriendly" Statuses and DoTs.
- Cleanse is not preventing new "unfriendly" Statuses and DoTs while it's active over time.

### Steps to Reproduce

1. create a spells that cleans up status effects (cleanse spell)

imagen

2. create an "unfriendly" spell with Stat Duration or DoT

imagen

3. While in game, test with two character, cast the unfriendly spell over one of the characters

4.  Try to dispell the unfriendly spell with the cleanse spell, result: 
  - According to @WeylonSantana it does cleans the spell
  - From my testing, it doesn't.. probably because i tested it up with DoTs (?) it should still... clean them up !
  
5. As seen in the screen shot, our cleanse effect should last for up to 5 seconds, change it up to 10 secs.

6.  Use the cleanse spell, then attack the character with the unfriendly spell we made earlier, results:
  - Expected: cleanse, while active over time, should prevent new unfriendly DoT/Status effects from taking place.
  - Reality: cleanse, while active over time, doesn't prevents new unfriendly DoT/Status effects from taking place.

### Version with bug

0.7.2.b15

Unknown

### Affected platforms

Windows, Linux

Yes, i'll post a PR for this.

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
Refactors and fix cleanse status effect behavior so it will now:
- Properly dispel currently active "unfriendly" Statuses and DoTs - also skips itself from being dispelled.
- Prevent/block "unfriendly" Statuses and DoTs while active over time.
- Should resolve #1812

Preview:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/a06b4431-2298-4d35-a999-dac544220313

### DIFF
--- a/Intersect.Server/Entities/Combat/DoT.cs
+++ b/Intersect.Server/Entities/Combat/DoT.cs
@@ -24,7 +24,6 @@ namespace Intersect.Server.Entities.Combat
         public DoT(Entity attacker, Guid spellId, Entity target)
         {
             SpellBase = SpellBase.Get(spellId);
-
             Attacker = attacker;
             Target = target;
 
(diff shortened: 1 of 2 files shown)


lodicolo:
This PR just doesn't make any sense. Why are other statuses being dispelled in the constructor for a DoT effect...? The cleanse handling in the constructor was to prevent _adding_ a new DoT.

Arufonsu:
> The cleanse handling in the constructor was to prevent _adding_ a new DoT

Because DoTs cast statuses and well, now cleanse not only prevents unfriendly spells/debuffs but also _cleans_ up the active ones

lodicolo:
> > The cleanse handling in the constructor was to prevent _adding_ a new DoT
> 
> Because DoTs cast statuses and well, now cleanse not only prevents unfriendly spells/debuffs but also _cleans_ up the active ones

tl;dr: I do not disagree with cleaning up existing statuses that are incompatible with Cleanse, but the existing statuses should be cleaned up when the Cleanse is added, not in the constructor for a DoT later on.

Yeah but my point was that in the effect handler that added the cleanse effect in the first place everything should have been cleaned up. The DoT constructor should just "abort" if it can't be added, but it shouldn't clean up a mess that should have already been cleaned up. Otherwise none of those statuses would get cleaned up unless the DoT is added, what if a DoT is never added on the player until after the cleanse expires? Then the other statuses would just be on the player when they shouldn't be.

PR #1826, opened 2023-07-06
chore: refactor Entity#CanMove()'s return types

### LINKED ISSUE #1100: chore: refactor Entity#CanMove()'s return types

lodicolo:
Fixes #1100 

Note to reviewers: the easiest way to review this is by reviewing each commit one-by-one in chronological order

The names of each commit say exactly what operation was done in each commit so some of the commits are basically not necessary to review.

### DIFF
--- a/Intersect (Core)/Enums/Direction.cs
+++ b/Intersect (Core)/Enums/Direction.cs
@@ -1,3 +1,5 @@
+using System;
+
 namespace Intersect.Enums
 {
     public enum Direction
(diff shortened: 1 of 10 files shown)


Arufonsu (approved):
Approved; tested locally with dash and other movement features; everything seems fine. This PR fulfills what #1775 was originally meant to solve, and it's less borked and much more organized, following a better commit structure with shorter changes.

PR #1829, opened 2023-07-13
don't target dead NPCs

### LINKED ISSUE #1730: bug: 'Ghosting' Target Box after entity death
### Description

Targeting an NPC, then killing it will make the target box appear 'ghosted' when trying to target another NPC (if there are no other NPCs to target)

Targeting a new NPC will clear the ghosting target box.

Video: https://youtu.be/7H-Yc1CGyls

### Steps to Reproduce

1. Make a map with 1 NPC
2. Target NPC, then kill it.
3. Try to target when there are no targets.

### Version with bug

v0.7.1.88

Unknown

### Affected platforms

Windows

no

### Relevant log output

_No response_

### Duplicate Bug Check

WeylonSantana:
Resolves #1730
Simple fix, after the entity was selected, there was no check if the selected entity existed within the list of entities, so the select happened, and the update cleared it. I put the same check as the update, before selecting the entity.

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -1415,16 +1415,22 @@ namespace Intersect.Client.Entities
                 return;
             }
 
-            if (mlastTargetList.ContainsKey(currentEntity))
+            Globals.Entities.TryGetValue(currentEntity.Id, out var targetedEntity);
+            if(targetedEntity == default)
             {
-                mlastTargetList[currentEntity].LastTimeSelected = Timing.Global.Milliseconds;
+                return;
             }
-            mLastEntitySelected = currentEntity as Entity;
 
-            if (TargetIndex != currentEntity.Id)
+            if (mlastTargetList.ContainsKey(targetedEntity))
             {
-                SetTargetBox(currentEntity as Entity);
-                TargetIndex = currentEntity.Id;
+                mlastTargetList[targetedEntity].LastTimeSelected = Timing.Global.Milliseconds;
+            }
+            mLastEntitySelected = targetedEntity;
+
+            if (TargetIndex != targetedEntity.Id)
+            {
+                SetTargetBox(targetedEntity);
+                TargetIndex = targetedEntity.Id;
                 TargetType = 0;
             }
         }


Intersect.Client/Entities/Player.cs
```diff
@@ -1415,16 +1415,22 @@ public void AutoTarget()
                 return;
             }
 
-            if (mlastTargetList.ContainsKey(currentEntity))
+            Globals.Entities.TryGetValue(currentEntity.Id, out var targetedEntity);
+            if(targetedEntity == default)
```
lodicolo, line 1419: Just turn this into `!Globals.Entities.TryGetValue(...)` -- TryGetValue returns false if the out variable is default

Intersect.Client/Entities/Player.cs
```diff
@@ -1415,16 +1415,22 @@ public void AutoTarget()
...
             {
-                mlastTargetList[currentEntity].LastTimeSelected = Timing.Global.Milliseconds;
+                return;
+            }
+
+            if (mlastTargetList.ContainsKey(targetedEntity))
```
lodicolo, line 1424: `mlastTargetList` should also have a TryGetValue to use instead of ContainsKey?

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
```diff
@@ -1415,16 +1415,21 @@ public void AutoTarget()
...
-            if (mlastTargetList.ContainsKey(currentEntity))
+            if(!Globals.Entities.TryGetValue(currentEntity.Id, out var targetedEntity))
             {
-                mlastTargetList[currentEntity].LastTimeSelected = Timing.Global.Milliseconds;
+                return;
+            }
+
+            if (mlastTargetList.TryGetValue(targetedEntity, out var _))
+            {
+                mlastTargetList[targetedEntity].LastTimeSelected = Timing.Global.Milliseconds;
```
lodicolo, line 1425: You are supposed to use the out variable here not look it up again in the target list.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
WeylonSantana: like this?
```
            if (mlastTargetList.TryGetValue(targetedEntity, out var lastTarget))
            {
                lastTarget.LastTimeSelected = Timing.Global.Milliseconds;
            }
```

lodicolo: Yes @WeylonSantana

lodicolo (changes requested):

lodicolo (approved):

PR #1832, opened 2023-07-14
fix: correctly removes currency amount inShop

### LINKED ISSUE #1727: bug: Trying to buy more items then you can store in inventory makes player loose cash, gain no items, no warning message
### Description

If you go to a store with a ton of cash and try to buy an item in bulk. The slider is automatically pushed all the way to the right so the player can buy max affordable amount
if the player decides to do this and does not have enough inventory slots to hold the items
the player's cash all disappears and no items are transferred to the inventory, with no error message or warning or something that prevents this action. It just vanishes

### Steps to Reproduce

Go to a shop with tons of cash
try to buy an item that does not stack, so each item takes up one inventory slot
buy as many of that item as the slider allows (should be many more than your inventory can hold)
notice how all your cash disappears, not items are given, no warning or errors, nothing preventing this from happening

here is a video: https://s3.us-east-2.amazonaws.com/ascensiongamedev/filehost/a8c0bb272047edd115979bf6cda70024.mp4

### Version with bug

0.7.1.92

unkown

### Affected platforms

Windows

No

### Relevant log output

_No response_

### Duplicate Bug Check

WeylonSantana:
resolves #1727 

Full refactoring of the server BuyItem method.

- Now the server says to client if there is no space in the inventory and won't let you buy (resolving the bug)
- Added bug checks if buy or sell item is null
- Correctly removes the amount of currency-item from the player
  - If the currency-item was not stockable (don't ask me why a currency-item would be non-stockable), the FindInventoryItemSlot methods only look for the first slot with the desired amount, I improved the search to remove the desired amount from all slots, regardless of whether it is stockable or not, in addition to using a better method to know the amount of items that the player has.

Note: Selling items is buggy the same way buying items was, when playing with inventory limits, my next pr will be to refactor the method of selling items, I'll keep the scope of this pr just about buying, so don't be alarmed when selling items and your items disappear and you don't receive the purchased item.

### DIFF
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -3520,13 +3520,13 @@ namespace Intersect.Server.Entities
         {
             if (InShop == default)
             {
-                PacketSender.SendChatMsg(this, Strings.Shops.error, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                PacketSender.SendChatMsg(this, Strings.Shops.TransactionFailed, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
                 return;
             }
 
             if (slot < 0 || slot >= InShop.SellingItems.Count)
             {
-                PacketSender.SendChatMsg(this, Strings.Shops.error, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                PacketSender.SendChatMsg(this, Strings.Shops.TransactionFailed, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
                 return;
             }
 
@@ -3534,14 +3534,14 @@ namespace Intersect.Server.Entities
             var boughtItemBase = ItemBase.Get(boughtItem.ItemId);
             if (boughtItemBase == default)
             {
-                PacketSender.SendChatMsg(this, Strings.Shops.error, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                PacketSender.SendChatMsg(this, Strings.Shops.TransactionFailed, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
                 return;
             }
 
             var currencyBase = ItemBase.Get(boughtItem.CostItemId);
             if (currencyBase == default)
             {
-                PacketSender.SendChatMsg(this, Strings.Shops.error, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                PacketSender.SendChatMsg(this, Strings.Shops.TransactionFailed, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
                 return;
             }
 
@@ -3602,7 +3602,7 @@ namespace Intersect.Server.Entities
                     TryGiveItem(removedItem.Key, removedItem.Value);
                 }
 
-                PacketSender.SendChatMsg(this, Strings.Shops.error, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                PacketSender.SendChatMsg(this, Strings.Shops.TransactionFailed, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
                 return;
             }
 
--- a/Intersect.Server/Entities/Player.cs
+++ b/Intersect.Server/Entities/Player.cs
@@ -3562,7 +3562,11 @@ namespace Intersect.Server.Entities
             if (!CanGiveItem(boughtItemBase.Id, boughtItemAmount))
             {
                 PacketSender.SendChatMsg(this, Strings.Shops.inventoryfull, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                return;
             }
+
+            bool sucess = true;
+            Dictionary<Guid, int> removedItems = new Dictionary<Guid, int>();
 
             if (itemCostTotal > 0)
             {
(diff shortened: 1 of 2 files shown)


Intersect.Server/Entities/Player.cs
```diff
@@ -3518,88 +3518,76 @@ public void SellItem(int slot, int amount)
...
-                                }
-                            }
-                        }
-                        else
-                        {
-                            PacketSender.SendChatMsg(this, Strings.Shops.cantafford, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
-                        }
+            if (!CanGiveItem(boughtItemBase.Id, boughtItemAmount))
+            {
+                PacketSender.SendChatMsg(this, Strings.Shops.inventoryfull, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
```
lodicolo, line 3564: Shouldn't there be a `return;` after this?

Intersect.Server/Entities/Player.cs
```diff
@@ -3518,88 +3518,76 @@ public void SellItem(int slot, int amount)
...
+
+            if (itemCostTotal > 0)
+            {
+                var currencySlots = FindInventoryItemSlots(currencyBase.Id);
+                int remainingCost = itemCostTotal;
+
+                foreach (var itemSlot in currencySlots)
+                {
+                    int quantityToRemove = Math.Min(remainingCost, itemSlot.Quantity);
+                    TryTakeItem(itemSlot.ItemId, quantityToRemove);
```
lodicolo, line 3575: Honestly if TryTakeItem fails it needs to roll-back all of the items it already took, and it needs to abort the for loop and show a message. It should not assume that it succeeded.

Intersect.Server/Entities/Player.cs
```diff
@@ -3518,88 +3518,76 @@ public void SellItem(int slot, int amount)
...
+                    remainingCost -= quantityToRemove;
+
+                    if (remainingCost <= 0)
+                    {
+                        break;
                     }
                 }
             }
+
+            TryGiveItem(boughtItemBase.Id, boughtItemAmount);
```
lodicolo, line 3585: If this fails it needs to log that it failed to give the item, to what player ID, and how much was purchased.

Intersect.Server/Entities/Player.cs
lodicolo: It should also log what was successfully taken, what failed to be taken, what the quantityToRemove was when it failed, and the player ID it failed on.

lodicolo (changes requested):

Intersect.Server/Entities/Player.cs
lodicolo: This call should also use TryTakeItem(itemSlot) instead of itemSlot.ItemId

lodicolo (changes requested):

Intersect.Server/Entities/Player.cs
Arufonsu: @WeylonSantana , following what @lodicolo said, try something like this perhaps:

```csharp
if (itemCostTotal > 0)
{
    var currencySlots = FindInventoryItemSlots(currencyBase.Id);
    int remainingCost = itemCostTotal;
    bool success = true;

    foreach (var itemSlot in currencySlots)
    {
        int quantityToRemove = Math.Min(remainingCost, itemSlot.Quantity);
        if (TryTakeItem(itemSlot, quantityToRemove))
        {
            remainingCost -= quantityToRemove;
            Log.Warn($"Successfully took {quantityToRemove} items from slot {itemSlot} for player {playerId}.");
        }
        else
        {
            success = false;
            Log.Warn($"Failed to take items from slot {itemSlot} for player {playerId}. Quantity to remove: {quantityToRemove}.");
            break;
        }

        if (remainingCost <= 0)
        {
            break;
        }
    }

    if (!success)
    {
        // TODO: Rollback code goes here ?
        Log.Warn("Rolling back items due to failed transaction.");
    }
}
```

could *probably* do something similar with the giving item case:
 https://github.com/AscensionGameDev/Intersect-Engine/pull/1832/files#r1264234149
 
 PS. this is just an idea/example, remember use Localization/Strings.cs for the strings here!
 
 ie: 
```csharp
public readonly LocalizedString GiveItemSuccess = "Successfully took {0} items from slot {1} for player {2}.";
```

NikoCummings:
I tested the code changes and now if you try to buy more than what your inventory can hold  no gold is removed and no items are given. I consider this fix confirmed.

Intersect.Server/Entities/Player.cs
```diff
@@ -3562,8 +3562,12 @@ public void BuyItem(int slot, int amount)
             if (!CanGiveItem(boughtItemBase.Id, boughtItemAmount))
             {
                 PacketSender.SendChatMsg(this, Strings.Shops.inventoryfull, ChatMessageType.Inventory, CustomColors.Alerts.Error, Name);
+                return;
             }
 
+            bool sucess = true;
```
Arufonsu, line 3568: small typo:
use `success` instead

Arufonsu (dismissed):

Intersect.Server/Localization/Strings.cs
```diff
@@ -1356,6 +1356,9 @@ public sealed partial class ShopsNamespace : LocaleNamespace
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public readonly LocalizedString bound = @"This item is bound to you and cannot be sold!";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString error = @"Transaction failed!";
```
lodicolo, line 1360: Change this to `TransactionFailed` instead of `error`

lodicolo (changes requested):

Arufonsu (approved):

lodicolo (approved):

PR #1838, opened 2023-07-16
avoid blocking with shield packets flood

WeylonSantana:
I also forgot to check if it is blocking when it tries to block, so it floods the server with unnecessary packets


jcsnider:
Makes sense to me. Approving.

jcsnider (approved):

lodicolo (approved):

PR #1851, opened 2023-08-10
fix: (Day) 1850 - Fix issue where you could kick players from an inst…

### LINKED ISSUE #1850: bug: You can invite players to your party while in a shared instance
### Description

This is a bug that allows players to get over the party limit for a dungeon/otherwise shared instance. This is not desired behavior.

### Steps to Reproduce

1. Enter a dungeon with max party members
2. Kick a party member from your party, then invite another player
3. The kicked player will stay in the dungeon, but the new player will be allowed in to the dungeon.

### Version with bug

Most recent

None

### Affected platforms

No

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
…ance to get past the party size limit

Fixes #1850


Intersect.Server/Entities/Player.cs
```diff
@@ -4612,13 +4615,25 @@ public void CancelTrade()
...
+            }
+
             if (Party.Count != 0)
             {
                 PacketSender.SendChatMsg(fromPlayer, Strings.Parties.inparty.ToString(Name), ChatMessageType.Party, CustomColors.Alerts.Error);
 
                 return;
             }
 
+            if (Options.Instance.Instancing.BlockPartyInvitesInInstance && (!InOpenInstance || !fromPlayer.InOpenInstance))
```
lodicolo, line 4630: I think this should only take into account the inviter, but I have some other thoughts that I am going to include in the more general review comment.

lodicolo (changes requested):
I don't really agree with this implementation because I think it leaves some things out.

I think that the inviter should be able to invite the invitee no matter what except for in the case the invitee is already in a party.

What changes should be when a invitee is warped to the inviter's map, and I disagree with letting people get kicked from the instance if they get kicked from the party. This is to some extent me imposing my will on the developers who use the engine, but I am most concerned with giving the players a good experience and that never is. If a developer wants to be spiteful to their players, they are going to have to do more work than just changing a flag.

I think the correct fix is to:
- allow the developer to disable players from being warped into instances if they weren't already in the party before the instance was created, this should be default off
- record how many players have already entered the instance, this should be updated each time someone joins the instance and never decreased
- even if joining an active instance is permitted by the developer, if 4 people already joined the instance and the party size limit is 4, even if there are not currently 4 people in the party of 4 people in the instance, an invitee should not be warped into the instance, they should be left outside
- kicking from a party should not reduce the reward anyone in the instance receives, and should never kick someone from an instance because as you've noted (in Discord) this can be abused by malicious players
- players should always be allowed to invite to their party
- players should always be allowed to accept party invites, and if they are permitted to enter an in-progress instance (and the instance has not reached capacity) they will be given the option to warp into the in-progress instance
  - if the instance is full or in-progress joining is disabled the invitee will be informed that the instance has reached capacity and they can't be warped in

sainguinkrist:
> * kicking from a party should not reduce the reward anyone in the instance receives, and should never kick someone from an instance because as you've noted (in Discord) this can be abused by malicious players


I have strong disagreements to this. Nearly every MMO in existence does exactly this, and for very good reason. That one person that you now can't force out of your instance can be *just* as malicious.

So you're basically forcing 3 people to restart/leave the instance to make one happy, and I can almost guarantee that'll often be the case if this is how the system is setup.

I think you should look at how the vast majority of MMOs handle this exact mechanic.

lodicolo:
> the vast majority of MMOs handle this exact mechanic

The vast majority of MMOs do not allow inviting people into ongoing instances.

> It will be a regular source of complaints, I'd wager money on that.
There have been complaints about people abusing the existing systems for stealing group rewards, and I'm specifically looking at preventing the malicious party member being the one that has all the power.

Except in the case of vote-kicks, I don't think MMOs have put a ton of thought into how or when they kick people, and certainly not with written down rationale.

Really though, even vote-kicks can be manipulated to take advantage of players who are alone too. I wouldn't be opposed to adding this in, but everyone in the group would need to be eligible for vote-kicks and the threshold would need to be a configurable majority.

sainguinkrist:
> > the vast majority of MMOs handle this exact mechanic
> 
> The vast majority of MMOs do not allow inviting people into ongoing instances.
> 
> > It will be a regular source of complaints, I'd wager money on that.
> > There have been complaints about people abusing the existing systems for stealing group rewards, and I'm specifically looking at preventing the malicious party member being the one that has all the power.
> 
> Except in the case of vote-kicks, I don't think MMOs have put a ton of thought into how or when they kick people, and certainly not with written down rationale.
> 
> Really though, even vote-kicks can be manipulated to take advantage of players who are alone too. I wouldn't be opposed to adding this in, but everyone in the group would need to be eligible for vote-kicks and the threshold would need to be a configurable majority.

I was referring to how kicking works, not inviting into instances, though most MMOs I've played in recent years do allow that as well. ESO, WoW, and GW2 off the top of my head. That's neither here or there though.

I agree the most optimal system is a vote kick with free pick on who can be kicked, but I really do think in absence of that, it still shouldn't be a system that allows a single person to squat on group instances. I understand it's abusable, but I really do think that's the lesser of two evils in this situation.

lodicolo:
I'm willing to just merge this as is, but the problem is that MapInstanceType.Party doesn't exist, not sure if that's a MA:O thing and if Shared is good enough for the base engine (or should this change also add MapInstanceType.Party?)

lodicolo:
Drafting until the compilation error is resolved

PR #1855, opened 2023-08-12
Allow firing projectiles while facing events whos trigger is not 'Action Button'

### LINKED ISSUE #1852: bug: projectiles don't shoot when standing next to/facing events
### Description

If you are one tile away from a map event that is triggered by player collide and facing it. Shooting projectiles does not work.

### Steps to Reproduce

Equip a projectile weapon. Create a map event that is triggered with player collide. Stand next to it (the title touching it) and face the event. Try shooting the projectile

### Version with bug

Latest

Unknown 

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

jcsnider:
Resolves #1852

### DIFF
--- a/Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
+++ b/Intersect (Core)/Network/Packets/Server/EventEntityPacket.cs
@@ -1,4 +1,5 @@
-﻿using Intersect.GameObjects.Events;
+﻿using Intersect.Enums;
+using Intersect.GameObjects.Events;
 using MessagePack;
 
 namespace Intersect.Network.Packets.Server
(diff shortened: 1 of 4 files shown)


WeylonSantana (dismissed):

WeylonSantana:
I approved because I already made this fix before, I just forgot to open it

lodicolo (dismissed):

lodicolo:
Approved, but I wonder if just sending the Trigger down directly instead of a boolean is a better idea? That way the client is better informed if anyone decides to do anything for the other trigger types.

PR #1864, opened 2023-08-13
re-design editors windows

WeylonSantana:


Intersect.Editor/Localization/Strings.cs
```diff
@@ -3749,6 +3753,8 @@ public partial struct ItemEditor
 
             public static LocalizedString SingleUseEvent = @"Destroy On Use?";
 
+            public static LocalizedString stackoptions = @"Stackable?";
```
lodicolo, line 3756: Since you are adding this it should be in PascalCase but the name would be better also as `Stackable`.

lodicolo (changes requested):

Intersect.Editor/Localization/Strings.cs
```diff
@@ -3753,7 +3753,7 @@ public partial struct ItemEditor
 
             public static LocalizedString SingleUseEvent = @"Destroy On Use?";
 
-            public static LocalizedString stackoptions = @"Stackable?";
+            public static LocalizedString StackOptions = @"Stackable";
 
             public static LocalizedString stackable = @"Stackable?";
```
lodicolo, line 3758: Duplicate? Just use the original `stackable`

lodicolo (changes requested):

WeylonSantana:
done
image

lodicolo (approved):

PR #1871, opened 2023-08-19
chore - Initial Intersect.Core clean up

WeylonSantana:
An initial cleanup of the Intersect.Core files

- Removing empty lines
- Removal of unused libraries
- Small tweaks to some variable/method definitions

I'll leave the cleaning of Core.Networking for the next pr


Intersect (Core)/GameObjects/LightBase.cs
```diff
@@ -68,7 +68,11 @@ Color color
         /// <returns></returns>
         public override int GetHashCode()
         {
-            return Intensity ^ (int)Expand ^ Color?.A ?? 0 ^ Color?.R ?? 0 ^ Color?.G ?? 0 ^ Color?.B ?? 0;
+            return Intensity ^ (int)Expand ^ Color?.A
+                ?? 0 ^ Color?.R
+                ?? 0 ^ Color?.G
+                ?? 0 ^ Color?.B
+                ?? 0;
```
lodicolo, line 75: This should be reverted or the `Color?.<X> ?? 0` segments need to have parentheses for readability and forced onto 1 line per parenthesized group.

That or honestly, this could change to `Intensity ^ (int)Expand ^ (Color?.GetHashCode() ?? 0)`

Intersect (Core)/GameObjects/LightBase.cs
```diff
@@ -79,7 +83,9 @@ public override int GetHashCode()
         public override bool Equals(object obj)
         {
             if (obj is LightBase light)
```
lodicolo, line 85: if (...) { ... }, no {} is bad -- also the return statement really doesn't need to be multiple lines, idk why your formatter split it like that

lodicolo (changes requested):

Intersect (Core)/GameObjects/LightBase.cs
```diff
@@ -79,7 +83,9 @@ public override int GetHashCode()
...
-                return Intensity == light.Intensity && Color == light.Color && Expand == light.Expand;
+                return Intensity == light.Intensity
+                    && Color == light.Color
+                    && Expand == light.Expand;
```
lodicolo, line 88: Change to one line, surround with {}

lodicolo (approved):

PR #1876, opened 2023-08-27
fix: window -> world coords when zoomed

lodicolo:


Arufonsu (approved):
👍🏽 

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/799ed32f-7b11-4531-a68b-03d250306946

PR #1888, opened 2023-08-31
fix: (Day) 1887 - Fixes issues with zoom out at >32px tile size

### LINKED ISSUE #1887: bug: Zooming out fails to render properly when tilesize >32px
### Description

Zooming out doesn't work correctly at tile sizes of greater than 32px wide/tall. This is because a few optimizations need updated on the client to reflect the client's world zoom.

### Steps to Reproduce

1. Make a map with tile size > 32px
2. Zoom out in the client on that map to see weird issues

### Version with bug

Latest

Pre-zoom

### Affected platforms

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
Fixes #1887

### DIFF
--- a/Intersect.Client/Core/Graphics.cs
+++ b/Intersect.Client/Core/Graphics.cs
@@ -1013,7 +1013,7 @@ namespace Intersect.Client.Core
                 return;
             }
 
-            var destRect = new FloatRect(new Pointf(), new Pointf(sDarknessTexture.GetWidth() / Globals.Database.WorldZoom, sDarknessTexture.GetHeight() / Globals.Database.WorldZoom));
+            var destRect = new FloatRect(new Pointf(), sDarknessTexture.Dimensions / Globals.Database.WorldZoom);
             if (map.IsIndoors)
             {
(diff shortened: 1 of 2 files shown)


Intersect.Client/Core/Graphics.cs
```diff
@@ -41,6 +41,8 @@ public static FloatRect CurrentView
                 Renderer.SetView(_currentView);
             }
         }
+        
+        public static FloatRect Viewport => new FloatRect(CurrentView.Position, CurrentView.Size / (Globals.Database?.WorldZoom ?? 1));
```
lodicolo, line 45: Can you change this to WorldViewport so this isn't confused when UI Scaling is eventually introduced?

lodicolo (changes requested):

Intersect.Client/Core/Graphics.cs
AlexVild: Done

lodicolo (approved):

PR #1889, opened 2023-09-02
fix: account for deleted player in rank/removal guild logic

lodicolo:


lodicolo:
Tested by Naz https://discord.com/channels/363106200243535872/965617981159788614/1147619436937879612

PR #1899, opened 2023-09-10
fix: entityExperienceToNextLevel as Long within EntityBox 

### LINKED ISSUE #1898: bug: ExpToNextLevel using Float instead of Long
### Description

if the value of experience to next level is over the limit of `int32`, which is `2,147,483,647`, the following happens:

imagen

Originally reported by Naziozeno:
imagen

### Steps to Reproduce

1.- Create a class with an exp table that increases a lot per level up
2.- Level up until the exp value to the next level requirement goes past `2,147,483,647`
3.- See the issue on exp bar, either with % or by using values

### Version with bug

v0.7.2-beta.49

Never did

### Affected platforms

Windows, Linux

Yes, will push a PR with the fix

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
Should resolve #1898 

Preview:

imagen

imagen

imagen

imagen

### DIFF
--- a/Intersect.Client/Interface/Game/EntityPanel/EntityBox.cs
+++ b/Intersect.Client/Interface/Game/EntityPanel/EntityBox.cs
@@ -801,7 +801,7 @@ namespace Intersect.Client.Interface.Game.EntityPanel
             float targetExpSize;
             var barDirectionSetting = ClientConfiguration.Instance.EntityBarDirections[(int)Vital.VitalCount];
             var barPercentageSetting = Globals.Database.ShowExperienceAsPercentage;
-            var entityExperienceToNextLevel = (float)((Player)MyEntity).GetNextLevelExperience();
+            var entityExperienceToNextLevel = ((Player)MyEntity).GetNextLevelExperience();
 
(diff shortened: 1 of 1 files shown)


Intersect.Client/Interface/Game/EntityPanel/EntityBox.cs
```diff
@@ -801,7 +801,7 @@ private void UpdateXpBar(float elapsedTime, bool instant = false)
             float targetExpSize;
             var barDirectionSetting = ClientConfiguration.Instance.EntityBarDirections[(int)Vital.VitalCount];
             var barPercentageSetting = Globals.Database.ShowExperienceAsPercentage;
-            var entityExperienceToNextLevel = (float)((Player)MyEntity).GetNextLevelExperience();
+            var entityExperienceToNextLevel = ((Player)MyEntity).GetNextLevelExperience();
```
lodicolo, line 804: This should be the only line needed for this PR. The rest of the PR is unnecessary.

lodicolo (changes requested):

lodicolo (approved):

PR #1903, opened 2023-09-15
enhancement: bank withdraw behavior

Arufonsu:
- Implements the 'FindAvailableInventorySpaceForItem' method to ensure sliders display the correct values for available inventory space.
- Instead of blindly opening the item withdraw slider with a max quantity of 0 when there's no space left in the inventory (pointless), we now inform the player: "There is no space left in your inventory for that item!"

Before:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/6646df47-a3f5-4ae0-b3bb-e8841f249918

After:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/a00f6826-54ae-4d86-970c-82ff492c7668

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -810,21 +810,20 @@ namespace Intersect.Client.Entities
                 }
             }
 
-            var quantity = bankSlot.Quantity;
-            var maxQuantity = quantity;
+            var itemQuantityInBank = bankSlot.Quantity;
+            int availableSpace = FindAvailableInventorySpaceForItem(itemDescriptor.Id);
 
-            if (itemDescriptor.IsStackable)
+            if (availableSpace < 1)
             {
-                maxQuantity = GetQuantityOfItemInBank(itemDescriptor.Id);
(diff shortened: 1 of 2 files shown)


Intersect.Client/Entities/Player.cs
```diff
@@ -782,35 +782,56 @@ public void TryWithdrawItem(int bankSlotIndex, int inventorySlotIndex = -1)
...
+
+            // Calculate the total available space in the player's inventory for this item.
+            for (int i = 0; i < Inventory.Length; i++)
+            {
+                var inventorySlot = Inventory[i];
+                if (inventorySlot?.ItemId == itemId)
+                {
+                    spaceLeft += maxInventoryStack - inventorySlot.Quantity;
+                }
+                else if (inventorySlot?.ItemId == Guid.Empty)
```
lodicolo, line 826: You can't do this because `default(Guid)` is `Guid.Empty` but default of `inventorySlot?.ItemId` is `null`

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: Alright 👍🏽 will be back on this PR by the end of the week, been kinda busy lately !

PR #1905, opened 2023-09-16
enh: Add timer to remove stale cooldowns during playtime

### LINKED ISSUE #1904: enhancement: Improve Cooldown Database/Memory Storage
### Description

Clear stale cooldown timers while a user is logged in, not just when a user logs out.

### Intended Use-Case

At the moment, cooldowns are stored on the player indefinitely as long as the player is online and active.
This means a lot of useless data is being sent back and forth for players even if the cooldown expiration timer is long since past.
At logout these stale timers are cleared, but not during runtime meaning this collection can grow to be quite significant over long play sessions.

(Partially because I did a dum dum and made Global Cooldowns simply put EVERYTHING on cooldown, will have to rectify that in the future)

### Duplicate Check

Cheshire92:
Resolves #1904 

Adds a timer to remove stale cooldowns while a player is logged in.
Instead of duplicating the removal code between logout and runtime removal, I turned removal of spell and item cooldowns into a re-usable method.

Won't solve major issues at the moment, but will at least stop sending endless stale records around constantly for the time being.

### DIFF
--- a/Intersect (Core)/Config/ProcessingOptions.cs
+++ b/Intersect (Core)/Config/ProcessingOptions.cs
@@ -103,6 +103,11 @@ namespace Intersect.Config
         /// </summary>
         public int CommonEventAutorunStartInterval { get; set; } = 500;
 
+        /// <summary>
+        /// How often should the server check for stale cooldowns on players to remove them from their respective collections.
+        /// </summary>
+        public int StaleCooldownRemovalTimer { get; set; } = 60000;
+
         [OnDeserialized]
         internal void OnDeserializedMethod(StreamingContext context)
         {
--- a/Intersect (Core)/Config/ProcessingOptions.cs
+++ b/Intersect (Core)/Config/ProcessingOptions.cs
@@ -175,6 +175,11 @@ namespace Intersect.Config
             {
                 throw new InvalidOperationException("Player save interval is too low and would cause performance issues, consider raising.");
             }
+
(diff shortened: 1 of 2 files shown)


Intersect (Core)/Config/ProcessingOptions.cs
```diff
@@ -103,6 +103,11 @@ public partial class ProcessingOptions
         /// </summary>
         public int CommonEventAutorunStartInterval { get; set; } = 500;
 
+        /// <summary>
+        /// How often should the server check for stale cooldowns on players to remove them from their respective collections.
+        /// </summary>
+        public int StaleCooldownRemovalTimer { get; set; } = 60000;
```
lodicolo, line 109: Can you add code to the Validate method to make sure this value is at least 100 so this doesn't run every tick?

lodicolo (changes requested):

Intersect (Core)/Config/ProcessingOptions.cs
Cheshire92: Would it be fine to have it throw an InvalidOperationException like the rest?
If so, what would a good error description be?

Cheshire92: Settled on "StaleCooldownRemovalTimer is too low, value should be at least 100ms to avoid running this check every server tick." on Discord.

lodicolo (dismissed):

lodicolo (approved):

PR #1940, opened 2023-10-13
fix: properly handle NPCs directions near map borders

### LINKED ISSUE #1875: bug: NPCs stop attacking the player when positioned at the edge of a map border
### Description

When an NPC is at the edge of a map border, and has aggroed a player that crosses the map border, they will stop attacking but still be attackable (ie they won't reset). It's as if they can no longer see the player directly on the edge of maps.

If you cross over one more tile into the map, and the NPC follows you, they resume attacking.

This is a pretty exploitable way to farm NPCs.

### Steps to Reproduce

1. Create 2 maps, create an NPC
2. Attack and pull the NPC to the exact map border, and position yourself directly on the other side
3. NPC should stop attacking but should still be attackable
4. ???
5. Exploit

Expected result: NPC should continue attacking players on borders
Actual result: NPC becomes an easily exploitable kill

### Version with bug

0.7.2-beta.22

Unknown

### Affected platforms

No workaround found yet.

I've tried messing with reset radius and combat reset time, but neither have any effect if you just keep attacking.

### Relevant log output

```shell
No logs for this issue.
```

### Duplicate Bug Check

Arufonsu:
- Should resolve #1875

Preview:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/fa88b328-3307-4ede-8d7a-9f053155706b

### DIFF
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -2735,8 +2735,11 @@ namespace Intersect.Server.Entities
                 return Dir;
             }
 
-            var originMapController = MapController.Get(MapId);
-            var targetMapController = MapController.Get(en.MapId);
+            if (!MapController.TryGet(MapId, out var originMapController) ||
+                !MapController.TryGet(en.MapId, out var targetMapController))
+            {
+                return Dir;
+            }
 
             var originY = Y + originMapController.MapGridY * Options.MapHeight;
             var originX = X + originMapController.MapGridX * Options.MapWidth;
--- a/Intersect.Server/Entities/Entity.cs
+++ b/Intersect.Server/Entities/Entity.cs
@@ -2769,20 +2769,22 @@ namespace Intersect.Server.Entities
                 return false;
             }
 
-            var x1 = X + MapController.Get(MapId).MapGridX * Options.MapWidth;
-            var y1 = Y + MapController.Get(MapId).MapGridY * Options.MapHeight;
-            var x2 = x + MapController.Get(mapId).MapGridX * Options.MapWidth;
-            var y2 = y + MapController.Get(mapId).MapGridY * Options.MapHeight;
+            var originMapController = MapController.Get(MapId);
+            var targetMapController = MapController.Get(mapId);
 
-            var dx = Math.Abs(x1 - x2);
-            var dy = Math.Abs(y1 - y2);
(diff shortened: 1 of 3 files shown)


Intersect.Server/Entities/Entity.cs
```diff
@@ -2720,41 +2728,21 @@ public virtual void NotifySwarm(Entity attacker)
...
-                if (en.Map.MapGridY < Map.MapGridY)
-                {
-                    originY += Options.MapHeight - 1;
-                }
-                else if (en.Map.MapGridY > Map.MapGridY)
-                {
-                    targetY += Options.MapHeight - 1;
-                }
+            var originMapController = MapController.Get(MapId);
+            var targetMapController = MapController.Get(en.MapId);
```
lodicolo, line 2739: Can you use TryGet and if it fails just return Dir?

Intersect.Server/Entities/Entity.cs
```diff
@@ -2789,20 +2777,22 @@ protected bool IsOneBlockAway(Guid mapId, int x, int y, int z = 0)
                 return false;
             }
 
-            var x1 = X + MapController.Get(MapId).MapGridX * Options.MapWidth;
-            var y1 = Y + MapController.Get(MapId).MapGridY * Options.MapHeight;
-            var x2 = x + MapController.Get(mapId).MapGridX * Options.MapWidth;
-            var y2 = y + MapController.Get(mapId).MapGridY * Options.MapHeight;
+            var originMapController = MapController.Get(MapId);
+            var targetMapController = MapController.Get(mapId);
```
lodicolo, line 2781: Same thing with TryGet and returning the default value if either of them fail

lodicolo (changes requested):

Intersect.Server/Entities/Entity.cs
```diff
@@ -2645,20 +2645,28 @@ protected bool IsFacingTarget(Entity target)
                 return false;
             }
 
-            var dx = target.X - X;
-            var dy = target.Y - Y;
+            var originMapController = MapController.Get(MapId);
+            var targetMapController = MapController.Get(target.MapId);
```
lodicolo, line 2649: Almost missed this -- also use TryGet and return early if either fail

lodicolo (changes requested):

lodicolo (approved):

PR #1990, opened 2023-10-20
enhancement: don't drop items in arena map.

theceltos:
I think many project owners use this type of map for some pvp system or battle in general, I'm sending this suggestion that will protect the player's item drops when dying on this type of map.

### DIFF
--- a/Intersect (Core)/Config/MapOptions.cs
+++ b/Intersect (Core)/Config/MapOptions.cs
@@ -21,6 +21,10 @@ namespace Intersect.Config
                 MovementDirections = mEnableDiagonalMovement ? 8 : 4;
             }
         }
+        /// <summary>
+        /// option to drop items on arena type maps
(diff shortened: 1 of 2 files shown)


lodicolo (changes requested):
I'm not opposed to adding this, however it needs to be configurable.

At minimum it should be added as a server configuration in MapOptions, something like `DisablePlayerDropsInArenaMaps` and the default value would be false.

A check to see if this is true would be added to your if statement.

Also please add braces for the if statement, and a blank line for readability, e.g.

```csharp
if (…)
{
    return;
}

…Old code…
```

theceltos:
I think I did everything you asked, can you check?

lodicolo:
@theceltos I had to make one minor change because you had put `= true` which is assignment (overwriting the configuration and making it always true) instead of comparison, but it's good now. Will merge.

PR #1992, opened 2023-10-21
refactoring: bank

Arufonsu:
- Should resolve https://github.com/AscensionGameDev/Intersect-Engine/issues/1681
- Implements _updated_ changes from #1903
- Implements _updated_ changes from #1720

Preview:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/2c7ee9df-20ce-4fe5-9628-ab5134d25cdf

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -759,6 +759,7 @@ namespace Intersect.Client.Entities
                 return;
             }
 
+            // TODO: Unify depositing single/multiple items.
             if (itemQuantityInInventory < 2)
             {
                 DepositSingleItem(inventorySlotIndex, bankSlotIndex);
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -796,19 +796,29 @@ namespace Intersect.Client.Entities
 
         private static int FindAvailableBankSpaceForItem(Guid itemId, int itemQuantityInInventory)
         {
+            if (!ItemBase.TryGet(itemId, out var itemDescriptor))
+            {
+                return 0;
+            }
+
             int spaceLeft = 0;
-            int maxBankStack = ItemBase.Get(itemId).MaxBankStack;
+            int maxBankStack = itemDescriptor.MaxBankStack;
+
+            if (spaceLeft >= itemQuantityInInventory)
+            {
+                return itemQuantityInInventory;
+            }
 
             // Calculate the total available space in the bank for the item
-            for (int i = 0; i < Globals.BankSlots; i++)
+            var bankSlots = Globals.Bank.ToArray();
+            foreach (var bankSlot in bankSlots)
             {
-                var bankItem = Globals.Bank[i];
-                if (bankItem != null && bankItem.ItemId == itemId)
+                if (bankSlot != null && bankSlot.ItemId == itemId)
                 {
-                    int quantityLeft = maxBankStack - bankItem.Quantity;
+                    int quantityLeft = maxBankStack - bankSlot.Quantity;
                     spaceLeft += quantityLeft;
                 }
-                else if (bankItem == null || bankItem.ItemId == Guid.Empty)
+                else if (bankSlot == null || bankSlot.ItemId == Guid.Empty)
                 {
                     spaceLeft += maxBankStack;
                 }
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -847,8 +847,8 @@ namespace Intersect.Client.Entities
                 return;
             }
 
-            var itemQuantityInBank = bankSlot.Quantity;
-            int availableSpace = FindAvailableInventorySpaceForItem(itemDescriptor.Id);
+            var itemQuantityInBank = GetQuantityOfItemInBank(itemDescriptor.Id);
+            int availableSpace = FindAvailableInventorySpaceForItem(itemDescriptor.Id, itemQuantityInBank);
 
             if (availableSpace < 1)
             {
@@ -856,7 +856,7 @@ namespace Intersect.Client.Entities
                 return;
             }
 
-            if (itemQuantityInBank < 2)
+            if (itemQuantityInBank == 1 || availableSpace == 1)
             {
                 WithdrawSingleItem(bankSlotIndex, inventorySlotIndex);
                 return;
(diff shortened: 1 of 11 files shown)


lodicolo:
@Arufonsu 

1. Why can you not withdraw multiple equipment instances when you are able to deposit multiple?
2. I only see context menu interactions, are you able to use double click to deposit/withdraw a single equipment instance/stack of items?

Arufonsu:
> @Arufonsu
> 
>     1. Why can you not withdraw multiple equipment instances when you are able to deposit multiple?
> 
>     2. I only see context menu interactions, are you able to use double click to deposit/withdraw a single equipment instance/stack of items?

1. This would be new - already sorted it out, will commit in a bit
2. I'll upload a video with that, forgot to show that off again in the new video

Intersect.Client/Entities/Player.cs
```diff
@@ -742,37 +742,78 @@ public void TryDepositItem(int inventorySlotIndex, int bankSlotIndex = -1)
...
-            );
+                userData: new[] { inventorySlotIndex, bankSlotIndex },
+                quantity: availableSpace,
+                maxQuantity: availableSpace);
+        }
+
+        private static int FindAvailableBankSpaceForItem(Guid itemId, int itemQuantityInInventory)
+        {
+            int spaceLeft = 0;
+            int maxBankStack = ItemBase.Get(itemId).MaxBankStack;
```
lodicolo, line 799: There's no null safety here. If the item descriptor is not retrievable, return 0 immediately, even before allocating `spaceLeft`.

i.e.

```csharp
if (!ItemBase.TryGet(itemId, out var itemDescriptor))
{
    return 0;
}
var maxBankStack = itemDescriptor.MaxBankStack;
var spaceLeft = 0;
```

Intersect.Client/Entities/Player.cs
```diff
@@ -799,46 +840,79 @@ public void TryWithdrawItem(int bankSlotIndex, int inventorySlotIndex = -1)
...
+                    spaceLeft += maxInventoryStack - inventorySlot.Quantity;
+                }
+                else if (inventorySlot == null || inventorySlot.ItemId == Guid.Empty)
+                {
+                    spaceLeft += maxInventoryStack;
+                }
+            }
+
+            return spaceLeft;
+        }
```
lodicolo, line 914: All of my comments re: scanning bank slots apply here.

1. Null-safety on `ItemBase.Get` (use `TryGet`)
2. `ToArray()` `Inventory` into a local variable to protect against race conditions
3. `foreach` over that local array
4. return as soon as you know that it exceeds the number that are even available in the bank (I do realize you don't even have a parameter for this yet but you said you are adding this in a bit)

Intersect.Client/Entities/Player.cs
```diff
@@ -799,46 +840,79 @@ public void TryWithdrawItem(int bankSlotIndex, int inventorySlotIndex = -1)
...
-            if (maxQuantity < 2)
+            if (itemQuantityInBank < 2)
             {
-                PacketSender.SendWithdrawItem(bankSlotIndex, 1, inventorySlotIndex);
+                WithdrawSingleItem(bankSlotIndex, inventorySlotIndex);
                 return;
             }
 
-            int[] userData = new int[2] { bankSlotIndex, inventorySlotIndex };
+            OpenWithdrawItemInputBox(itemDescriptor, bankSlotIndex, inventorySlotIndex, availableSpace);
```
lodicolo, line 865: Will review after your next commit.

Ideally this should be the exact same as the `TryDeposit`, just swapping the words "bank" and "inventory" to make it go the other direction.

Eventually that logic should also be unified (and when it does it should be unified with the server implementation, we do not need multiple implementations of inventory scanning/transferring in the engine) -- but _don't_ try to make a PR doing that yet.

lodicolo (changes requested):
I'll review the server after your next commit

Arufonsu:
https://github.com/AscensionGameDev/Intersect-Engine/pull/1992/commits/c930ebec94371d9a70bab1492dfb5784b6a68d2e

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/f22a83b1-5573-493a-a3a4-c00a02a66471

Intersect.Client/Entities/Player.cs
Arufonsu: Made sure that both are equally the same in terms of the logic used, will not address the unifying process for now

lodicolo:
@Arufonsu It looks to me like double left clicking on something that has multiple slots brings up the quantity selector?

I think that either it should just move one stack, or when it pops up the quantity selector that the default amount should be the quantity of items in that slot. This is to reduce user clicks

When a user right clicks and uses the context menu to withdraw, it should still pop up a quantity selector with the default of maximum.

Intersect.Client/Entities/Player.cs
```diff
@@ -733,33 +733,95 @@ private void SellInputBoxOkay(object sender, EventArgs e)
...
+        /// <summary>
+        /// Tries to deposit items into the bank.
+        /// </summary>
+        /// <param name="inventorySlotIndex">The slot of the item in the player's inventory.</param>
+        /// <param name="bankSlotIndex">The slot of the bank to deposit the item into.</param>
+        /// <param name="contextMenu">Indicates if the action was triggered by a context menu.</param>
+        /// <param name="shiftKey">Indicates if the action was triggered by the shift key</param>
+        public void TryDepositItem(int inventorySlotIndex,
+            int bankSlotIndex = -1,
+            bool contextMenu = false,
```
lodicolo, line 745: This should have been quantityHint as I said to do on Discord

Intersect.Client/Entities/Player.cs
```diff
@@ -733,33 +733,95 @@ private void SellInputBoxOkay(object sender, EventArgs e)
...
+            bool shiftKey = false)
```
lodicolo, line 746: This shoud have been called prompt as I said on Discord, the name of the key does not convey what the parameter is for

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
Arufonsu: shiftKey is used for the alternative shift + double click logic !

lodicolo: Ok but the functionality is showing or not showing a prompt which is relevant in this part of the code.

The key to do it is irrelevant at this level of abstraction, and when the client and server inventory abstractions are merged it will be even less relevant to the code.

Arufonsu (commented):
just tested out @lodicolo 's commits and in terms of the expected behavior it's all good ! :

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/e669b78e-a0c8-4500-83cc-6f8af7e0ef69

PR #1995, opened 2023-10-23
enhancement: don't loss exp in arena map

theceltos:
As the title already explains, this will allow you to choose between losing experience on maps such as arena or not.


AlefMichel:
That would be really nice, I thought about it some time ago.

lodicolo (approved):

PR #2065, opened 2023-11-12
feat: (Day) Added castSpellOn event command, to Aes' chagrin

AlexVild:


Intersect.Server.Core/Entities/Events/CommandProcessing.cs
```diff
@@ -1546,6 +1546,48 @@ Stack<CommandInstance> callStack
...
+            }
+
+            if (command.PartyMembers && player.IsInParty)
+            {
+                affectedPlayers.AddRange(player.Party.Where(pl => pl.Id != player.Id));
+            }
+
+            if (command.GuildMembers && player.IsInGuild)
+            {
+                affectedPlayers.AddRange(player.Guild.FindOnlineMembers().Where(pl => pl.Id != player.Id));
```
lodicolo, line 1577: `DistinctBy` is necessary for the overlap between the party/guild, so you don't need to filter out `player.Id` here since it will be taken care of by `DistinctBy`

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -254,6 +256,9 @@ public MapBase SharedInstanceRespawn
 
         private long mGlobalCooldownTimer;
 
+        [NotMapped, JsonIgnore]
+        public bool IsInParty => Party != null && Party.Count >= 2;
```
lodicolo, line 260: `Party.Count > 1`

Intersect.Server/Core/CommandParsing/CommandParser.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
lodicolo, line 1: Revert changes to this file, looks like BOM

lodicolo (changes requested):

Intersect.Server.Core/Entities/Events/CommandProcessing.cs
AlexVild: I have this here in the (unlikely) event someone wants to perform a spell on guild members but _not_ themselves or their party. I think it's still necessary

Intersect.Server.Core/Entities/Player.cs
AlexVild: Done

Intersect.Server/Core/CommandParsing/CommandParser.cs
AlexVild: Done

Intersect.Server.Core/Entities/Events/CommandProcessing.cs
lodicolo: Oh yeah, good point.

lodicolo (approved):

PR #2073, opened 2023-11-13
feat: (Day) Adds custom player name colors for in-party/in-guild/hostile

AlexVild:
Default hostile color:
image

Default party color:
image

Default guild color:
image

### DIFF
--- a/Intersect (Core)/CustomColors.cs
+++ b/Intersect (Core)/CustomColors.cs
@@ -82,8 +82,8 @@ namespace Intersect
             public Dictionary<string, LabelColor> Players = new Dictionary<string, LabelColor>()
             {
                 {"Normal", new LabelColor(new Color(255, 255, 255, 255), Color.Black, new Color(180, 0, 0, 0))},
-                {"PartyMember", new LabelColor(new Color(255, 70, 192, 219), Color.Black, new Color(180, 0, 0, 0))},
-                {"GuildMember", new LabelColor(new Color(255, 5, 179, 26), Color.Black, new Color(180, 0, 0, 0))},
+                {"Party", new LabelColor(new Color(255, 70, 192, 219), Color.Black, new Color(180, 0, 0, 0))},
+                {"Guild", new LabelColor(new Color(255, 5, 179, 26), Color.Black, new Color(180, 0, 0, 0))},
                 {"Hostile", new LabelColor(new Color(255, 255, 81, 0), Color.Black, new Color(180, 0, 0, 0))},
                 {"Moderator", new LabelColor(new Color(255, 0, 255, 255), Color.Black, new Color(180, 0, 0, 0))},
                 {"Admin", new LabelColor(new Color(255, 255, 255, 0), Color.Black, new Color(180, 0, 0, 0))},
--- a/Intersect (Core)/CustomColors.cs
+++ b/Intersect (Core)/CustomColors.cs
@@ -82,6 +82,9 @@ namespace Intersect
             public Dictionary<string, LabelColor> Players = new Dictionary<string, LabelColor>()
             {
                 {"Normal", new LabelColor(new Color(255, 255, 255, 255), Color.Black, new Color(180, 0, 0, 0))},
+                {"PartyMember", new LabelColor(new Color(255, 70, 192, 219), Color.Black, new Color(180, 0, 0, 0))},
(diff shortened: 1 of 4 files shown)


Intersect (Core)/CustomColors.cs
```diff
@@ -82,6 +82,9 @@ public sealed partial class NamesNamespace
             public Dictionary<string, LabelColor> Players = new Dictionary<string, LabelColor>()
             {
                 {"Normal", new LabelColor(new Color(255, 255, 255, 255), Color.Black, new Color(180, 0, 0, 0))},
+                {"PartyMember", new LabelColor(new Color(255, 70, 192, 219), Color.Black, new Color(180, 0, 0, 0))},
+                {"GuildMember", new LabelColor(new Color(255, 5, 179, 26), Color.Black, new Color(180, 0, 0, 0))},
```
lodicolo, line 86: can you change these to just Party and Guild

lodicolo: reason: I want to take advantage of `nameof()` wherever possible in the future and I don't see us ever creating types called `PartyMember` and `GuildMember`

Intersect.Client/Entities/Player.cs
```diff
@@ -2406,9 +2409,45 @@ public override void DrawName(Color textColor, Color borderColor, Color backgrou
...
+                // Guildies
+                else if (Globals.Me.IsInGuild && Guild == Globals.Me.Guild && CustomColors.Names.Players.TryGetValue("GuildMember", out var guildColors))
+                {
+                    textColor = guildColors.Name;
+                    borderColor = guildColors.Outline;
+                    backgroundColor = guildColors.Background;
+                }
+
+                // Enemies in PvP
+                if (!Globals.Me.IsAllyOf(this) && Globals.Me.MapInstance.ZoneType != MapZone.Safe && CustomColors.Names.Players.TryGetValue("Hostile", out var hostileColors))
```
lodicolo, line 2430: Is there a reason this isn't `else if`d?

Intersect.Client/Entities/Player.cs
```diff
@@ -2406,9 +2409,45 @@ public override void DrawName(Color textColor, Color borderColor, Color backgrou
...
         }
 
+        public override bool IsAllyOf(Player en)
+        {
+            if (base.IsAllyOf(en)) 
+            {
+                return true;
+            }
+
+            return Globals.Me.IsInMyParty(en) || Globals.Me.IsInMyGuild(en) || en.MapInstance.ZoneType == MapZone.Safe;
```
lodicolo, line 2448: Shouldn't the first two conditions be `IsInMyParty()` and `IsInMyGuild()` without the `Globals.Me.` qualification?

The `IsAllyOf()` method is being misused if the `Globals.Me.` qualification.

Intersect.Client/Entities/Entity.cs
```diff
@@ -489,6 +489,34 @@ public void AddAnimations(List<AnimationBase> anims)
...
+            }
+
+            // Events ONLY have allies
+            if (Type == EntityType.Event)
+            {
+                return true;
+            }
+
+            // Yourself is always an ally
+            if (en.Id == Globals.Me.Id)
```
lodicolo, line 512: I think this method should be comparing `en.Id` to `Id` not `Globals.Me.Id`. The `IsAllyOf()` method is being misused if the `Globals.Me.` qualification.

lodicolo (changes requested):

Intersect.Client/Entities/Player.cs
AlexVild: My reasoning is that in the _future_, if there are additional rules about what constitutes an ally (i.e, maybe you can kill guildies in an Arena map), then we won't miss this. Basically, if someone is hostile, their name color will _always_ be overridden.

Intersect.Client/Entities/Entity.cs
AlexVild: Good catch, fixed

Intersect (Core)/CustomColors.cs
AlexVild: Good catch, fixed

Intersect.Client/Entities/Player.cs
AlexVild: Good catch, fixed

lodicolo (approved):

PR #2075, opened 2023-11-13
feat: (Day) Add ability to enable self-cast when casting friendly spe…

AlexVild:
…lls on hostile

Basically, if you are targeting a hostile with this option enabled, and you cast a friendly spell, the engine will try to cast the spell on your self instead. This makes combat more fluid and makes it easier to panic heal/buff yourself, or maintain buffs

### DIFF
--- a/Intersect (Core)/Config/CombatOptions.cs
+++ b/Intersect (Core)/Config/CombatOptions.cs
@@ -72,6 +72,6 @@
         /// <summary>
         /// If enabled, this makes it so a player casting a friendly spell on a hostile target instead casts the spell upon themselves
         /// </summary>
-        public bool EnableAutoSelfCastOnFriendlyHostile = false;
+        public bool EnableAutoSelfCastOnFriendlyHostile { get; set; } = false;
     }
 }
--- a/Intersect (Core)/Config/CombatOptions.cs
+++ b/Intersect (Core)/Config/CombatOptions.cs
@@ -72,6 +72,6 @@
         /// <summary>
         /// If enabled, this makes it so a player casting a friendly spell on a hostile target instead casts the spell upon themselves
         /// </summary>
-        public bool EnableAutoSelfCastOnFriendlyHostile { get; set; } = false;
+        public bool EnableAutoSelfCastFriendlySpellsWhenTargetingHostile { get; set; } = false;
     }
 }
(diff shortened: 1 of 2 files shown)


Intersect (Core)/Config/CombatOptions.cs
```diff
@@ -68,5 +68,10 @@ public partial class CombatOptions
         /// If enabled, this allows entities to turn around while casting
         /// </summary>
         public bool EnableTurnAroundWhileCasting = false;
+
+        /// <summary>
+        /// If enabled, this makes it so a player casting a friendly spell on a hostile target instead casts the spell upon themselves
+        /// </summary>
+        public bool EnableAutoSelfCastOnFriendlyHostile = false;
```
lodicolo, line 75: Make this a property please

lodicolo (changes requested):
Wouldn't this PR break being able to cast something like heal on an undead (which would damage it) since it would always force self-target? I think this is just going to be misused by a dev who doesn't read the documentation and think about the decision they're making.

AlexVild:
> Wouldn't this PR break being able to cast something like heal on an undead (which would damage it) since it would always force self-target? I think this is just going to be misused by a dev who doesn't read the documentation and think about the decision they're making.

We don't have anything like that (i.e elemental damage or modifiers on NPCs like "undead").

Intersect (Core)/Config/CombatOptions.cs
AlexVild: Done

Intersect (Core)/Config/CombatOptions.cs
```diff
@@ -68,5 +68,10 @@ public partial class CombatOptions
         /// If enabled, this allows entities to turn around while casting
         /// </summary>
         public bool EnableTurnAroundWhileCasting = false;
+
+        /// <summary>
+        /// If enabled, this makes it so a player casting a friendly spell on a hostile target instead casts the spell upon themselves
+        /// </summary>
+        public bool EnableAutoSelfCastOnFriendlyHostile { get; set; } = false;
```
lodicolo, line 75: `EnableAutoSelfCastFriendlySpellsWhenTargetingHostile`

lodicolo (changes requested):

Intersect (Core)/Config/CombatOptions.cs
AlexVild: done

lodicolo (approved):

PR #2097, opened 2023-11-17
feature/chore: format for abbreviated time

### LINKED ISSUE #783: feature: abbreviation on spell effects and item reuse
I have noticed that when making spells or objects that have a reuse or effect greater than 60 seconds, they do not apply the correct abbreviation
It would be great if when they exceed 60 seconds the abbreviation "M" is used and the value becomes in minutes

Just use the "S" (Seconds), "M" (Minutes) and "H" (Hours) for Spells and item effects/reuse

**Additional context**
Screenshot_131
Screenshot_132

Arufonsu:
- should resolve #783
- allows to have abbreviated time strings, which are localized so other languajes can adjust the abbreviations as neeeded

Preview:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/6b7d19c4-61f4-4bd1-af0a-9dcd4a9b8558

imagenimagen

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/4f52a981-51c1-421a-83d8-fa25c0d172e2

imagen imagen


Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
             }
         }
 
+        public static string FormatTimeAbbreviated(long milliseconds)
+        {
+            if (milliseconds == 0)
+            {
+                return string.Empty;
+            }
```
lodicolo, line 84: Don't use this case

Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
...
+
+                // weeks
+                case <= 6048000000:
+                    timeValue = milliseconds / 604800000.0;
+                    postfix = Time.Week;
+                    break;
+
+                default:
+                    return "OOB";
+            }
```
lodicolo, line 123: Switch on `<timeSpan>.Ticks` and cases should use `TimeSpan.TicksPer<unit>` for the values

Units should be `ns`, `μs`, `ms`, `s`, `m`, `d`, `w`, `y`

There is no constant for nanoseconds, weeks, or years so nanoseconds would be `<= TimeSpan.TicksPerMicrosecond`, days is `<= TimeSpan.TicksPerDay * 7` and weeks is ` <= TimeSpan.TicksPerDay * 365` -- years is `default` case

Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
             }
         }
 
+        public static string FormatTimeAbbreviated(long milliseconds)
```
lodicolo, line 79: This should also accept a `string? numericalFormat = default` parameter

Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
...
+        {
+            if (milliseconds == 0)
+            {
+                return string.Empty;
+            }
+
+            double timeValue;
```
lodicolo, line 86: Rename to `timeInUnit`

Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
...
+            string postfix;
```
lodicolo, line 87: Rename to `suffix`

Intersect.Client/Localization/Strings.cs
```diff
@@ -76,6 +76,62 @@ public static string FormatQuantityAbbreviated(long value)
...
+            }
+
+            if (timeValue >= 10)
+            {
+                timeValue = Math.Floor(timeValue);
+                return timeValue + postfix;
+            }
+
+            timeValue = Math.Floor(timeValue * 10) / 10.0;
+            return timeValue.ToString("F1") + postfix;
```
lodicolo, line 132: Replace with
```csharp
var formattedTimeSpan = timeInUnit.ToString($"{numericalFormat}'{suffix}");
return formattedTimeSpan;
```

lodicolo: All `FormatWithSuffix()` invocations can pass `"0."` as the format string to round to the nearest whole number

lodicolo (changes requested):

PR #2107, opened 2023-11-21
Rejoin shared instance

AlexVild:
Allows ex-party members to rejoin a shared instance

### DIFF
--- a/Intersect.Server.Core/Entities/Player.cs
+++ b/Intersect.Server.Core/Entities/Player.cs
@@ -4967,10 +4967,11 @@ namespace Intersect.Server.Entities
             Party.Add(target);
 
             //Update all members of the party with the new list
-            for (var i = 0; i < Party.Count; i++)
+            var cachedParty = Party.ToList();
+            foreach (var member in cachedParty)
             {
-                Party[i].Party = Party;
-                PacketSender.SendParty(Party[i]);
+                member.Party = cachedParty;
+                PacketSender.SendParty(member);
                 PacketSender.SendChatMsg(
                     Party[i], Strings.Parties.joined.ToString(target.Name), ChatMessageType.Party, CustomColors.Alerts.Accepted
                 );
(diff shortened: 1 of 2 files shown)


Intersect.Server.Core/Entities/Player.cs
```diff
@@ -4906,42 +4940,42 @@ public void AddParty(Player target)
...
-            else
+
+            target.LeaveParty();
+            Party.Add(target);
+
+            //Update all members of the party with the new list
+            for (var i = 0; i < Party.Count; i++)
             {
-                PacketSender.SendChatMsg(this, Strings.Parties.limitreached, ChatMessageType.Party, CustomColors.Alerts.Error);
+                Party[i].Party = Party;
```
lodicolo, line 4972: Cache Party[i]

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -2125,11 +2125,45 @@ public bool CanChangeToInstanceType(MapInstanceType? instanceType, bool fromLogi
...
-                        isValid = false;
+                        // Check to see if the instance still exists & is populated
+                        if (!InstanceProcessor.TryGetInstanceController(SharedMapInstanceId, out var sharedController) || sharedController.Players.Count == 0)
+                        {
+                            return false;
+                        }
+
+                        // If it does, get the party leader from it...
+                        var partyMember = sharedController.Players.FirstOrDefault();
+                        if (partyMember == default)
```
lodicolo, line 2136: What happens if _you_ were the party leader? I actually have no idea what happens to a 3+ person party if the lead disconnects

lodicolo (changes requested):

Intersect.Server.Core/Entities/Player.cs
AlexVild: The leadership goes to the next in line of the array.

Intersect.Server.Core/Entities/Player.cs
AlexVild: Done

lodicolo (approved):

PR #2143, opened 2023-12-13
bug fix #1716

### LINKED ISSUE #1716: bug: Removing/Changing position of Lights [L] in editor doesn't function correctly
### Description

I'm not sure which update this bug took place (my bad for not reporting sooner) but sometime in 7.1

Editing a map with Lights already placed, will cause the incorrect light to be deleted/moved.
This bug also happens when moving lights - both on single/multi layer edits.

It actually looks like it's also duplicating the light tile as well:
https://user-images.githubusercontent.com/110916959/215350302-55c9b03c-fe5c-47b5-8c30-7b979366a9aa.mp4

### Steps to Reproduce

1. Add light to map via editor
2. save map
3. Try to update those lights

### Version with bug

v0.7.1.88

Bug was introduced sometime in 7.1 I believe.  A few months back at least.

### Affected platforms

Windows, macOS

no

### Relevant log output

_No response_

### Duplicate Bug Check

TimtyG:
Fixes #1716

I've added TileX and TileY to both methods GetHashCode() & Equals() in LightBase.cs

This change ensure that the hash code and equality comparison take into account the TileX and TileY properties, necessary for drag / copy / undo ...

### DIFF
--- a/Intersect (Core)/GameObjects/LightBase.cs
+++ b/Intersect (Core)/GameObjects/LightBase.cs
@@ -68,7 +68,7 @@ namespace Intersect.GameObjects
         /// <returns></returns>
         public override int GetHashCode()
         {
-            return Intensity ^ (int)Expand ^ (Color?.GetHashCode() ?? 0) ^ TileX.GetHashCode() ^ TileY.GetHashCode();
+            return HashCode.Combine(Intensity, (int)Expand, Color, TileX, TileY);
         }
 
         /// <summary>
--- a/Intersect (Core)/GameObjects/LightBase.cs
+++ b/Intersect (Core)/GameObjects/LightBase.cs
@@ -68,7 +68,7 @@ namespace Intersect.GameObjects
         /// <returns></returns>
         public override int GetHashCode()
         {
-            return Intensity ^ (int)Expand ^ (Color?.GetHashCode() ?? 0);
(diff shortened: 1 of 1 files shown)


Intersect (Core)/GameObjects/LightBase.cs
```diff
@@ -68,7 +68,7 @@ Color color
         /// <returns></returns>
         public override int GetHashCode()
         {
-            return Intensity ^ (int)Expand ^ (Color?.GetHashCode() ?? 0);
+            return Intensity ^ (int)Expand ^ (Color?.GetHashCode() ?? 0) ^ TileX.GetHashCode() ^ TileY.GetHashCode();
```
lodicolo, line 71: Change to `HashCode.Combine(...)`, don't pass `GetHashCode()` into the `Combine(...)` method

lodicolo (changes requested):

Intersect (Core)/GameObjects/LightBase.cs
TimtyG: return HashCode.Combine(Intensity, (int)Expand, Color, TileX, TileY);

Will work for you Panda?

lodicolo: @TimtyG Don't cast to int here, just pass Expand directly into the method

TimtyG: Alright, on it

lodicolo (approved):
Approved, waiting on autobuild before merging

PR #2144, opened 2023-12-15
fix: (Day) Fixes issue where shields did not scale with equipment stats'

### LINKED ISSUE #2118: enhancement: shield spells do not scale with scaling stat, only give flat amount
### Description

Giving a spell the "Shield" bonus effect will only make the shield as strong as the base amount, the shield does not scale with a scaling stat like other abilities do.

https://streamable.com/m32fpo

### Steps to Reproduce

1. Make a shield spell with a scaling stat and scaling amount
2. Cast the spell
3. The shield only applies the base amount of the spell, no scaling at all

### Version with bug

v.8.0.111

Unknown

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

AlexVild:
Resolves #2118

### DIFF
--- a/Intersect.Server.Core/Entities/Combat/Status.cs
+++ b/Intersect.Server.Core/Entities/Combat/Status.cs
@@ -106,12 +106,12 @@ namespace Intersect.Server.Entities.Combat
             // If we're adding a shield, actually add that according to the settings.
             if (type == SpellEffect.Shield)
             {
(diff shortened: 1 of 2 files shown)


lodicolo (approved):
Approving, assuming it compiles. Will wait on the PR build before ~~yeeting~~ merging

PR #2156, opened 2024-02-02
Fix #2155

### LINKED ISSUE #2155: bug : bank deposits equipped items without unequipping them
### Description

The problem is that when we have two non-stackable items, such as equipment, if we want to hide one of them, an input box appears asking how many items we want to hide.
The maximum number is immediately marked and we can put it in the bank. (using shift it immediately shifts both non-stackable ones)
If we have an equipped item and the other one is not equipped, and we click either using the contextbox or shift to hide the unequipped item, it will move both of them.
And now, if we take out another item, regardless of whether we meet the conditions to equip it or whether it is an item that cannot be equipped, it is equipped in the slot where we previously had the item (and now we don't have it because it is in the bank). .

The next problem is that it imposes HP and MP and ExtraBuffy on this item, so you can have over 100% MP or HP and you can equip things that you shouldn't.

After leaving and entering the server, these items may disappear, but not always, which may mean that, for example, we put armor on our heads that will give us higher stats than a helmet.

I am attaching videos

### Steps to Reproduce

1. Create 2 different equipable items in the editor.
2. Go to the customer and give yourself 2 pieces of one item and 1 piece of the other.
3. Go to the bank.
4. Wear the item you have 2.
5. Hide the item of which you have 1 piece (The other one)
6. Use the bank in the contextmenu or use shift or double-click on the item you have 2 and which is not equipped and hide it in the bank this way. (Hide the maximum amount if the inputbox appears)
7. When the items you had two pieces of are hidden, cash out the item you put in the bank earlier with the single one.
8. BUG: this item should be in the slot where you previously had the item equipped.

### Version with bug

0.8.127

I dont know

### Affected platforms

Windows

I will send a PR soon

https://github.com/AscensionGameDev/Intersect-Engine/assets/144380768/e1f7182c-9869-4422-a54d-721df731f6fb

https://github.com/AscensionGameDev/Intersect-Engine/assets/144380768/34bd39e7-ccba-4932-9a59-1c3042561d22

Zrzut_ekranu_88

### Relevant log output

_No response_

### Duplicate Bug Check

cydyn:
A check has been added to the code to see if an item is non-stackable.
If it is non-stackable, as in the case of equipment, a window does not open asking how many items we want to put in the bank.

This solves the problem with transferring items from the inventory to the bank, because now an item equipped on a player will not be transferred with another identical item.

Individually transferred inventory items are removed correctly, so there is no possibility of an error/bug

Shift has also been changed because using the shift button and the mouse button will move all the same items that we have in the inventory.

Now it simply moves a given slot without displaying the window showing how much we want to put in the bank.

If we do not use shift, the window appears but not in the case of non-stackable items

EDIT: I changed the inputbox to be shown when withdrawing and it was possible to withdraw more than 1 item. Because in this case there is no way to make an error

### DIFF
--- a/Intersect.Client/Entities/Player.cs
+++ b/Intersect.Client/Entities/Player.cs
@@ -962,29 +962,22 @@ namespace Intersect.Client.Entities
                 PacketSender.SendWithdrawItem(bankSlotIndex, movableQuantity, inventorySlotIndex);
                 return true;
             }
-            // Checking if the item is non-stackable
-            if (!itemDescriptor.IsStackable || skipPrompt)
-            {
-                PacketSender.SendWithdrawItem(bankSlotIndex, 1, inventorySlotIndex);  // Only 1 item
-                return true;
-            }
-            else
-            {
-                InputBox.Open(
-                    title: Strings.Bank.withdrawitem,
-                    prompt: Strings.Bank.withdrawitemprompt.ToString(itemDescriptor.Name),
-                    modal: true,
-                    inputType: InputBox.InputType.NumericSliderInput,
-                    onSuccess: WithdrawItemInputBoxOkay,
-                    onCancel: null,
-                    userData: new[] { bankSlotIndex, inventorySlotIndex },
-                    quantity: movableQuantity,
-                    maxQuantity: maximumQuantity
-                );
 
-                return true;
-            }
+            InputBox.Open(
+                title: Strings.Bank.withdrawitem,
+                prompt: Strings.Bank.withdrawitemprompt.ToString(itemDescriptor.Name),
+                modal: true,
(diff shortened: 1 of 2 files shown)


cydyn:
@WeylonSantana The bank was changed, and as far as I know, it has been possible to deposit items that we have on file for a long time.

Even if I couldn't deposit the item I have, when I have a second one, an input box appears that tells me how many items I want to deposit. Then there's a problem because it puts both.

cydyn:
I decided that this way I would get rid of the problem forever.
Because thanks to this, you can place items that are already on the player and they will be removed automatically. However, with a double quantity of the same item, he wants to place both, and when paying out the item, the game treats the new item in the inventory as equipped

PR #2173, opened 2024-03-09
2172: Add support for a crafting journal

### LINKED ISSUE #2172: enhancement: Add a "journal mode" for the "Open Crafting Table" event command
### Description

This option would allow users to view crafts, but would not allow them to _craft_ crafts. This is a very useful addition to the user experience, as it can be hard to keep track of what is needed for which crafts otherwise. It's also an incredibly simple fix.

### Intended Use-Case

This would allow users to quickly reference their available crafts.

### Duplicate Check

AlexVild:
Resolves #2172

Adds a "Journal Mode" for the "Open Crafting Window" event command. This prevents a user from being able to craft, but displays the crafts of the table to them.

In-editor:
image

image

In-game:

**Journal Mode**
image

**Non-journal mode**
image


WeylonSantana (approved):

Intersect.Server.Core/Localization/Strings.cs
```diff
@@ -535,6 +535,9 @@ public sealed partial class CraftingNamespace : LocaleNamespace
             [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
             public readonly LocalizedString CraftFailureLostItems = @"The attempt to create the item {00} failed and you lost the materials!";
 
+            [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+            public readonly LocalizedString InJournalMode = @"You cannot craft from your crafting journal!";
```
Arufonsu, line 539: :white_check_mark:

Arufonsu (approved):

PR #2174, opened 2024-03-10
fix: last projectile spell casting

### LINKED ISSUE #1683: bug: spells do not correctly shoot their last ammunition
### Description

Hey, It's me again, and I don't know again and again if my report is good but I report this : 

- When a player uses a spell that has an "Ammunition Requirements", if the player uses his spell with exactly the required number of items, then the spell executes but without animation, and consuming the ammunition. 

This happens no matter how much ammo is required, all you need to do is have exactly the number requested and cast the spell then the bug appears.

Link Video : https://youtu.be/VPM22Nf8b60 

### Steps to Reproduce

Step 1 : Make an animation
Step 2 : Create a projectile with animation + Ammunition Requirements with "x" number (1,2,3,4..)
Step 3 : Create a spell with this projectile
Step 4 : Use the skill with the exact amount of Ammunition and the bug is here 

### Version with bug

v0.7.1.088

Unknow

### Affected platforms

Windows

Just have more than the Ammunition Requirements and it's good, the spell work

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
Fixes #1683 
Which reported a bug where a spell with ammunition requirements would consume the last projectile without actually casting it out. It is also intended to provide additional chore changes and move player related code out to it's corresponding overrides.

Files changed:
1. Entity.cs - Moved projectile removal to player's CastSpell override method. Also moved the player's specific handlers for spell's cooldown to player's CastSpell override method.
2. Player.cs - Removed redundant and repeated code within the UseSpell method (ie. CastSpell(..) is already called within Entity's Update method when required) and added projectile removal and cooldown handlers within the CastSpell override method.

Bug:

https://user-images.githubusercontent.com/17498701/206912343-1b49c1e3-e80f-4e66-96b6-376a60164f68.mp4

After the fix:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/ca7e1663-867d-4424-8a3a-e227f8a2982b

### DIFF
--- a/Intersect.Server.Core/Entities/Entity.cs
+++ b/Intersect.Server.Core/Entities/Entity.cs
@@ -2477,24 +2477,10 @@ namespace Intersect.Server.Entities
                         break;
                 }
 
-                if (spellSlot >= 0 && spellSlot < Options.MaxPlayerSkills)
+                if (this is not Player && spellSlot >= 0 && spellSlot < Options.MaxPlayerSkills)
                 {
-                    // Player cooldown handling is done elsewhere!
-                    if (this is Player player)
-                    {
-                        player.UpdateCooldown(spellBase);
-
-                        // Trigger the global cooldown, if we're allowed to.
-                        if (!spellBase.IgnoreGlobalCooldown)
-                        {
-                            player.UpdateGlobalCooldown();
-                        }
-                    }
-                    else
-                    {
-                        SpellCooldowns[Spells[spellSlot].SpellId] =
-                            Timing.Global.MillisecondsUtc + spellBase.CooldownDuration;
-                    }
+                    SpellCooldowns[Spells[spellSlot].SpellId] =
+                        Timing.Global.MillisecondsUtc + spellBase.CooldownDuration;
                 }
             }
             finally
--- a/Intersect.Server.Core/Entities/Entity.cs
+++ b/Intersect.Server.Core/Entities/Entity.cs
@@ -378,6 +378,20 @@ namespace Intersect.Server.Entities
         }
 
         /// <summary>
+        ///     Updates the entity's spell cooldown for the specified <paramref name="spellBase"/>.
+        ///     <para> This method is called when a spell is casted by an entity. </para>
+        /// </summary>
+        public virtual void UpdateSpellCooldown(SpellBase spellBase, int spellSlot)
+        {
+            if (spellSlot < 0 || spellSlot >= Options.MaxPlayerSkills)
+            {
+                return;
+            }
+
+            SpellCooldowns[Spells[spellSlot].SpellId] = Timing.Global.MillisecondsUtc + spellBase.CooldownDuration;
+        }
+
+        /// <summary>
         ///     Determines if this entity can move in the specified <paramref name="direction"/>.
         /// </summary>
         /// <param name="direction">The <see cref="Direction"/> the entity is attempting to move in.</param>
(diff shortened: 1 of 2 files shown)


Intersect.Server.Core/Entities/Player.cs
```diff
@@ -5438,36 +5437,16 @@ public void UseSpell(int spellSlot, Entity target)
...
                 }
 
-                // Check if the player isn't casting a spell already.
-                if (!IsCasting)
+                //Tell the client we are channeling the spell
+                if (IsCasting)
                 {
-                    // Player is not casting a spell, cast now!
-                    CastTime = 0;
-                    CastSpell(Spells[SpellCastSlot].SpellId, SpellCastSlot);
```
Arufonsu, line 5465: Explanation for removal: duplicated logic, we were already doing this here:

https://github.com/AscensionGameDev/Intersect-Engine/blob/116dd44e9260807fba89ac928a9f240849b6e57e/Intersect.Server.Core/Entities/Entity.cs#L314-L320

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -5438,36 +5437,16 @@ public void UseSpell(int spellSlot, Entity target)
...
-                    spell.Combat.TargetType == SpellTargetType.Projectile &&
-                    spell.Combat.ProjectileId != Guid.Empty)
-                {
-                    var projectileBase = spell.Combat.Projectile;
-                    if (projectileBase != null && projectileBase.AmmoItemId != Guid.Empty)
-                    {
-                        TryTakeItem(projectileBase.AmmoItemId, projectileBase.AmmoRequired);
-                    }
-                }
-
```
Arufonsu, line 5452: Moved this out to the override void CastSpell because this was removing projectiles before we could even cast them out properly, which leads to the actual bug of not being able to properly use the last projectile

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -5438,36 +5437,16 @@ public void UseSpell(int spellSlot, Entity target)
...
+                if (IsCasting)
                 {
-                    // Player is not casting a spell, cast now!
-                    CastTime = 0;
-                    CastSpell(Spells[SpellCastSlot].SpellId, SpellCastSlot);
-                    CastTarget = null;
-                }
-                else
-                {
-                    //Tell the client we are channeling the spell
```
Arufonsu, line 5470: Explanation for removal: duplicated logic, we were already doing this here:

https://github.com/AscensionGameDev/Intersect-Engine/blob/116dd44e9260807fba89ac928a9f240849b6e57e/Intersect.Server.Core/Entities/Entity.cs#L314-L320

Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -2477,24 +2477,10 @@ public virtual void CastSpell(Guid spellId, int spellSlot = -1)
...
-                        if (!spellBase.IgnoreGlobalCooldown)
-                        {
-                            player.UpdateGlobalCooldown();
-                        }
-                    }
-                    else
-                    {
-                        SpellCooldowns[Spells[spellSlot].SpellId] =
-                            Timing.Global.MillisecondsUtc + spellBase.CooldownDuration;
-                    }
```
Arufonsu, line 2497: Players have their own CastSpell override method so i moved these over

AlexVild (approved):

WeylonSantana (approved):

Arufonsu:
just to make sure, some extra testing with NPCs, all good:

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/0300c4da-38dc-4c92-aef1-45d11f2003ba

PR #2200, opened 2024-03-25
enhancement: negative bonus effects

cydyn:
Added the ability to set negative bonus effect values.
Negative CDR increases spell cooldown.
Negative Tenacity increases the duration of effect, e.g. stun, but does not cause the damage to be duplicated over time.
A negative exp and luck bonus results in lower chances of drops and lower exp for the player.
Lifesteal and manasteal are negative, it works like 0, but if other equipment adds lifesteal or manasteal, it is known that the minus value takes it into account. PS: Works as 0 because healing spells have minus values, which causes a minus lifesteal to double the healing
The characterwindow has also been changed so that negative values ​​can be shown

### DIFF
--- a/Intersect.Server.Core/Entities/Entity.cs
+++ b/Intersect.Server.Core/Entities/Entity.cs
@@ -2081,128 +2081,129 @@ namespace Intersect.Server.Entities
             var thisPlayer = this as Player;
 
             //Check for lifesteal/manasteal
-            if (this is Player && !(enemy is Resource))
-            {
-                var lifestealRate = thisPlayer.GetEquipmentBonusEffect(ItemEffect.Lifesteal) / 100f;
+            if (baseDamage > 0 && this != enemy)
+                if (this is Player && !(enemy is Resource))
+                {
+                    var lifestealRate = thisPlayer.GetEquipmentBonusEffect(ItemEffect.Lifesteal) / 100f;
                     if (lifestealRate < 0)
-                {
-                    lifestealRate = 0; // If lifestealRate is negative, set to zero
-                }
-                var idealHealthRecovered = lifestealRate * baseDamage;
-                var actualHealthRecovered = Math.Min(enemyVitals[(int)Vital.Health], idealHealthRecovered);
+                    {
+                        lifestealRate = 0; // Jeśli lifestealRate jest ujemny, ustaw na zero
+                    }
+                    var idealHealthRecovered = lifestealRate * baseDamage;
+                    var actualHealthRecovered = Math.Min(enemyVitals[(int)Vital.Health], idealHealthRecovered);
 
-                if (actualHealthRecovered > 0)
-                {
-                    // Don't send any +0 msg's.
-                    AddVital(Vital.Health, (int)actualHealthRecovered);
-                    PacketSender.SendActionMsg(
-                        this,
-                        Strings.Combat.addsymbol + (int)actualHealthRecovered,
-                        CustomColors.Combat.Heal
-                    );
-                }
+                    if (actualHealthRecovered > 0)
+                    {
+                        // Don't send any +0 msg's.
+                        AddVital(Vital.Health, (int)actualHealthRecovered);
+                        PacketSender.SendActionMsg(
+                            this,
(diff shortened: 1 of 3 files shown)


Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -2084,6 +2084,10 @@ public virtual void TryAttack(Entity target)
             if (this is Player && !(enemy is Resource))
             {
                 var lifestealRate = thisPlayer.GetEquipmentBonusEffect(ItemEffect.Lifesteal) / 100f;
+                    if (lifestealRate < 0)
```
WeylonSantana, line 2087: remove this if, regardless of the value being negative, it will not pass the next check on the line 2094

Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -2099,6 +2103,10 @@ public virtual void TryAttack(Entity target)
                 }
 
                 var manastealRate = (thisPlayer.GetEquipmentBonusEffect(ItemEffect.Manasteal) / 100f);
+                   if (manastealRate < 0)
```
WeylonSantana, line 2106: remove this if, regardless of the value being negative, it will not pass the next check on the line 2113

WeylonSantana (changes requested):

Intersect.Server.Core/Entities/Entity.cs
cydyn: I don't understand why he shouldn't pass the next inspection? After all, this is just an exception which, if there is a minus value, sets it to 0, i.e. the default basic amount. This change must be made, otherwise lifesteal will not work properly, as I have already explained. unless I don't understand what you mean

Intersect.Server.Core/Entities/Entity.cs
cydyn: Same situation here. After all, it's just an exception that checks whether the lifestealrate or manastealrate is minus. if so, change it to 0, i.e. the default value. So it should not interfere with the next check regarding +0msg

Intersect.Server.Core/Entities/Entity.cs
cydyn: unless we want the negative lifesteal and manasteal to take the user's life after the attack. then instead, you can give a condition for basedamage that if basedamage is a negative value, it changes the sign from - to +. This is due to the problem with healing spells, which in their recipe must have negative values ​​to heal.

cydyn: I will explain this further. If we didn't have this "if", then if we had -30% lifesteal and used a healing spell whose formula had e.g. -10 to heal by 10, then there would be an error in the code. because lifestealrate -30% * basedamage -10, gives a positive result, which causes additional treatment.

cydyn:
Okay, I know what weylon means. I will make sure that the treatment works properly and lifesteal takes the user's life with a negative value.

cydyn:
Two changes are introduced.
One adds options in the code
            if (baseDamage > 0 && this != enemy)

This means that if we do not deal damage and if we deal damage to ourselves, e.g. a trap that uses a self spell that deals damage, then neither lifesteal nor manasteal works.

Moreover, I added options in the characterwindow so that if lifesteal or manasteal had a negative value, it does not show them, it only shows 0.

I don't know how to do it via the server.
The calculation of all effects is not separate, but common, so....
       public int GetEquipmentBonusEffect(ItemEffect effect)
        {
            var value = 0;

            foreach (var item in EquippedItems)
            {
                if (!item.Descriptor.EffectsEnabled.Contains(effect))
                {
                    continue;
                }
                value += item.Descriptor.GetEffectPercentage(effect);
            }

            return value;
        }

cydyn:
How about adding that life theft would show negative values ​​in charcterwindows as 0 (lifestealrate). thanks to which we have both 0 and a negative value, thanks to which we know how much we currently have

lodicolo:
This is not an enhancement, it is a feature request, and one that needs more attention to detail from a wider audience. A more thorough proposal should have been done in [Discussions](https://github.com/AscensionGameDev/Intersect-Engine/discussions) first.

Closing because this feature is not currently under consideration.

Issue #2226, opened 2024-04-11
bug: EquipmentOnHitTriggers when shooting projectiles


cydyn:
### Description

Weapons with the projectile setting do not use or incorrectly use the onhit event when this option is added.
PS: I give video in discord. I cant here, file too big

### Steps to Reproduce

1. Make a weapon with a projectile and a spell or anything on commonevent onhit.
2. Use it.
3. Make a weapon with extra attack effect and add this common event.
4. The gun works.
5. The projectile weapon will also work even if there is no event because it inherits from the previous one

### Version with bug

Latest

### Last version that worked well

Unknown

### Affected platforms

Windows, Linux

### Did you find any workaround?

No

### Duplicate Bug Check

- [X] This bug report is not a duplicate to the best of my knowledge.

Arufonsu:
Thanks to @PyroTech03 for explaining me this further:

![image](https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/a78b2aa5-f5ad-4189-ba03-79040959c38a)

so.. ie:
weapon A triggers event A
weapon B triggers event B

when switching around weapons, i'm now unable to trigger event A no matter what at some point
![image](https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/ef31a120-0b7e-4745-8d3e-5501e3a801c1) ![image](https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/891b6236-9b0e-4a16-b3a1-3b200a79fcf3) ![image](https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/cc7ff0d7-f241-4343-bda7-530b31c2aa32)

PR #2227, opened 2024-04-12
Fix: Proper operation of projectile

cydyn:
Fixes: #2226 

In the new commit adding the ability to add common events to weapons or items, there was a problem with incorrect inheritance in the case of projectile weapons and melee weapons.
The video is on the bugs discord channel because it's too big.
Here I am posting a smaller picture of what the situation looks like after fixing the error, in my opinion.
[bandicam 2024-04-12 08-21-31-021.zip](https://github.com/AscensionGameDev/Intersect-Engine/files/14954724/bandicam.2024-04-12.08-21-31-021.zip)

### DIFF
--- a/Intersect.Server.Core/Entities/Player.cs
+++ b/Intersect.Server.Core/Entities/Player.cs
@@ -218,12 +218,6 @@ namespace Intersect.Server.Entities
         [NotMapped, JsonIgnore]
         public bool IsFading { get; set; }
 
-        /// <summary>
-        /// Reference stored of the last weapon used for an auto-attack
-        /// </summary>
-        [NotMapped, JsonIgnore]
-        public ItemBase LastAttackingWeapon { get; set; }
-
         // Instancing
         public MapInstanceType InstanceType { get; set; } = MapInstanceType.Overworld;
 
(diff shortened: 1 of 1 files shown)


Mighty-Professional:
Wont this remove the on hit event? I am a bit confused of the change. Do you mind explaining the logic a bit more?

cydyn:
@Mighty-Professional It does not remove the on-hit event, but causes the projectiles to fire. I've already presented everything on the bugs channel. a video of how it worked, before the repair is on the channel and after the repair is here. everything works as it should.
The bullets didn't work at all, and if they did, they inherited from the previous weapon indefinitely, so you could take the effect from other weapons and thus improve your ranged weapon.

cydyn:
@Mighty-Professional Maybe I'll explain it a little more precisely. The lastattackweapon was removed, which was intended to prevent the activation of the effect of a new weapon when quickly changing weapons, and it inherited the effect of the previous weapon. (I don't know why it was added, but you can prevent quick changing via CDR) Now correctly activates the weapon's trigger when auto-attacking. When lastattackweapon was available, projectile did not work at all and did not trigger the required effect

Cheshire92:
I don't agree with this solutions, deleting random code is not the way to go.
Looking at how this works around the issue, a more likely culprit seems to be that firing a projectile doesn't set the LastAttackingWeapon variable correctly.

Arufonsu:
Thanks to @PyroTech03 for explaining me this further:

image

so.. ie:
weapon A triggers event A
weapon B triggers event B

when switching around weapons, i'm now unable to trigger event A no matter what at some point
image image image

Arufonsu:
Closing this as the issue was solved by #2241

PR #2236, opened 2024-04-29
enhancement: trade offer accepted

Arufonsu:
- The trade button is now disabled upon pressing it.
- Players will now receive an informative message when the counterparty accepts theirs offer, indicating readiness for the trade to proceed.

Preview:
image
image image


Arufonsu:
> Please just double check they can trade two times in a row to make sure the trade disabled flag is cleared.

All good !

image

Issue #2242, opened 2024-05-05
bug: Packing sound and music assets


cydyn:
### Description

When we want to create an update file in the editor as assets, the raw sound and music files do not have the ClientIgnore mark, which means that sound and music packages and raw files are downloaded, although it should only download packages

### Steps to Reproduce

1. In the editor, we set the packing of files into assets
2. We create an update file
3. In upadte.json, sounds do not have the ClientIgnore mark and are downloaded

### Version with bug

0.8.00

### Last version that worked well

Dont know

### Affected platforms

Windows, Linux

### Did you find any workaround?

No. I have fix.

### Duplicate Bug Check

- [X] This bug report is not a duplicate to the best of my knowledge.

xZACKYx:
Excellent idea, not many of us like to create musical sounds so that they can be easily stolen, it is not easy to simply pay a lot of money for copyrights, it is also an extra layer to the effort.

lodicolo:
Fixed in ea2f135

PR #2243, opened 2024-05-05
fix: package music and sounds

cydyn:
Fix issue: #2242 
It will now properly add ClientIgnore to raw audio and music files, and will only download packages


Arufonsu:
okay
- do packed music/sounds work at all? and i mean, remove the non-packed music/sounds, and test if music/sounds actually play within the client
- there was a reason for music not being packed, pretty sure it was intentional from the authors/maintainers at some point, i just can't find that information anywhere at this point, this is why i'm asking you to test this out and provide us with a video of this feature actually fully working while it packs and plays the corresponding audio assets.

cydyn:
@Arufonsu Music is packaged and played correctly, and when downloading, only packages without raw music are downloaded.
Will try to send a video tomorrow how it works.

lodicolo:
This solution removed comments that should have stayed, and relied directly on a magic hardcoded string when we currently want to be reducing instances of that, not adding more.

The underlying issue has been fixed in ea2f135 and this PR isn't needed.

PR #2250, opened 2024-05-14
chore: long value type for Vitals, Exp and ConsumableData

Arufonsu:
We previously updated the EXP values to long type so we could store values from -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807, but we didn't quite updated everything else related to this, plus, we ended up with lack of support for large values on vitals and consumable data... this chore PR makes sure to cover the missing stuff

### Preview:  
image
image image image image
image

https://github.com/AscensionGameDev/Intersect-Engine/assets/17498701/ca1df184-d91e-47b4-ba37-2bd241c8819e

### DIFF
--- a/Intersect (Core)/GameObjects/ClassBase.cs
+++ b/Intersect (Core)/GameObjects/ClassBase.cs
@@ -22,7 +22,7 @@ namespace Intersect.GameObjects
         public int[] BaseStat = new int[Enum.GetValues<Stat>().Length];
 
         [NotMapped]
-        public int[] BaseVital = new int[Enum.GetValues<Vital>().Length];
+        public long[] BaseVital = new long[Enum.GetValues<Vital>().Length];
 
         [NotMapped]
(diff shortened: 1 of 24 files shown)


WeylonSantana:
One thing that should come along with this change would be to shorten it too, right? Starting from 1 million, start to be abbreviated and avoid the giant numbers coming out of the container, etc.

Arufonsu:
> One thing that should come along with this change would be to shorten it too, right? Starting from 1 million, start to be abbreviated and avoid the giant numbers coming out of the container, etc.

I agree on this but i don't think it should be part of this PR specifically 🤔 i'll give it a go and improve these in another enhancement PR (would be cool if you start an enhancement request with this), i have an entire set of ideas for improvements, including a minimize button for it (similar to the chatbox eye button).

Mighty-Professional (approved):

EspectroW94:
https://github.com/AscensionGameDev/Intersect-Engine/assets/127634629/e4c2cc4e-7dc6-434e-b7c2-a2ae088935b5

PR #2254, opened 2024-05-23
fix: don't clear current target when taunted

### LINKED ISSUE #2253: bug: clear target when taunted
### Description

When we are under the taunt effect, we cannot select other targets, ok, but we can remove the target, which is wrong

### Steps to Reproduce

need a check here

https://github.com/AscensionGameDev/Intersect-Engine/blob/4eff95eca51bb98aac144c2ba0f08cd408e8c3f1/Intersect.Client/Entities/Player.cs#L2045

### Version with bug

current

Unknown

### Affected platforms

Windows

_No response_

### Relevant log output

_No response_

### Duplicate Bug Check

Arufonsu:
This PR Fixes #2253

- Expected behavior: taunt's effect forces it's target to be the caster, removing the target shouldn't be possible.
- Current-not-expected behavior: pressing the esc / menu key clears the current target while taunted.

- After this PR: taunt's effect forces it's target to be the caster, removing the target is not possible at all

### DIFF
--- a/Intersect.Client/Core/Input.cs
+++ b/Intersect.Client/Core/Input.cs
@@ -120,7 +120,7 @@ namespace Intersect.Client.Core
                 {
                     // We've closed our windows, don't do anything else. :)
                 }
-                else if (Globals.Me != null && Globals.Me.TargetIndex != Guid.Empty)
+                else if (Globals.Me != null && Globals.Me.TargetIndex != Guid.Empty && !Globals.Me.Status.Any(s => s.Type == Enums.SpellEffect.Taunt))
                 {
                     Globals.Me.ClearTarget();
                 }


WeylonSantana:
In my opinion, it makes more sense to put it as the first check within the clear target instead of the input, if we have a way to clear the target in other ways than the input, then it better be handled within the method.

Arufonsu:
> In my opinion, it makes more sense to put it as the first check within the clear target instead of the input, if we have a way to clear the target in other ways than the input, then it better be handled within the method.

ClearTarget is a boolean function, which is used 5 times, where only one is related to the player's input, other 4 uses i found for it are not related to player's input but interface and null entity targeting checks and cleanup processes, which are there to prevent issues like: having a null entity box up and crash, hiding the entity box when destroying / cleaning up gwen, etc. 

Give it a try and see how it works, also have a look on the 5 use cases, this is literally the one and only place where you want this check to be at, otherwise you will end up with issues along the way by checking this status on all use cases

WeylonSantana:
okays, fine for me then.

WeylonSantana (approved):

PR #2268, opened 2024-06-07
feature: amount range for item drops

WeylonSantana:
resolves #2209

The default Amount for all games now becomes MaxQuantity in the logic of this pr, but nothing has been renamed, and there is a migration that updates the npcs and resources tables to add MinQuantity. So after this pr, all games will drop a minimum of 1 - Amount when they update.

https://github.com/AscensionGameDev/Intersect-Engine/assets/63019821/53e4851f-c94d-420f-8630-bae629023b36

https://github.com/AscensionGameDev/Intersect-Engine/assets/63019821/03d36584-7aa2-4987-bc22-7b835cccf7df

### DIFF
--- a/Intersect.Server.Core/Entities/Npc.cs
+++ b/Intersect.Server.Core/Entities/Npc.cs
@@ -134,7 +134,7 @@ namespace Intersect.Server.Entities
             foreach (var drop in myBase.Drops)
             {
                 var slot = new InventorySlot(itemSlot);
-                slot.Set(new Item(drop.ItemId, drop.Quantity));
+                slot.Set(new Item(drop.ItemId, Randomization.Next(drop.MinQuantity, drop.Quantity + 1)));
                 slot.DropChance = drop.Chance;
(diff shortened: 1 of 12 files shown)


Intersect.Server.Core/Entities/Npc.cs
```diff
@@ -134,7 +134,7 @@ public Npc(NpcBase myBase, bool despawnable = false) : base()
             foreach (var drop in myBase.Drops)
             {
                 var slot = new InventorySlot(itemSlot);
-                slot.Set(new Item(drop.ItemId, drop.Quantity));
+                slot.Set(new Item(drop.ItemId, Randomization.Next(drop.MinQuantity, drop.Quantity + 1)));
```
Arufonsu, line 137: 👍🏽

Arufonsu (approved):
tested and working, code seems to be in order 👍🏽 
(PR is huge due to chore changes and migrations)

Issue #2287, opened 2024-06-16
bug: server closes when attacking entity with projectile


theceltos:
### Description

When trying to kill a monster using a projectile, the server closes by itself.
If the monster dies with just 1 hit, nothing happens, but if you don't kill it in the first hit, the server dies.

### Steps to Reproduce

1 - Select any monster that has more health than you can take damage from.
2 - Throw a projectile at this monster.
3 - Watch the server die.

### Version with bug

0.8.0.0 after pr #2281

### Last version that worked well

Before #2281

### Affected platforms

Windows, I was *not* able test on other platforms

### Relevant log output

```shell
2024-06-16 01:35:38.599 [Error] Received unhandled exception from System.Threading.Thread: Index was outside the bounds of the array.

--------------------------------------------------------------------------------
2024-06-16 01:35:38.610 [Error] IndexOutOfRangeException: Index was outside the bounds of the array.
    Stack:    at Intersect.Server.Entities.Projectile.CheckForCollision(ProjectileSpawn spawn) in D:\The Celtos Game\Engine\Intersect-Engine\Intersect.Server.Core\Entities\Projectile.cs:line 345
   at Intersect.Server.Entities.Projectile.ProcessFragments(List`1 projDeaths, List`1 spawnDeaths) in D:\The Celtos Game\Engine\Intersect-Engine\Intersect.Server.Core\Entities\Projectile.cs:line 268
   at Intersect.Server.Maps.MapInstance.UpdateProjectiles(Int64 timeMs) in D:\The Celtos Game\Engine\Intersect-Engine\Intersect.Server.Core\Maps\MapInstance.cs:line 1383
   at Intersect.Server.Maps.MapInstance.Update(Int64 timeMs) in D:\The Celtos Game\Engine\Intersect-Engine\Intersect.Server.Core\Maps\MapInstance.cs:line 281
   at Intersect.Server.Core.LogicService.LogicThread.UpdateMap(MapInstance mapInstance, Boolean onlyProjectiles) in D:\The Celtos Game\Engine\Intersect-Engine\Intersect.Server.Core\Core\LogicService.LogicThread.cs:line 403


--------------------------------------------------------------------------------
```

### Duplicate Bug Check

- [x] This bug report is not a duplicate to the best of my knowledge.

WeylonSantana:
`Select any monster that has more health than you can take damage from.`
[...]

PR #2292, opened 2024-06-23
Fix: Homing Projectile Crash Server

cydyn:
Fix Issue #2287 
Please test this. I tested it and Alef tested it.
We both did not detect any error with the server crashing or other changes, but please test it.


theceltos:
I tested these changes here and everything went well.

Arufonsu:
image

PR #2301, opened 2024-07-03
enhancement: don't drop items in safe maps. pt1

theceltos:


WeylonSantana:
you can't check a new option in a pr and then send the other pr with just the check, I'm going to close both, besides we're trying to remove as much of "this is" from the entity, so we won't approve it if you send as is

PR #2302, opened 2024-07-03
enhancement: don't drop items in safe maps. pt2

theceltos:


WeylonSantana:
you can't check a new option in a pr and then send the other pr with just the check, I'm going to close both, besides we're trying to remove as much of "this is" from the entity, so we won't approve it if you send as is

theceltos:
Okay, it was a suggestion to put it in the engine, I already have this in my base. Could you at least suggest the change you need instead of *THIS IS*, instead of closing and saying you won't approve...

WeylonSantana:
> Okay, it was a suggestion to put it in the engine, I already have this in my base. Could you at least suggest the change you need instead of _THIS IS_, instead of closing and saying you won't approve...

I closed it because the pr is invalid

As I said above, there is no way we can accept a change that depends on another separated in this way, even if we approved it, github would not approve it, because github compiles the code and your code would not be compiled as the need for this pr to work would require from the other and vice versa, it has to be a complete change for pr

and I explained about "this is"
> we're trying to remove as much of "this is" from the entity

in your code there is `this is Player`, just like in other parts of the intersect there is `this is Resource`, `this is Npc`, and we are trying to remove this type of check from the entity file, and you are adding one more

PR #2353, opened 2024-07-28
chore: refactor register window

WeylonSantana:


lodicolo (changes requested):
I'm guessing @Arufonsu will want names un-abbreviated here too

lodicolo (approved):

PR #2373, opened 2024-08-07
feat: window background transparency

lodicolo:
- Add DrawShadow property (default true) to control drawing the backdrop shadow
- DrawBackground of the window now controls if the "frame" (the background, minus the titlebar) is visible
- DrawBackground of the Titlebar node now controls if the "titlebar" image is drawn

Resolves https://www.ascensiongamedev.com/topic/7075-how-do-i-get-this-off-my-screen-v2/


Arufonsu (approved):

ZerinSad:
nice!!

PR #2378, opened 2024-08-22
feat: add url buttons

RuyHiroTama:
- after the first launch of the client, the line "ButtonsURL" will appear in config.json, you need to enter a line in it according to this example "button name-url link".
- after the second launch, all buttons entered in "ButtonsURL" will be in the file ..resources\gui\layouts\game\ButtonsURL.json, they need to be laid out and the image replaced.
изображение_2024-08-22_131900461
изображение_2024-08-22_131934174
image
изображение_2024-08-22_131953454
изображение_2024-08-22_132115888


lodicolo:
This is best done via plugins or creating partial classes that we already have set up, this change is not one we want in the core engine.

PR #2380, opened 2024-08-26
feature: drop range

WeylonSantana:
resolves https://github.com/AscensionGameDev/Intersect-Engine/discussions/2209

The default Amount for all games now becomes MaxQuantity in the logic of this pr, but nothing has been renamed, and there is a migration that updates the npcs and resources tables to add MinQuantity. So after this pr, all games will drop a minimum of 1 - Amount when they update.

### DIFF
--- a/Intersect.Server/MIGRATIONS.md
+++ b/Intersect.Server/MIGRATIONS.md
@@ -37,10 +37,11 @@ Example:
 
 `dotnet ef migrations add Net7Upgrade --context MySqlLoggingContext --namespace Intersect.Server.Migrations.MySql.Logging --output-dir Migrations/MySql/Logging/ -- --databaseType MySql --connectionString "Username=<username>;Password=<password>;Database=<a db that exists>"`
 
+**Additionally, when trying to generate a migration on Windows you may need to include `Server=localhost` in your connection string.**
 
 **_A connection string is required for generating MariaDB/MySQL migrations_**, if you fail to provide one (or provide an invalid one) it will crash and fail with something like the following:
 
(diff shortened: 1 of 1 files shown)


Intersect.Editor/Forms/Editors/frmResource.Designer.cs
```diff
@@ -30,1069 +30,1102 @@ protected override void Dispose(bool disposing)
...
+            grpDrops.TabStop = false;
+            grpDrops.Text = "Drops";
+            // 
+            // nudDropMinAmount
+            // 
+            nudDropMinAmount.BackColor = System.Drawing.Color.FromArgb(69, 73, 74);
+            nudDropMinAmount.ForeColor = System.Drawing.Color.Gainsboro;
+            nudDropMinAmount.Location = new System.Drawing.Point(10, 200);
+            nudDropMinAmount.Margin = new Padding(4, 3, 4, 3);
+            nudDropMinAmount.Maximum = new decimal(new int[] { 10000000, 0, 0, 0 });
```
lodicolo, line 812: frmResource nudDropMinAmount missing Minimum

lodicolo (changes requested):
Because frmNpc has a minimum set it can't be used to verify that the value is actually 1 according to the server, two screenshots from the server, one from [DbInterface.cs#L780](https://github.com/WeylonSantana/Intersect-Engine/blob/9f2ae44770c08f8ff80644ea29508068b6c4230d/Intersect.Server.Core/Database/DbInterface.cs#L780) to verify NPCs and one from [DbInterface.cs#L801](https://github.com/WeylonSantana/Intersect-Engine/blob/9f2ae44770c08f8ff80644ea29508068b6c4230d/Intersect.Server.Core/Database/DbInterface.cs#L801) to verify Resources should be provided.

PR #2390, opened 2024-10-21
feature: Access to Source of Dropped/Spawned Items for Server Plugins

blinkuz:
This PR will allow access to the source of an item that is dropped/spawned on the map. The main goal is to expose this logic to the plugin system, specifically in the context of the server. It’s still a work in progress, but I would like to keep updating the draft in case they have any suggestions for the changes.


Intersect.Server.Framework/Items/EntityItemSource.cs
```diff
@@ -0,0 +1,16 @@
+using Intersect.Enums;
+using Intersect.Server.Framework.Entities;
+
+namespace Intersect.Server.Framework;
+
+public class EntityItemSource: IItemSource
+{
+    public Guid Id { get; set; }
+    public EntityType EntityType { get; set; }
+    public WeakReference<IEntity> EntityReference { get; set; }
```
lodicolo, line 10: All of these properties should be `init` instead of `set`

Intersect.Server.Framework/Items/EntityItemSource.cs
```diff
@@ -0,0 +1,16 @@
...
+}
+
+public class EntityItemSource<TEntity> : EntityItemSource where TEntity : class, IEntity
+{
+    public new WeakReference<TEntity> EntityReference { get; set; }
```
lodicolo, line 15: Back this with a local variable, `set` should be `init` and `init` should set both the local variable and `base.EntityReference`.

Intersect.Server.Framework/Entities/IEntity.cs
```diff
@@ -0,0 +1,13 @@
+using Intersect.Enums;
+
+namespace Intersect.Server.Framework.Entities;
+
+public interface IEntity: IDisposable
+{
+    Guid Id { get; }
+    string Name { get; }
+    Guid MapInstanceId { get; set; }
```
lodicolo, line 9: Either this should be read-only, or `X`, `Y`, and `Z` should also be writable

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -924,7 +931,7 @@ private void SpawnAttributeItem(int x, int y)
             {
                 mapItem.Quantity = 1;
             }
-            AddItem(mapItem);
+            AddItem(null, mapItem);
```
lodicolo, line 934: The one misgiving I have about doing this is because I honestly want to create a `MapItemSource` class that refers to the map instance (again via map instance ID and `WeakReference<MapInstance>`) since it's still valuable information.

Intersect.Server.Framework/Items/IItemSource.cs
```diff
@@ -0,0 +1,11 @@
+using Intersect.Enums;
+using Intersect.Server.Framework.Entities;
+
+namespace Intersect.Server.Framework;
+
+public interface IItemSource
+{
+    Guid Id { get; }
+    EntityType EntityType { get; }
```
lodicolo, line 9: I made a mistake in my original description in Discord, `EntityType` should be in `EntityItemSource` and this enum should be added and referenced here as `ItemSourceType SourceType { get; }` which `EntityItemSource` should implement as `public ItemSourceType SourceType => ItemSourceType.Entity;`:

```csharp
public enum ItemSourceType
{
    Unknown = 0,
    Entity = 1,
}
```

With respect to my comment from another file, I honestly want another enum value `Map = 2`.

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -713,26 +715,31 @@ private void AddItem(MapItem item)
...
     /// <summary>
     /// Spawn an item to this map instance.
     /// </summary>
+    /// <param name="source">The source of the item, which can be an entity or player</param>
     /// <param name="x">The horizontal location of this item</param>
     /// <param name="y">The vertical location of this item.</param>
     /// <param name="item">The <see cref="Item"/> to spawn on the map.</param>
     /// <param name="amount">The amount of times to spawn this item to the map. Set to the <see cref="Item"/> quantity, overwrites quantity if stackable!</param>
-    public void SpawnItem(int x, int y, Item item, int amount) => SpawnItem(x, y, item, amount, Guid.Empty);
+    public void SpawnItem(IItemSource? source,int x, int y, Item item, int amount) => SpawnItem(source, x, y, item, amount, Guid.Empty);
```
lodicolo, line 731: Formatting

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -713,26 +715,31 @@ private void AddItem(MapItem item)
...
     /// Spawn an item to this map instance.
     /// </summary>
+    /// <param name="source">The source of the item, which can be an entity or player</param>
     /// <param name="x">The horizontal location of this item</param>
     /// <param name="y">The vertical location of this item.</param>
     /// <param name="item">The <see cref="Item"/> to spawn on the map.</param>
     /// <param name="amount">The amount of times to spawn this item to the map. Set to the <see cref="Item"/> quantity, overwrites quantity if stackable!</param>
     /// <param name="owner">The player Id that will be the temporary owner of this item.</param>
-    public void SpawnItem(int x, int y, Item item, int amount, Guid owner, bool sendUpdate = true)
+    public void SpawnItem(IItemSource? source ,int x, int y, Item item, int amount, Guid owner, bool sendUpdate = true)
```
lodicolo, line 742: Formatting

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -713,26 +715,31 @@ private void AddItem(MapItem item)
...
         TileItems[item.TileIndex]?.TryAdd(item.UniqueId, item);
+        
+        //TODO: Invoke the event for the item being added to the map, maybe: 
+        // _mapHelper?.ItemAdded?.Invoke(this, new ItemEventArgs(source, item));
     }
 
     /// <summary>
     /// Spawn an item to this map instance.
     /// </summary>
+    /// <param name="source">The source of the item, which can be an entity or player</param>
```
lodicolo, line 726: Description should be `The source of the item, e.g. a player who dropped it, or a monster who spawned it on death` and if we add map sources it should be `The source of the item, e.g. a player who dropped it, or a monster who spawned it on death, or the map instance in which it was spawned`

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -713,26 +715,31 @@ private void AddItem(MapItem item)
...
     /// <param name="y">The vertical location of this item.</param>
     /// <param name="item">The <see cref="Item"/> to spawn on the map.</param>
     /// <param name="amount">The amount of times to spawn this item to the map. Set to the <see cref="Item"/> quantity, overwrites quantity if stackable!</param>
-    public void SpawnItem(int x, int y, Item item, int amount) => SpawnItem(x, y, item, amount, Guid.Empty);
+    public void SpawnItem(IItemSource? source,int x, int y, Item item, int amount) => SpawnItem(source, x, y, item, amount, Guid.Empty);
 
     /// <summary>
     /// Spawn an item to this map instance.
     /// </summary>
+    /// <param name="source">The source of the item, which can be an entity or player</param>
```
lodicolo, line 736: Description should be `The source of the item, e.g. a player who dropped it, or a monster who spawned it on death` and if we add map sources it should be `The source of the item, e.g. a player who dropped it, or a monster who spawned it on death, or the map instance in which it was spawned`

Intersect.Server.Core/Maps/MapInstance.cs
```diff
@@ -703,7 +704,8 @@ private void DespawnResources()
     /// <param name="x">The X location of this item.</param>
     /// <param name="y">The Y location of this item.</param>
     /// <param name="item">The <see cref="MapItem"/> to add to the map.</param>
-    private void AddItem(MapItem item)
+    /// <param name="source">The source of the item.</param>
```
lodicolo, line 707: This should be first, and the `x` and `y` parameter documentation should be removed since they don't exist here anymore

Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -21,12 +23,12 @@
...
 namespace Intersect.Server.Entities;
 
-public abstract partial class Entity : IDisposable
+public abstract partial class Entity : IEntity
 {
     //Instance Values
     private Guid _id = Guid.NewGuid();
 
-    public Guid MapInstanceId = Guid.Empty;
+    [NotMapped, JsonIgnore] public Guid MapInstanceId { get; set; } = Guid.Empty;
```
lodicolo, line 31: Why was `[NotMapped, JsonIgnore]` added?

Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -3056,7 +3058,26 @@ protected virtual void DropItems(Entity killer, bool sendUpdate = true)
...
+                else if (this is Npc npc)
+                {
+                    itemSource.Id = npc.Base.Id;
+                }
+                else if (this is Resource resource)
+                {
+                    itemSource.Id = resource.Base.Id;
+                }
+
+                instance.SpawnItem(itemSource, X, Y, drop, drop.Quantity, lootOwner, sendUpdate);
```
lodicolo, line 3080: This should be refactored out into an abstract method that all of the entity types implement, we should not add in type-specific code like this in an abstract class.

Intersect.Server.Framework/Items/MapItemSource.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Intersect.Server.Framework.Items;
+
+public partial class MapItemSource: IItemSource
+{
+    public Guid Id { get; init; }
+    public ItemSourceType SourceType => ItemSourceType.Map;
+    public WeakReference<IMapInstance> MapInstanceReference { get; init; }
+    public Guid MapInstanceId { get; init; }
+    public Guid MapControllerId { get; init; }
```
lodicolo, line 11: One of these two should be stored in `Id` -- I'm fine with `Id` and `InstanceId` or `Id` and `DescriptorId`.

Intersect.Server.Core/Entities/Projectile.cs
```diff
@@ -654,5 +655,10 @@ public override EntityType GetEntityType()
     {
         return EntityType.Projectile;
     }
+    
+    protected override EntityItemSource GetItemSource()
+    {
+        return null;
+    }
```
lodicolo, line 662: Method isn't nullable, it shouldn't return null. Projectile should just be

```csharp
        return new EntityItemSource
        {
            EntityType = GetEntityType(),
            EntityReference = new WeakReference<IEntity>(this),
            Id = this.Base.Id
        };
```

Same as Resource and Npc right?

Intersect.Server.Core/Entities/Events/EventPageInstance.cs
```diff
@@ -892,5 +894,10 @@ public bool ShouldDespawn(MapController map)
 
         return false;
     }
+    
+    protected override EntityItemSource GetItemSource()
+    {
+        return null;
+    }
```
lodicolo, line 901: Same as in `Projectile`, the method isn't nullable so it shouldn't return `null`, this should be showing up as a warning

I think for this one the following might be correct?

```csharp
        return new EntityItemSource
        {
            EntityType = GetEntityType(),
            EntityReference = new WeakReference<IEntity>(this),
            Id = this.MyPage.Id
        };
```

Intersect.Server.Core/Entities/Entity.cs
```diff
@@ -3056,14 +3059,17 @@ protected virtual void DropItems(Entity killer, bool sendUpdate = true)
...
+                var itemSource = this.GetItemSource();
+                instance.SpawnItem(itemSource, X, Y, drop, drop.Quantity, lootOwner, sendUpdate);
             }
 
             // Process the drop (for players this would remove it from their inventory)
             OnDropItem(slot, drop);
         }
     }
 
+    protected abstract EntityItemSource GetItemSource();
```
lodicolo, line 3071: This should be called `AsItemSource()`

The naming rationale is that `Get` is trying to fetch a copy of something the object has, but the item source is a reference, not a property of the entity (which would just be a property in cases where there's no parameter).

lodicolo (changes requested):
If you think Projectile and EventPageInstance's methods should return null, I'm ok with that _but_ you need to change the method signature in all of the classes from `EntityItemSource` to `EntityItemSource?` so that callers know if the value is nullable or not

Intersect.Server.Framework/Plugins/Helpers/IMapHelper.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Intersect.Server.Framework.Items;
+using Intersect.Server.Framework.Maps;
+
+namespace Intersect.Server.Plugins.Helpers;
+
+public interface IMapHelper
+{
+    event Action<IItemSource, IItem> ItemAdded;
+    IMapInstance GetMapInstanceByDescriptorId(Guid mapId, Guid mapInstanceId);
```
lodicolo, line 9: Honestly this should be a Try, since this _could_ return null.

Change to `bool TryGetMapInstance(Guid mapId, Guid instanceId, [NotNullWhen(true)] out IMapInstance? instance)`

The implementation should be `returnMapController.TryGetInstanceFromMap(mapId, instanceId, out instance);`

Intersect.Server.Framework/Plugins/Helpers/IMapHelper.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Intersect.Server.Framework.Items;
+using Intersect.Server.Framework.Maps;
+
+namespace Intersect.Server.Plugins.Helpers;
+
+public interface IMapHelper
```
lodicolo, line 6: Add `public delegate void ItemAddedHandler(IItemSource?, IItem);`

The `?` is important since the source can be null and we want static analysis to not lie to us

Intersect.Server.Framework/Plugins/Helpers/IMapHelper.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    event Action<IItemSource, IItem> ItemAdded;
```
lodicolo, line 8: Use `ItemAddedHandler` instead of `Action<T1, T2>`

Intersect.Server.Framework/Plugins/Helpers/IMapHelper.cs
```diff
@@ -0,0 +1,11 @@
...
+    IMapInstance GetMapInstanceByDescriptorId(Guid mapId, Guid mapInstanceId);
+    void InvokeItemAdded(IItemSource source, IItem item);
```
lodicolo, line 10: This shouldn't be on the interface because we only want the engine to be calling this

Intersect.Server.Core/Plugins/Helpers/MapHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public event Action<IItemSource, IItem> ItemAdded;
+
+    public IMapInstance GetMapInstanceByDescriptorId(Guid mapId, Guid mapInstanceId)
+    {
+        MapController.TryGetInstanceFromMap(mapId, mapInstanceId, out var instance);
+        return instance;
+    }
+
+    public void InvokeItemAdded(IItemSource source, IItem item)
```
lodicolo, line 23: This should stay here in the implementation only, because the server code will be able to call it off `Instance` which is typed to `MapHelper` instead of `IMapHelper`

Intersect.Server.Core/Plugins/Helpers/MapHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    private MapHelper() { }
+
+    public event Action<IItemSource, IItem> ItemAdded;
+
+    public IMapInstance GetMapInstanceByDescriptorId(Guid mapId, Guid mapInstanceId)
+    {
+        MapController.TryGetInstanceFromMap(mapId, mapInstanceId, out var instance);
+        return instance;
+    }
```
lodicolo, line 21: See `IMapHelper` for changes to this method

Intersect.Server.Core/Plugins/Helpers/MapHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+
+public partial class MapHelper : IMapHelper
+{
+    private static readonly Lazy<MapHelper> _instance = new Lazy<MapHelper>(() => new MapHelper());
+
+    public static MapHelper Instance => _instance.Value;
+
+    private MapHelper() { }
+
+    public event Action<IItemSource, IItem> ItemAdded;
```
lodicolo, line 15: See `IMapHelper` for changes to this event

lodicolo (changes requested):

blinkuz:
Leaving this comment as a note to confirm that I've been testing the item source implementation in a test plugin, and everything has been working as expected.

lodicolo (approved):

PR #2454, opened 2025-01-06
fix: dont render paperdoll on face

WeylonSantana:


lodicolo (changes requested):
The check should be outside of the loop, why are we going to loop through a bunch of paperdoll layers that we're not even going to render?

lodicolo (approved):

PR #2461, opened 2025-01-11
chore: Refactor tool cursor updates in MapEditor

TimtyG:
- Removed `CursorUpdateTimer_Tick` method.
- Added global `CurrentTool` property with `ToolChanged` event to manage tool state.
- Reduced unnecessary calls to `SetCursorSpriteInGrid()`.


Arufonsu:
hmm :thinking: the cursor was supposed to be drawn whenever we hovered the map grid or we clicked it 
(but that doesn't seems to happen anymore)

https://github.com/user-attachments/assets/893966ad-e642-4e64-90c6-f9b0242b94aa

Arufonsu (approved):
everything works as intended after testing commit 29452a9, 
thanks for fixing this @TimtyG !

PR #2465, opened 2025-01-12
enhancement: Add target context menu item

TimtyG:
_targetNameMenuItem to display the target player's name in the context menu. Instead of using label, ensured  the menu item is non-interactive.

Co-authored-by @Arufonsu 

https://github.com/user-attachments/assets/0d5d4e2f-de36-405c-83c6-92135c4e2abf

### DIFF
--- a/Intersect.Client.Core/Interface/Game/TargetContextMenu.cs
+++ b/Intersect.Client.Core/Interface/Game/TargetContextMenu.cs
@@ -37,9 +37,6 @@ public sealed partial class TargetContextMenu : Framework.Gwen.Control.Menu
         _me = Globals.Me;
 
         _targetNameMenuItem = AddItem(string.Empty);
-        _targetNameMenuItem.SetText("");
-        _targetNameMenuItem.MouseInputEnabled = false;
-
         AddDivider();
 
         _tradeMenuItem = AddItem(Strings.EntityContextMenu.Trade);
@@ -67,6 +64,8 @@ public sealed partial class TargetContextMenu : Framework.Gwen.Control.Menu
             return;
         }
 
+        bool shouldShowTargetNameMenuItem = false;
+
         float posX, posY, newX, newY;
 
         switch (target)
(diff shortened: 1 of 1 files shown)


Intersect.Client.Core/Interface/Game/TargetContextMenu.cs
```diff
@@ -116,6 +132,21 @@ public void ToggleHidden(object? target)
...
+    {
+        _targetNameMenuItem.SetText(shouldShow ? _entity.Name : string.Empty);
+
+        if (shouldShow && !Children.Contains(_targetNameMenuItem))
+        {
+            Children.Insert(0, _targetNameMenuItem);
+        }
+        else if (!shouldShow && Children.Contains(_targetNameMenuItem))
+        {
+            Children.Remove(_targetNameMenuItem);
```
lodicolo, line 146: Just do !shouldShow, `Remove()` is a safe method to call without Contains first

Intersect.Client.Core/Interface/Game/TargetContextMenu.cs
```diff
@@ -116,6 +132,21 @@ public void ToggleHidden(object? target)
...
     }
 
+
+    private void TryShowTargetButton(bool shouldShow)
+    {
+        _targetNameMenuItem.SetText(shouldShow ? _entity.Name : string.Empty);
+
+        if (shouldShow && !Children.Contains(_targetNameMenuItem))
+        {
+            Children.Insert(0, _targetNameMenuItem);
```
lodicolo, line 142: do the following pseudocode instead to account for all of the cases exhaustively

```
if (shouldShow) {
   indexOf
   if (indexOf > 0) remove it // just remove, it's inserted in the correct location on the next line
   if (indexOf != 0) insert at 0 // (this takes into account both cases where it's not present at all, and is present but in the wrong place
}
```

lodicolo (changes requested):

Intersect.Client.Core/Interface/Game/TargetContextMenu.cs
TimtyG: based on what you proposed bellow,  else would be enough?

lodicolo: Yup, I just happened to write the pseudocode afterwaters

PR #2474, opened 2025-01-14
fix: some npc behaviors

### LINKED ISSUE #2460: bug: NPCs spells can't hit steath players
### Description

When a player uses steath, the NPCs' abilities do not cause damage to them.
But when another player uses some skill, be it AoE or projectile that hits the player with steath, he will take damage and appear.

### Steps to Reproduce

1 - Create a steath skill for the player
2 - Create a aoe and projectile skill for an NPC
3 - make a npc use that skilll
4 - be at position to get damage and use the Stealth
5 - see the player don't get the hit and ignore the spell damage from the monsters
6 - Try step 3 with another character (PvP) and see the stealth be canceled when the steath players being hitted.

### Version with bug

Latest version

Unknow

### Affected platforms

Windows, Linux, I was *not* able test on other platforms

Can't find any workaround.

### Relevant log output

_No response_

### Duplicate Bug Check

WeylonSantana:
resolves #2460 

1. Hit entity when entity is invisble in 3 cases
  a. when using projectiles spells
  b. when using aoe spells
  c. when using single target spell on a visible player, but there is player in stealth inside hit radius
2. Dont keep looking at the entity if the entity is invisible
3. If there is another valid target on range when the current target becomes invisible, change target.

### DIFF
--- a/Intersect.Server.Core/Entities/Entity.cs
+++ b/Intersect.Server.Core/Entities/Entity.cs
@@ -48,7 +48,7 @@ public abstract partial class Entity : IEntity
     public IReadOnlyDictionary<Vital, long> VitalsLookup => _vitals.Select((value, index) => (value, index))
         .ToDictionary(t => (Vital)t.index, t => t.value).AsReadOnly();
 
-    [NotMapped, JsonIgnore] public Entity Target { get; set; } = null;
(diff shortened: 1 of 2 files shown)


dhmello:
Does this also work with the isHide function used by events?

lodicolo (changes requested):

PR #2486, opened 2025-01-20
feat: random drops

### LINKED ISSUE #2488: feature: ability to set amount range for items drop within the NPC editor
### Discussed in https://github.com/AscensionGameDev/Intersect-Engine/discussions/2209

<sup>Originally posted by **Superspyi** May 21, 2020</sup>

Yes. What I would like to do is make an NPC drop a random amount of an item within a certain range. For Example:

- For early game NPCs I would like for them to drop between 1-5 gold as one item to pickup.
- End game bosses would be much larger between 1200-2400 gold.

Currently I when I create an NPC with drops such as gold I have to set a static amount of gold to drop at a certain percentage. The only way I can get close to what I'd like is to create multiple drops of the same item of 1x at the same percentage but then the user would have to pick up multiple items. This becomes impossible for something like end game bosses.

Introduce a "range" option near the amount section in NPC Editor.

**Additional context**
I believe this would be a great feature to add into the Intersect engine due to most RPGs I've played before having random amounts of currency or items on NPCs when defeated such as $ in RDR2 or Gold in Oblivion & Skyrim. Attached is statistics for gold from an NPC on another MMORPG that I play that includes amount ranges.
AmountRange
</div>

WeylonSantana:
Resolves #2488 

The default Amount for all games now becomes MaxQuantity in the logic of this pr, but nothing has been renamed. So after this pr, all games will drop a minimum of 1 - Amount when they update.

https://github.com/user-attachments/assets/cc991795-4cff-419f-8cf2-3eb8fd3c7a14

### DIFF
--- a/Intersect.Editor/Localization/Strings.cs
+++ b/Intersect.Editor/Localization/Strings.cs
@@ -4465,7 +4465,9 @@ Tick timer saved in server config.json.";
 
         public static LocalizedString dropitem = @"Item:";
 
-        public static LocalizedString dropamount = @"Amount:";
+        public static LocalizedString DropMaxAmount = @"Max Amount:";
+
+        public static LocalizedString DropMinAmount = @"Min Amount:";
 
         public static LocalizedString dropchance = @"Chance (%):";
 
@@ -4473,7 +4475,7 @@ Tick timer saved in server config.json.";
 
         public static LocalizedString dropremove = @"Remove";
 
-        public static LocalizedString dropdisplay = @"{00} x{01} - {02}%";
+        public static LocalizedString dropdisplay = @"{00} x{01} - {02} | {03}%";
 
         public static LocalizedString enabled = @"Enabled?";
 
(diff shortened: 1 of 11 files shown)


Intersect.Editor/Localization/Strings.cs
```diff
@@ -4940,15 +4942,17 @@ public partial struct ResourceEditor
...
+        public static LocalizedString DropMinAmount = @"Min Amount:";
 
         public static LocalizedString dropchance = @"Chance (%):";
 
         public static LocalizedString dropadd = @"Add";
 
         public static LocalizedString dropremove = @"Remove";
 
-        public static LocalizedString dropdisplay = @"{00} x{01} - {02}%";
+        public static LocalizedString dropdisplay = @"{00} x{01} - {02} | {03}%";
```
lodicolo, line 4955: Change to `{00} x{01}-{02} | {03}%`

Intersect.Editor/Localization/Strings.cs
```diff
@@ -4465,15 +4465,17 @@ public partial struct NpcEditor
...
```
lodicolo, line 4478: Change to `{00} x{01}-{02} | {03}%`

lodicolo (changes requested):

lodicolo (commented):
I made a minor QoL change to the add button in frmNpc, it looks like frmResource just needs to be updated to match frmNpc and then this is good to go

lodicolo (approved):

PR #2492, opened 2025-01-22
Add level down support

### LINKED ISSUE #2491: Negative xp support.
### Discussed in https://github.com/AscensionGameDev/Intersect-Engine/discussions/1492

<sup>Originally posted by **WeylonSantana** February  8, 2021</sup>

Change the give exp, to "set exp", and then accept positive and negative values to be added to or subtracted from the current xp.

Ps: Due to the #384 / #381 , currently is possible to remove xp from the player, just set a variable to a negative value and give xp to the player based on the value of the variable. (no errors, no bugs).

Which player reaches 0 exp. It is no longer possible to remove exp. (In case to make it return to the previous level.)

I am opening this feature request so that there may be (who knows) in the future the feature to subtract exp in the "normal" way too, in addition to the level being subtracted when the player reaches 0 exp when another attempt to subtract exp is executed. (in this case he would return to the previous level with 99% of the xp needed to return to the level he was in).

Another reason is that you probably didn't program this feature to subtract exp either, and then I would be using the engine in a way that I shouldn't (among many other things I use, in the way that I shouldn't)

I ended up quoting the solution above

**Additional context**
Add any other context or screenshots about the feature request here.</div>

WeylonSantana:
resolves #2491

- refactor give exp editor
- adding support for decreasing the player's level using the give exp command
- added a new event common trigger

https://github.com/user-attachments/assets/a0e0b033-a8a6-4aa8-87a5-07a4a787ab9e

### DIFF
--- a/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
+++ b/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
@@ -460,7 +460,7 @@ public static partial class CommandProcessing
         }
         else if (quantity < 0)
         {
-            player.TakeExperience(Math.Abs(quantity), command.EnableLevelDown);
+            player.TakeExperience(Math.Abs(quantity), command.EnableLosingLevels);
         }
     }
 
--- a/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
+++ b/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
@@ -421,7 +421,7 @@ public static partial class CommandProcessing
         Stack<CommandInstance> callStack
     )
     {
-        player.LevelUp();
+        player.AddLevels();
     }
 
     //Give Experience Command
@@ -454,7 +454,14 @@ public static partial class CommandProcessing
             }
         }
 
-        player.GiveExperience(quantity);
+        if(quantity > 0)
+        {
+            player.GiveExperience(quantity);
+        }
+        else if (quantity < 0)
+        {
+            player.TakeExperience(Math.Abs(quantity), command.EnableLevelDown);
+        }
     }
 
     //Change Level Command
--- a/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
+++ b/Intersect.Server.Core/Entities/Events/CommandProcessing.cs
@@ -460,7 +460,7 @@ public static partial class CommandProcessing
         }
         else if (quantity < 0)
         {
-            player.TakeExperience(Math.Abs(quantity), command.EnableLosingLevels);
+            player.TakeExperience(Math.Abs(quantity), command.EnableLosingLevels, force: true);
         }
     }
 
--- a/Intersect.Server.Core/Entities/Player.cs
+++ b/Intersect.Server.Core/Entities/Player.cs
@@ -1151,16 +1151,10 @@ public partial class Player : Entity
         {
             if (Options.Instance.Player.ExpLossOnDeathPercent > 0)
             {
-                if (Options.Instance.Player.ExpLossFromCurrentExp)
-                {
-                    var ExpLoss = (this.Exp * (Options.Instance.Player.ExpLossOnDeathPercent / 100.0));
-                    TakeExperience((long)ExpLoss);
-                }
-                else
-                {
-                    var ExpLoss = (GetExperienceToNextLevel(this.Level) * (Options.Instance.Player.ExpLossOnDeathPercent / 100.0));
-                    TakeExperience((long)ExpLoss);
-                }
+                var baseExp = Options.Instance.Player.ExpLossFromCurrentExp ? Exp : GetExperienceToNextLevel(Level);
+                var expRatioLost = Options.Instance.Player.ExpLossOnDeathPercent / 100.0;
+                var expLost = (long)(baseExp * expRatioLost);
+                TakeExperience(expLost);
             }
         }
         PacketSender.SendEntityDie(this);
--- a/Intersect.Server.Core/Entities/Player.cs
+++ b/Intersect.Server.Core/Entities/Player.cs
@@ -1349,7 +1349,7 @@ public partial class Player : Entity
                         ForgetSpell(FindSpell(spell.Id), true);
                         PacketSender.SendChatMsg(
                             this,
-                            Strings.Player.SpellForgottenLevelDown.ToString(SpellBase.GetName(spell.Id)),
+                            Strings.Player.ForgotSpell.ToString(SpellBase.GetName(spell.Id)),
                             ChatMessageType.Experience,
                             CustomColors.Alerts.Info,
                             Name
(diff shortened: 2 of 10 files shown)


Intersect.Server.Core/Localization/Strings.cs
```diff
@@ -1097,6 +1100,9 @@ public sealed partial class PlayerNamespace : LocaleNamespace
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public readonly LocalizedString Left = @"{00} has left {01}.";
 
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public readonly LocalizedString LevelDown = @"You have leveled down! You are now level {00}!";
```
lodicolo, line 1104: This isn't super natural in english, it would be better to say `You have lost a level! You are now level {00}!`

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1347,20 +1394,39 @@ public void GiveExperience(long amount)
...
-        }
+            if (!enableLevelDown || Level == 1)
+            {
+                Exp = 0;
+                PacketSender.SendExperience(this);
+            }
+            else
+            {
+                var levelCount = 0;
+                while (Exp < 0 && Level - levelCount > 1)
```
lodicolo, line 1415: `+ levelCount`

Intersect.Server.Core/Localization/Strings.cs
lodicolo: Also change the name of the string to `LevelLost`

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1275,51 +1276,100 @@ public void SetLevel(int level, bool resetExperience = false)
...
+                    foreach (var spell in classDescriptor.Spells)
                     {
-                        continue;
+                        if (spell.Level != Level)
+                        {
+                            continue;
+                        }
+
+                        var spellInstance = new Spell(spell.Id);
+                        _ = TryTeachSpell(spellInstance, true);
```
lodicolo, line 1311: Where's the message for learning spells? Why is there only a message for forgetting them in this method?

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1275,51 +1276,100 @@ public void SetLevel(int level, bool resetExperience = false)
...
+
+                PacketSender.SendChatMsg(
+                    this,
+                    Strings.Player.LevelUp.ToString(Level),
+                    ChatMessageType.Experience,
+                    CustomColors.Combat.LevelUp,
+                    Name
+                );
+
+                if (ClassBase.TryGet(ClassId, out var classDescriptor) && classDescriptor?.Spells != default)
```
lodicolo, line 1301: Put `ClassBase? classDescriptor = null;` before `if (levels > 0)`

Here, change the `if` to `if ((classDescriptor?.Id == ClassId || ClassBase.TryGet(ClassId, out classDescriptor)) && classDescriptor?.Spells != default)`

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1275,51 +1276,100 @@ public void SetLevel(int level, bool resetExperience = false)
...
+
+                PacketSender.SendChatMsg(
+                    this,
+                    Strings.Player.LevelDown.ToString(Level),
+                    ChatMessageType.Experience,
+                    CustomColors.Combat.LevelDown,
+                    Name
+                );
+
+                if (ClassBase.TryGet(ClassId, out var classDescriptor) && classDescriptor?.Spells != default)
```
lodicolo, line 1340: Here, change the `if` to `if ((classDescriptor?.Id == ClassId || ClassBase.TryGet(ClassId, out classDescriptor)) && classDescriptor?.Spells != default)`

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1347,20 +1394,39 @@ public void GiveExperience(long amount)
         }
     }
 
-    public void TakeExperience(long amount)
+    public void TakeExperience(long amount, bool enableLevelDown = false)
```
lodicolo, line 1397: Add `, bool force = false` as the final parameter

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1347,20 +1394,39 @@ public void GiveExperience(long amount)
...
     {
-        if (this is Player && Options.Instance.Map.DisableExpLossInArenaMaps && Map.ZoneType == MapZone.Arena)
+        if (Options.Instance.Map.DisableExpLossInArenaMaps && Map.ZoneType == MapZone.Arena)
```
lodicolo, line 1399: Change this to `if (force || (Options.Instance.Map.DisableExpLossInArenaMaps && Map.ZoneType == MapZone.Arena))`

lodicolo: Events taking experience should not be stopped by map type

Intersect.Server.Core/Entities/Events/CommandProcessing.cs
```diff
@@ -454,7 +454,14 @@ Stack<CommandInstance> callStack
...
         }
 
-        player.GiveExperience(quantity);
+        if(quantity > 0)
+        {
+            player.GiveExperience(quantity);
+        }
+        else if (quantity < 0)
+        {
+            player.TakeExperience(Math.Abs(quantity), command.EnableLevelDown);
```
lodicolo, line 463: add `, force: true` to the end of the args

lodicolo (changes requested):

Intersect.Server.Core/Entities/Player.cs
WeylonSantana: It is within the method itself (TryTeachSpell), when I tested it, before recording the video, the messages when learning magic when passing a level were duplicated, as this method already sends a message when teaching

Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1347,20 +1394,39 @@ public void GiveExperience(long amount)
...
+                    levelCount++;
+                }
 
-        PacketSender.SendExperience(this);
+                AddLevels(-levelCount);
+
+                if (Exp < 0)
+                {
+                    Exp = 0;
+                    PacketSender.SendExperience(this);
```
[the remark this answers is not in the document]
WeylonSantana: Any exp for any other level is sent at the end of AddLevels()
I added this if because, when I went down a level and was at level 1, after AddLevels() was executed, the exp would become negative if I removed more than I had

So I added this if as "if after removing the player's levels, the exp remains negative, reset and send"
the exp will never be negative after AddLevels(), if when the function finishes its level is above 1
It only happened exclusively in this situation, because the exp is not reset when leveling down, and it does not reset when it is negative when the player is at level 1.

Intersect.Server.Core/Entities/Player.cs
WeylonSantana: fixed now

lodicolo (approved):

Issue #2542, opened 2025-02-09
bug: Event Command "Give Experience" Experience Reduction


RecoStar:
### Description

The ability to remove player experience seems bugged. Reported by a different user as I dont use this, but on testing I was not able to remove player experience using a map or common event. 

When you try to reduce experience and have lose levels toggled OFF, it wont remove experience.
When you try to reduce experience and have lose levels toggled ON, it will sometimes just remove a level and not any experience.

### Steps to Reproduce

Add a map or common event.
Add command "Give Experience"
Set it to a minus number "-100"
Trigger Event

### Version with bug

v0.8.0-beta.373

### Last version that worked well

Not sure, never used it. 

### Affected platforms

Windows

### Did you find any workaround?

No

### Relevant log output

```shell

```

### Duplicate Bug Check

- [x] This bug report is not a duplicate to the best of my knowledge.

PR #2569, opened 2025-02-13
Document and Refactor of Player Leveling logic

DerekGooding:
I was trying to solve issue #2542 but there was nothing glaring in Player. 

I did a bit of refactoring to remove some redundant calls to Packet. Also documented the methods I touched. 
Hopefully this helps find the bug where it might be elsewhere.


Intersect.Server.Core/Entities/Player.cs
```diff
@@ -1395,19 +1395,19 @@ public void TakeExperience(long amount, bool enableLosingLevels = false, bool fo
...
+                    Exp += GetExperienceToNextLevel(Level - levelsToRemove);
                 }
 
                 if (Exp < 0)
                 {
                     Exp = 0;
                 }
 
-                AddLevels(levelCount); //AddLevels includes PacketSender.SendExperience(this);
+                AddLevels(-levelsToRemove); //AddLevels includes PacketSender.SendExperience(this);
```
WeylonSantana: I wrote the code as you wrote it now, at the beginning, but panda told me to revert it to how it is currently, so he won't approve this change

WeylonSantana:
We appreciate your change and contribution, but I would also like you to send a video showing that all behaviors are working as they should, including the correction of the reported bug, but the final word here is Panda.

DerekGooding:
Seems like I've wasted more time than helped with. Rejection is fine. Thanks for looking at it.

lodicolo (approved):

Issue #2546, opened 2025-02-12
bug: Escape menu bad alignment


lodicolo:
### Description

![Image](https://github.com/user-attachments/assets/38651acb-2f06-42be-9d91-fbeab8d2dfd3)

### Steps to Reproduce

N/A

### Version with bug

v0.8.0.373+

### Last version that worked well

N/A

### Affected platforms

I was *not* able test on other platforms

### Relevant log output

```shell

```

### Duplicate Bug Check

- [x] This bug report is not a duplicate to the best of my knowledge.

PR #2576, opened 2025-02-15
Escape menu cleanup

DerekGooding:
This should fix the layout so it looks like this now:

image

Might need to make an update to the Asset repo to reflect these changes. The label edits are overridden by the EscapeMenu.json unless you completely comment out the label, compile and run, then uncomment. 

Maybe I'll take a look at how the load and save json works later. Save that headache. 

Fixes Issue #2546

### DIFF
--- a/Intersect.Client.Core/Interface/Game/EscapeMenu.cs
+++ b/Intersect.Client.Core/Interface/Game/EscapeMenu.cs
@@ -35,6 +35,7 @@ public partial class EscapeMenu : ImagePanel
 
         var buttonSettings = new Button(container, "SettingsButton")
         {
+            IsTabable = true,
             Text = Strings.EscapeMenu.Settings,
         };
         buttonSettings.Clicked += (s, e) => OpenSettingsWindow(true);
@@ -42,6 +43,7 @@ public partial class EscapeMenu : ImagePanel
         // Character Select Button
         _buttonCharacterSelect = new Button(container, "CharacterSelectButton")
         {
+            IsTabable = true,
             Text = Strings.EscapeMenu.CharacterSelect,
         };
         _buttonCharacterSelect.Clicked += _buttonCharacterSelect_Clicked;
@@ -49,6 +51,7 @@ public partial class EscapeMenu : ImagePanel
         // Logout Button
         var buttonLogout = new Button(container, "LogoutButton")
         {
+            IsTabable = true,
             Text = Strings.EscapeMenu.Logout,
         };
         buttonLogout.Clicked += buttonLogout_Clicked;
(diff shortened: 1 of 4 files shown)


Intersect.Client.Core/Interface/Game/GameInterface.cs
```diff
@@ -92,21 +92,22 @@ public partial class GameInterface : MutableInterface
...
+        }
     }
 
     public GameInterface(Canvas canvas) : base(canvas)
     {
         GameCanvas = canvas;
-        EscapeMenu = new EscapeMenu(GameCanvas, GetOrCreateSettingsWindow) {IsHidden = true};
-        SimplifiedEscapeMenu = new SimplifiedEscapeMenu(GameCanvas, GetOrCreateSettingsWindow) {IsHidden = true};
+        EscapeMenu = new EscapeMenu(GameCanvas, SettingsWindow) {IsHidden = true};
+        SimplifiedEscapeMenu = new SimplifiedEscapeMenu(GameCanvas, SettingsWindow) {IsHidden = true};
```
lodicolo, line 110: tl;dr: Revert the changes in this file

`SettingsWindow` is a bit heavy and I intentionally wanted to defer creation until it's needed. To that end, passing a delegate was necessary and the getter, while on the surface is doing the same thing, isn't actually having the same effect.

In this case it's creating only one, but it's creating it immediately in the constructor of `GameInterface`. With the methods, it was only ever created when `_settingsWindowProvider()` was invoked.

Intersect.Client.Core/Interface/Game/SimplifiedEscapeMenu.cs
lodicolo, line 1: Also revert, since the only changes are `SettingsWindow` related

Intersect.Client.Core/Interface/Game/EscapeMenu.cs
```diff
@@ -11,60 +12,62 @@ namespace Intersect.Client.Interface.Game;
...
     private readonly Button _buttonCharacterSelect;
     private readonly Panel _versionPanel;
 
-    public EscapeMenu(Canvas gameCanvas, Func<SettingsWindow> settingsWindowProvider) : base(gameCanvas, nameof(EscapeMenu))
+    private readonly Framework.Graphics.GameFont _font;
+
+    public EscapeMenu(Canvas gameCanvas, SettingsWindow settingsWindow) : base(gameCanvas, nameof(EscapeMenu))
     {
-        _settingsWindowProvider = settingsWindowProvider;
+        _settingsWindow = settingsWindow;
```
lodicolo, line 23: We definitely want to continue using the delegate, revert the SettingsWindow changes here (but keep the font one)

lodicolo (changes requested):

Intersect.Client.Core/Interface/Game/EscapeMenu.cs
DerekGooding: That makes sense. I hadn't considered the lazy implementation. I guess I assumed settingsmenu would be called long before the EscapeMenu would be constructed. But you're correct. If someone plays the game and exits without ever hitting settings, the delegate version saves having to load the settings menu.

lodicolo (approved):

DerekGooding:
This is no longer required. New PR obsoletes these changes.

PR #2590, opened 2025-02-17
Escape menu window

DerekGooding:
3rd times the charm. 

Create a new Escape Menu window using the designer pattern. Call it instead of the old version.


DerekGooding:
Fixes issue #2546

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            _settings.SetToolTipText(Strings.MainMenu.SettingsTooltip);
+        }
+
+        if (Options.Instance.Player.MaxCharacters <= 1)
+        {
+            _charselect.IsDisabled = true;
+        }
+
+        _versionPanel = new VersionPanel(canvas, name: nameof(_versionPanel));
```
lodicolo, line 82: Is the version panel showing up underneath the modal?

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
```diff
@@ -0,0 +1,211 @@
...
+
+namespace Intersect.Client.Interface.Menu;
+
+public partial class EscapeMenuWindow : Window
+{
+    private readonly Button _settings;
+    private readonly Button _charselect;
+    private readonly Button _logout;
+    private readonly Button _exitdesktop;
+    private readonly Button _closemenu;
```
lodicolo, line 18: Can you make the names be camel-case and more thorough? e.g. `_returnToCharacterSelectButton`

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
```diff
@@ -0,0 +1,211 @@
...
+
+    private readonly Panel _versionPanel;
+
+    private readonly Func<SettingsWindow> _settingsWindowProvider;
+
+    public EscapeMenuWindow(Canvas canvas, Func<SettingsWindow> settingsWindowProvider)
+        : base(canvas, Strings.EscapeMenu.Title, false,
+            $"{nameof(EscapeMenuWindow)}_{(ClientContext.IsSinglePlayer ? "singleplayer" : "online")}")
```
lodicolo, line 26: Good catch, I didn't even think about this having two separate layouts

lodicolo (changes requested):
Shouldn't the original `EscapeMenu` class be deleted in this PR too?

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
DerekGooding: These are tied to the UI files. 

I can, but then I have to hardcode the string values.

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
DerekGooding: I'll look into this. The only escapemenu had it float at the very bottom of the screen but the window isn't large enough with this version. I'll likely create an extra imagepanel and add it as a child of that.

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
lodicolo: I realize they're tied to the UI files, but I'm ok with invalidating existing stuff (especially since the entire UI system needs to be replaced anyway).

Intersect.Client.Core/Interface/Menu/EscapeMenuWindow.cs
lodicolo: The version panel needs to be a child of the modal in this case, rather than another `ImagePanel`.

lodicolo (approved):

PR #2616, opened 2025-02-22
fix: select character arrow visibility

### LINKED ISSUE #2604: bug: Select char arrows appear when they shouldn't
### Description

When the maximum number of characters on the account is 1 (intersect's default setting), the character selection arrows appear when they shouldn't.

https://github.com/user-attachments/assets/2b5b8147-9d65-4b77-81a6-5c03cf1c9911

### Steps to Reproduce

1. Default server configuration
2. Start the game and create a character
3. Select character and see arrows that shouldn't exist

### Version with bug

v0.8.0.413-beta+c8c78ba27fcd3f98468664156635adde6309a95a

Unknown

### Affected platforms

Windows 10

Clicking on them makes them disappear.

### Relevant log output

```shell

```

### Duplicate Bug Check

WeylonSantana:
resolves #2604 

UpdateDisplay() is called before EnsureEnsureInitialized()

Update display hides the arrows, when the condition is true, EnsureEnsureInitialized loads the json value which is false, so the arrows appear when they shouldn't

### DIFF
--- a/Intersect.Client.Core/Interface/Menu/SelectCharacterWindow.cs
+++ b/Intersect.Client.Core/Interface/Menu/SelectCharacterWindow.cs
@@ -173,7 +173,7 @@ public partial class SelectCharacterWindow : Window
         SizeToChildren(recursive: true);
 
         LoadJsonUi(GameContentManager.UI.Menu, Graphics.Renderer?.GetResolutionString());
-        _ensureArrowsVisibility();
+        EnsureArrowsVisibility();
     }
 
     //Methods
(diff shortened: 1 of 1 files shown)


Intersect.Client.Core/Interface/Menu/SelectCharacterWindow.cs
```diff
@@ -173,6 +173,14 @@ protected override void EnsureInitialized()
...
 
         LoadJsonUi(GameContentManager.UI.Menu, Graphics.Renderer?.GetResolutionString());
+
+        // in case the json load is changing the visibility of the arrows when it shouldn't
+        // lets ensure the right value
+        if (CharacterSelectionPreviews != default)
+        {
+            _selectCharacterRightButton.IsHidden = CharacterSelectionPreviews.Length <= 1;
+            _selectCharacterLeftButton.IsHidden = CharacterSelectionPreviews.Length <= 1;
+        }
```
lodicolo, line 183: Move this all to `Show()` instead, otherwise they will display wrong if someone logs in, the server closes, the configuration is changed, and then they re-log in

lodicolo (changes requested):

PR #2639, opened 2025-03-06
Have the client independently spawn the resource exhaustion animation to avoid having frames between when the animation begins playing and when the resource's graphic changes

### LINKED ISSUE #2572: bug: Resource Animation Delay
### Description

Sometimes, a flicker can be noticed when the animation plays while harvesting/cutting a resource. This flicker doesn't always occur, as discussed on Discord; it is likely caused by the server sending the animation, as seen in this part of the [Resource](https://github.com/AscensionGameDev/Intersect-Engine/blob/904037bb5874355878c8aa9b9a7631f8aec6cd19/Intersect.Server.Core/Entities/Resource.cs#L68) entity code on the server.

I'll leave these GIFs here in case they serve as a visual reference to identify the bug:

https://github.com/user-attachments/assets/e5f0a5cc-3c32-4368-bd51-b16699da9d91

### Steps to Reproduce

1-Assign an animation to a resource
2-Harvest/cut that resource
3-Look for the flicker; if it doesn’t appear, repeat step 2, as my tests show it occurs 4 out of 10 times.

### Version with bug

v0.8.0.0-custom+2376987dbc362d7a2fdb0bc9550ab73ec9a7554a

Unknown

### Affected platforms

Windows 10

It's possible that the issue is related to the resource animation being sent from the server. You can find the code [here](https://github.com/AscensionGameDev/Intersect-Engine/blob/904037bb5874355878c8aa9b9a7631f8aec6cd19/Intersect.Server.Core/Entities/Resource.cs#L68)

### Duplicate Bug Check

Files: [resource_animation_delay.zip](https://github.com/user-attachments/files/19070080/resource_animation_delay.zip)

lodicolo:
Resolves #2572

### DIFF
--- a/Intersect.Client.Core/Entities/Animation.cs
+++ b/Intersect.Client.Core/Entities/Animation.cs
@@ -14,9 +14,12 @@ namespace Intersect.Client.Entities;
 
 public partial class Animation : IAnimation
 {
+    public event Action<IAnimation>? Disposed;
+    public event Action<IAnimation>? Finished;
+
     public bool AutoRotate { get; set; }
 
-    private bool disposed = false;
+    private bool _disposed = false;
 
     public bool Hidden { get; set; }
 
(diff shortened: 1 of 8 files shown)


lodicolo:
https://github.com/user-attachments/assets/6785f6f4-6bbc-4f0f-9383-445b79ac418b

PR #2653, opened 2025-03-11
feat: is on combat condition

### LINKED ISSUE #2652: In Combat Conditional?
### Discussed in https://github.com/AscensionGameDev/Intersect-Engine/discussions/1487

<sup>Originally posted by **Richy1111** October 22, 2020</sup>
I was wondering if we can maybe have an 'In Combat' conditional check for events. I don't want to lock a player inside a PvP Arena but I want to make sure they can't leave that Arena while in combat either?
Thanks.</div>

WeylonSantana:
resolves #2652

image
image

https://github.com/user-attachments/assets/01630594-918f-4177-bc12-12548ffacd62

### DIFF
--- a/Framework/Intersect.Framework.Core/GameObjects/Conditions/ConditionType.cs
+++ b/Framework/Intersect.Framework.Core/GameObjects/Conditions/ConditionType.cs
@@ -40,5 +40,5 @@ public enum ConditionType
 
     CheckEquipment,
 
-    IsOnCombat,
+    IsInCombat,
 }
--- a/Framework/Intersect.Framework.Core/GameObjects/Conditions/ConditionType.cs
+++ b/Framework/Intersect.Framework.Core/GameObjects/Conditions/ConditionType.cs
@@ -39,4 +39,6 @@ public enum ConditionType
     MapZoneTypeIs,
 
     CheckEquipment,
+
+    IsOnCombat,
 }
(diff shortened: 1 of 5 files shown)


Framework/Intersect.Framework.Core/GameObjects/Conditions/ConditionType.cs
```diff
@@ -39,4 +39,6 @@ public enum ConditionType
     MapZoneTypeIs,
 
     CheckEquipment,
+
+    IsOnCombat,
```
lodicolo, line 43: Should be `IsInCombat`

Intersect.Editor/Localization/Strings.cs
```diff
@@ -364,6 +365,11 @@ public static string GetEventConditionalDesc(CheckEquippedSlot condition)
         return EventConditionDesc.checkequippedslot.ToString(condition.Name);
     }
 
+    public static string GetEventConditionalDesc(CombatCondition condition)
+    {
+        return EventConditionDesc.Combat.ToString();
+    }
+
```
lodicolo, line 372: I don't see this being used, what is it for?

lodicolo (changes requested):

Intersect.Editor/Localization/Strings.cs
lodicolo: This is being involved dynamically :grimacing:

lodicolo (approved):

PR #2667, opened 2025-03-13
Refactor drag and drop

WeylonSantana:
redoing the drag and drop to use the native gwen and not a bunch of calculations checking intersections of rectangles

What's going on here

Gwen's drag and drop system works based on "sending packages" and "receiving packages", with some situations where you can benefit from such as starting the drag, accepting the packet, when it drops (these are not events, unfortunately)

Anyone who wants to create their draggable image component just needs to inherit from Draggable, so I made all the IconImages of the SlotsItens as Draggables, which already comes with a "default configuration" of what a draggable component should be

Furthermore, in the Gwen Base component, I made it so that all components are able to accept any package, but no real drop will happen if the

`public override bool DragAndDrop_HandleDrop(Package package, int x, int y)`

is not implemented in child, which is responsible for returning a response to the component being dragged whether the drop was successful or not.

The basic implementation I did for all items was a loop, starting from the current component to the root, checking what the component is, because if it is acceptable something will be done and true will be returned, if not, when returning false, we cancel the drop.

For example, when dragging an IconImage from a BankItem to an empty inventory slot, this component is an InventoryItem, so it will be removed from the bank, but if the slot is occupied, then the target component was a Draggable, it could be a label too, and so on, which is why the previous Parents are checked, Parent will be InventoryItem and it will be removed from the bank.

I kept the default behavior above all, hiding labels and such when dragging, and I also added behavior to not close windows or hide UI when dragging which caused this bug (latest clean)

https://github.com/user-attachments/assets/a65378d3-ae55-4c51-a597-c5b54b7af68e

and finally I added a new function which is to drag the item into the shop window and it will be sold

full video of everything working

https://github.com/user-attachments/assets/c5ee723c-3317-4893-a0f2-238d339a6aba

### DIFF
--- a/Intersect.Client.Core/Core/Input.cs
+++ b/Intersect.Client.Core/Core/Input.cs
@@ -255,6 +255,11 @@ public static partial class Input
                         break;
                     }
 
+                    if (DragAndDrop.IsDragging)
+                    {
+                        break;
+                    }
+
                     switch (control)
                     {
                         case Control.Block:
--- a/Intersect.Client.Core/Core/Input.cs
+++ b/Intersect.Client.Core/Core/Input.cs
@@ -2,6 +2,7 @@ using Intersect.Admin.Actions;
 using Intersect.Client.Entities;
 using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Graphics;
+using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Gwen.Input;
 using Intersect.Client.Framework.Input;
 using Intersect.Client.General;
@@ -138,6 +139,11 @@ public static partial class Input
                 return;
             }
 
+            if (DragAndDrop.IsDragging)
+            {
+                return;
+            }
+
             var gameUi = Interface.Interface.GameUi;
 
             // First try and unfocus chat then close all UI elements, then untarget our target.. and THEN open the escape menu.
@@ -196,7 +202,7 @@ public static partial class Input
                     break;
 
                 case Control.ToggleGui:
-                    if (currentGameState == GameStates.InGame)
+                    if (currentGameState == GameStates.InGame && !DragAndDrop.IsDragging)
                     {
                         Interface.Interface.HideUi = !Interface.Interface.HideUi;
                     }
--- a/Intersect.Client.Core/Entities/Entity.cs
+++ b/Intersect.Client.Core/Entities/Entity.cs
@@ -7,7 +7,6 @@ using Intersect.Client.Framework.Content;
 using Intersect.Client.Framework.Entities;
 using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Graphics;
-using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Items;
 using Intersect.Client.Framework.Maps;
 using Intersect.Client.General;
@@ -1369,7 +1368,7 @@ public partial class Entity : IEntity
             // Checks if the entity is attacking or not.
             // Note: the calculation differs with IsAttacking because
             // frames are intended to behave differently with normal sprite-sheets.
-            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds && !DragAndDrop.IsDragging)
+            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds)
             {
                 frame = Options.Instance.Sprites.NormalAttackFrame;
             }
--- a/Intersect.Client.Core/Entities/Entity.cs
+++ b/Intersect.Client.Core/Entities/Entity.cs
@@ -7,6 +7,7 @@ using Intersect.Client.Framework.Content;
 using Intersect.Client.Framework.Entities;
 using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Graphics;
+using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Items;
 using Intersect.Client.Framework.Maps;
 using Intersect.Client.General;
@@ -1368,7 +1369,7 @@ public partial class Entity : IEntity
             // Checks if the entity is attacking or not.
             // Note: the calculation differs with IsAttacking because
             // frames are intended to behave differently with normal sprite-sheets.
-            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds)
+            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds && !DragAndDrop.IsDragging)
             {
                 frame = Options.Instance.Sprites.NormalAttackFrame;
             }
--- a/Intersect.Client.Core/Entities/Player.cs
+++ b/Intersect.Client.Core/Entities/Player.cs
@@ -7,7 +7,6 @@ using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Gwen.Control;
 using Intersect.Client.Framework.Gwen.Control.EventArguments;
 using Intersect.Client.Framework.Gwen.Control.EventArguments.InputSubmissionEvent;
-using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Input;
 using Intersect.Client.Framework.Items;
 using Intersect.Client.General;
@@ -1957,11 +1956,6 @@ public partial class Player : Entity, IPlayer
     public bool TryAttack()
     {
         if (IsAttacking || IsBlocking || (IsMoving && !Options.Instance.Player.AllowCombatMovement) || Globals.Me == default)
-        {
-            return false;
-        }
-
-        if (DragAndDrop.IsDragging)
         {
             return false;
         }
--- a/Intersect.Client.Core/Entities/Player.cs
+++ b/Intersect.Client.Core/Entities/Player.cs
@@ -7,6 +7,7 @@ using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Gwen.Control;
 using Intersect.Client.Framework.Gwen.Control.EventArguments;
 using Intersect.Client.Framework.Gwen.Control.EventArguments.InputSubmissionEvent;
+using Intersect.Client.Framework.Gwen.DragDrop;
 using Intersect.Client.Framework.Input;
 using Intersect.Client.Framework.Items;
 using Intersect.Client.General;
@@ -1956,6 +1957,11 @@ public partial class Player : Entity, IPlayer
     public bool TryAttack()
     {
         if (IsAttacking || IsBlocking || (IsMoving && !Options.Instance.Player.AllowCombatMovement) || Globals.Me == default)
+        {
+            return false;
+        }
+
+        if (DragAndDrop.IsDragging)
         {
             return false;
         }
--- a/Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
+++ b/Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
@@ -1,6 +1,5 @@
 using Intersect.Client.Core;
 using Intersect.Client.Framework.File_Management;
-using Intersect.Client.Framework.GenericClasses;
 using Intersect.Client.Framework.Gwen;
 using Intersect.Client.Framework.Gwen.Control;
 using Intersect.Client.Framework.Gwen.Control.EventArguments;
@@ -8,11 +7,8 @@ using Intersect.Client.Framework.Gwen.Input;
 using Intersect.Client.Framework.Input;
 using Intersect.Client.General;
 using Intersect.Client.Interface.Game.DescriptionWindows;
-using Intersect.Client.Interface.Game.Inventory;
 using Intersect.Client.Localization;
-using Intersect.Client.Networking;
 using Intersect.Configuration;
-using Intersect.Framework.Core;
 using Intersect.Framework.Core.GameObjects.Items;
 
 namespace Intersect.Client.Interface.Game.Bag;
@@ -22,19 +18,7 @@ public partial class BagItem : SlotItem
     // Controls
     private readonly Label _quantityLabel;
     private readonly BagWindow _bagWindow;
-    private Draggable? _dragIcon;
     private ItemDescriptionWindow? _descWindow;
-
-    // Drag Handling
-    public bool IsDragging;
-    private bool _canDrag;
-    private long _clickTime;
-    private bool _mouseOver;
-    private int _mouseX = -1;
-    private int _mouseY = -1;
-
-    // Data control
-    private string? _textureLoaded;
 
     // Context Menu Handling
     private readonly MenuItem _withdrawContextItem;
@@ -107,19 +91,13 @@ public partial class BagItem : SlotItem
             return;
         }
 
-        _mouseOver = true;
-        _canDrag = true;
         if (Globals.InputManager.IsMouseButtonDown(MouseButton.Left))
         {
-            _canDrag = false;
             return;
         }
 
-        if (_descWindow != default)
-        {
-            _descWindow.Dispose();
-            _descWindow = default;
-        }
+        _descWindow?.Dispose();
+        _descWindow = default;
 
         if (Globals.BagSlots is not { Length: > 0 } bagSlots)
         {
@@ -143,36 +121,22 @@ public partial class BagItem : SlotItem
 
     private void _iconImage_HoverLeave(Base sender, EventArgs arguments)
     {
-        _mouseOver = false;
-        _mouseX = -1;
-        _mouseY = -1;
-
-        if (_descWindow != default)
-        {
-            _descWindow.Dispose();
-            _descWindow = default;
-        }
+        _descWindow?.Dispose();
+        _descWindow = default;
     }
 
     private void _iconImage_Clicked(Base sender, MouseButtonState arguments)
     {
-        // ReSharper disable once SwitchStatementMissingSomeEnumCasesNoDefault
-        switch (arguments.MouseButton)
+        if (arguments.MouseButton is MouseButton.Right)
         {
-            case MouseButton.Right:
-                if (ClientConfiguration.Instance.EnableContextMenus)
-                {
-                    OpenContextMenu();
-                }
-                else
-                {
-                    _iconImage_DoubleClicked(sender, arguments);
-                }
-                break;
-
-            case MouseButton.Left:
-                _clickTime = Timing.Global.MillisecondsUtc + 500;
-                break;
+            if (ClientConfiguration.Instance.EnableContextMenus)
+            {
+                OpenContextMenu();
+            }
+            else
+            {
+                _iconImage_DoubleClicked(sender, arguments);
+            }
         }
     }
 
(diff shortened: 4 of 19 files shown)


Intersect.Client.Core/Core/Input.cs
```diff
@@ -138,6 +139,11 @@ public static void OnKeyPressed(Keys modifier, Keys key)
                 return;
             }
 
+            if (DragAndDrop.IsDragging)
+            {
+                return;
+            }
```
lodicolo, line 145: What is this supposed to accomplish?

Intersect.Client.Core/Core/Input.cs
```diff
@@ -196,7 +202,7 @@ public static void OnKeyPressed(Keys modifier, Keys key)
                     break;
 
                 case Control.ToggleGui:
-                    if (currentGameState == GameStates.InGame)
+                    if (currentGameState == GameStates.InGame && !DragAndDrop.IsDragging)
```
lodicolo, line 205: This one actually makes sense, but it's always true because of L142?

Intersect.Client.Core/Entities/Entity.cs
```diff
@@ -1368,7 +1369,7 @@ private int NormalSpriteAnimationFrame
             // Checks if the entity is attacking or not.
             // Note: the calculation differs with IsAttacking because
             // frames are intended to behave differently with normal sprite-sheets.
-            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds)
+            else if (AttackTimer - (CalculateAttackTime() / 2) > Timing.Global.Milliseconds && !DragAndDrop.IsDragging)
```
lodicolo, line 1372: Why is drag and drop code in `Entity.NormalSpriteAnimationFrame`???

lodicolo (changes requested):

Intersect.Client.Core/Entities/Player.cs
```diff
@@ -1960,6 +1961,11 @@ public bool TryAttack()
             return false;
         }
 
+        if (DragAndDrop.IsDragging)
+        {
+            return false;
+        }
```
lodicolo, line 1967: This does not belong here.

Presumably this is to stop the left click from triggering an attack when dragging and dropping, but this should be processed in the controls instead (i.e. if dragging and dropping, the controls binding should not return "active" for left mouse button). There shouldn't be controls logic in this method.

lodicolo (changes requested):

Intersect.Client.Core/Interface/Game/SlotItem.cs
```diff
@@ -17,7 +17,7 @@ public SlotItem(Base parent, string name, int index, ContextMenu? contextMenu) :
         Margin = new Margin(4);
         MouseInputEnabled = true;
 
-        _iconImage = new ImagePanel(this, "Icon")
+        IconImage = new Draggable(this, "Icon")
```
lodicolo, line 20: Rename to `Icon` and change the name to `nameof(Icon)`

lodicolo (changes requested):

Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
```diff
@@ -100,26 +87,20 @@ private void _withdrawMenuItem_Clicked(Base sender, MouseButtonState arguments)
...
             return;
         }
 
-        if (_descWindow != default)
-        {
-            _descWindow.Dispose();
-            _descWindow = default;
-        }
+        _descWindow?.Dispose();
+        _descWindow = default;
```
lodicolo, line 103: `_descriptionWindow`

Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
```diff
@@ -141,47 +122,69 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+            }
         }
     }
 
-    private void _iconImage_Clicked(Base sender, MouseButtonState arguments)
+    private void IconImage_DoubleClicked(Base sender, MouseButtonState arguments)
     {
-        // ReSharper disable once SwitchStatementMissingSomeEnumCasesNoDefault
-        switch (arguments.MouseButton)
+        if (Globals.InBag)
```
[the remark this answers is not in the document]
lodicolo: Also I don't think we even need the `Globals.InBag` check... the window shouldn't be open (meaning `BagItem` isn't visible) if it's not true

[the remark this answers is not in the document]
lodicolo: (So yeah, replace `Globals.InBag` with `arguments.MouseButton is MouseButton.Left`)

Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
```diff
@@ -141,47 +122,69 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+                default:
+                    targetNode = targetNode.Parent;
+                    break;
+            }
+            
+            // If we've reached the top of the tree, we can't drop here, so cancel drop
+            if (targetNode == null)
+            {
+                return false;
+            }
```
lodicolo, line 184: This is unnecessary, immediately after this if-statement it would hit the `while` condition and return false anyway

Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
```diff
@@ -141,47 +122,69 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
 
-    private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
+    #endregion
+
+    #region Drag and Drop
+
+    public override bool DragAndDrop_HandleDrop(Package package, int x, int y)
     {
-        if (Globals.InBag)
+        var targetNode = Interface.FindComponentUnderCursor(NodeFilter.None);
```
lodicolo, line 160: You don't need `NodeFilter.None`, the parameter is optional and `None` is the default.

Intersect.Client.Core/Interface/Game/Bag/BagItem.cs
```diff
@@ -201,194 +204,41 @@ public override void Update()
...
-            {
-                _descWindow.Dispose();
-                _descWindow = default;
-                _iconImage_HoverEnter(default, default);
-            }
+            return;
         }
 
-        if (!IsDragging)
+        var itemTexture = Globals.ContentManager?.GetTexture(Framework.Content.TextureType.Item, descriptor.Icon);
```
lodicolo, line 225: `Globals.ContentManager?` -> `GameContentManager.Current` (no `?`)

Intersect.Client.Core/Interface/Game/Bag/BagWindow.cs
```diff
@@ -87,16 +82,16 @@ public void Update()
...
+        // dont hide window if we are dragging something
+        if (Items.Any(c => c.IconImage == DragAndDrop.CurrentPackage?.DrawControl))
         {
-            X = ToCanvas(new Point(0, 0)).X - 4 / 2,
-            Y = ToCanvas(new Point(0, 0)).Y - 4 / 2,
-            Width = Width + 4,
-            Height = Height + 4
-        };
+            return;
+        }
```
lodicolo, line 93:
```csharp
if (DragAndDrop.CurrentPackage is { } package)
{
    if (Items.Any(c => c.IconImage == package.DrawControl))
    {
        return;
    }
}
```

Intersect.Client.Core/Interface/Game/Bank/BankWindow.cs
```diff
@@ -90,16 +90,16 @@ public void Update()
...
+        // dont hide window if we are dragging something
+        if (Items.Any(c => c.IconImage == DragAndDrop.CurrentPackage?.DrawControl))
         {
-            X = ToCanvas(new Point(0, 0)).X - 4 / 2,
-            Y = ToCanvas(new Point(0, 0)).Y - 4 / 2,
-            Width = Width + 4,
-            Height = Height + 4
-        };
+            return;
+        }
```
lodicolo, line 101: Refer to my comment in `BagWindow`

Intersect.Client.Core/Interface/Game/Draggable.cs
```diff
@@ -1,68 +1,43 @@
...
-using Intersect.Client.Framework.Input;
-using Intersect.Client.General;
 
 namespace Intersect.Client.Interface.Game;
 
-
-partial class Draggable
+public partial class Draggable(Base parent, string name) : ImagePanel(parent, name)
 {
+    public bool DisableDragAndDrop { get; set; } = false;
```
lodicolo, line 9: This property is not needed, `IsDisabled` exists and should be used

Intersect.Client.Core/Interface/Game/Draggable.cs
```diff
@@ -1,68 +1,43 @@
...
-
-    public int X
-    {
-        get => mPnl.X;
-        set => mPnl.X = value;
-    }
+    public bool IsDragging => DragAndDrop.CurrentPackage?.DrawControl == this;
 
-    public int Y
+    public override bool DragAndDrop_Draggable()
```
lodicolo, line 13: Please add a `// TODO: Fix drag and drop names` in `Base` for me to address

Intersect.Client.Core/Interface/Game/Draggable.cs
```diff
@@ -1,68 +1,43 @@
...
     {
-        get => mPnl.Y;
-        set => mPnl.Y = value;
+        return true;
     }
 
-    public bool Update()
+    public override Package DragAndDrop_GetPackage(int x, int y)
```
lodicolo, line 18: Should be `Package?` and please change `Base` to also have `Package?`

Intersect.Client.Core/Interface/Game/Hotbar/HotBarWindow.cs
```diff
@@ -50,16 +51,14 @@ public void Update()
         }
     }
 
-    public FloatRect RenderBounds()
+    public override void Hide()
     {
-        var rect = new FloatRect
+        // dont hide window if we are dragging something
+        if (Items.Any(c => c.IconImage == DragAndDrop.CurrentPackage?.DrawControl))
```
lodicolo, line 57: Same comment as in `BagWindow`

Intersect.Client.Core/Interface/Game/Inventory/InventoryWindow.cs
```diff
@@ -97,21 +97,13 @@ public override void Hide()
...
-        base.Hide();
-    }
 
-    // TODO: Window has RenderBounds as property, but InventoryWindow has it as a method. This should be consistent.
-    public FloatRect RenderBounds()
-    {
-        var tempSlot = Items[0];
-        var rect = new FloatRect()
+        // dont hide window if we are dragging something
+        if (Items.Any(c => c.IconImage == DragAndDrop.CurrentPackage?.DrawControl))
```
lodicolo, line 102: Same comment as in `BagWindow`

Also since this is repeated at least 4 times I think you should add a `// TODO: Create a base 'SlotWindow' that BagWindow/BankWindow/InventoryWindow/SpellsWindow should inherit from so this code can be written only once`

Intersect.Client.Core/Interface/Game/Spells/SpellItem.cs
```diff
@@ -145,43 +129,66 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+    }
+
+    #endregion
+
+    #region Drag and Drop
+
+    public override bool DragAndDrop_HandleDrop(Package package, int x, int y)
     {
-        switch (arguments.MouseButton)
+        var targetNode = Interface.FindComponentUnderCursor(NodeFilter.None);
```
lodicolo, line 164: Remove `NodeFilter.None`, the default is already that

Intersect.Client.Core/Interface/Game/Spells/SpellItem.cs
```diff
@@ -145,43 +129,66 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+        // Find the first parent acceptable in that tree that can accept the package
+        while (targetNode != default)
         {
-            case MouseButton.Left:
-                _clickTime = Timing.Global.MillisecondsUtc + 500;
-                break;
+            switch (targetNode)
+            {
+                case SpellItem spellItem:
+                    Globals.Me?.SwapSpells(SlotIndex, spellItem.SlotIndex);
```
lodicolo, line 172: Cache `Globals.Me` and return false _before_ `var targetNode = ...;` if it's null

Intersect.Client.Core/Interface/Game/Spells/SpellItem.cs
```diff
@@ -145,43 +129,66 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+                default:
+                    targetNode = targetNode.Parent;
+                    break;
+            }
+
+            // If we've reached the top of the tree, we can't drop here, so cancel drop
+            if (targetNode == null)
+            {
+                return false;
+            }
```
lodicolo, line 188: Unnecessary, already handled by the `while` and `return false;`

Intersect.Client.Core/Interface/Game/Spells/SpellsWindow.cs
```diff
@@ -74,21 +74,13 @@ private void InitItemContainer()
...
     {
         _contextMenu?.Close();
-        base.Hide();
-    }
 
-    public FloatRect RenderBounds()
-    {
-        var rect = new FloatRect()
+        // dont hide window if we are dragging something
+        if (Items.Any(c => c.IconImage == DragAndDrop.CurrentPackage?.DrawControl))
```
lodicolo, line 79: Same comment as I had in the other windows (I added a longer message in `InventoryWindow` though)

Intersect.Client.Framework/Gwen/Control/Base.cs
```diff
@@ -4330,7 +4330,7 @@ public virtual void DragAndDrop_Hover(Package p, int x, int y)
     // receiver
     public virtual bool DragAndDrop_CanAcceptPackage(Package p)
     {
-        return false;
+        return true;
```
lodicolo, line 4333: This should be false. It being true means a random label or button would be returning "true" for `CanAcceptPackage` which is totally wrong

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -94,29 +84,25 @@ private void _withdrawMenuItem_Clicked(Base sender, MouseButtonState arguments)
...
             return;
         }
 
-        if (_descWindow != null)
-        {
-            _descWindow.Dispose();
-            _descWindow = null;
-        }
+
+        _descWindow?.Dispose();
```
lodicolo, line 104: `_descriptionWindow`

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -138,41 +124,28 @@ private void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+            {
+                IconImage_DoubleClicked(sender, arguments);
+            }
         }
     }
 
-    private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
+    private void IconImage_DoubleClicked(Base sender, MouseButtonState arguments)
     {
         if (!Globals.InBank)
```
lodicolo, line 150: This if should be replaced with an if checking that it was left click

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -206,6 +179,81 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
 
     #endregion
 
+    #region Drag and Drop
+
+    public override bool DragAndDrop_HandleDrop(Package package, int x, int y)
+    {
+        var targetNode = Interface.FindComponentUnderCursor(NodeFilter.None);
```
lodicolo, line 186: Remove `NodeFilter.None`

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -206,6 +179,81 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+        {
+            switch (targetNode)
+            {
+                case BankItem bankItem:
+                    if (Globals.IsGuildBank)
+                    {
+                        if (Globals.Me is not { } player)
+                        {
+                            return false;
+                        }
```
lodicolo, line 199: This should be cached _before_ `var targetNode`

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -206,6 +179,81 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+
+                        var rank = player.GuildRank;
+                        var isInGuild = !string.IsNullOrWhiteSpace(player.Guild);
```
lodicolo, line 202: These should be cached before target node as well

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -206,6 +179,81 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+                            ChatboxMsg.AddMessage(
+                                new ChatboxMsg(
+                                    Strings.Guilds.NotAllowedSwap.ToString(player.Guild),
+                                    CustomColors.Alerts.Error,
+                                    ChatMessageType.Bank
+                                )
+                            );
+
+                            return false;
+                        }
```
lodicolo, line 214: This should also be before targetNode

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -206,6 +179,81 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+                default:
+                    targetNode = targetNode.Parent;
+                    break;
+            }
+            
+            // If we've reached the top of the tree, we can't drop here, so cancel drop
+            if (targetNode == null)
+            {
+                return false;
+            }
```
lodicolo, line 249: Unnecessary, remove

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -221,226 +269,41 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
-            {
-                _descWindow.Dispose();
-                _descWindow = default;
-                _iconImage_HoverEnter(default, default);
-            }
+            return;
         }
 
-        if (!IsDragging)
+        var itemTexture = Globals.ContentManager?.GetTexture(Framework.Content.TextureType.Item, descriptor.Icon);
```
lodicolo, line 290: `GameContentManager.Current`

Intersect.Client.Core/Interface/Game/Hotbar/HotbarItem.cs
```diff
@@ -229,29 +206,40 @@ private void _iconImage_HoverEnter(Base sender, EventArgs arguments)
...
+            else
+            {
+                targetNode = targetNode.Parent;
+            }
+
+            // If we've reached the top of the tree, we can't drop here, so cancel drop
+            if (targetNode == null)
+            {
+                return false;
+            }
```
lodicolo, line 239: Remove, unnecessary

Intersect.Client.Core/Interface/Game/Hotbar/HotbarItem.cs
```diff
@@ -229,29 +206,40 @@ private void _iconImage_HoverEnter(Base sender, EventArgs arguments)
...
                 _currentSpell.Id, _hotbarWindow.X + (_hotbarWindow.Width / 2), _hotbarWindow.Y + _hotbarWindow.Height + 2
             );
         }
     }
 
-    public FloatRect RenderBounds()
+    public override bool DragAndDrop_HandleDrop(Package package, int x, int y)
     {
-        var rect = new FloatRect()
+        var targetNode = Interface.FindComponentUnderCursor(NodeFilter.None);
```
lodicolo, line 220: Remove `NodeFilter.Nonde`

Intersect.Client.Core/Interface/Game/Hotbar/HotbarItem.cs
```diff
@@ -159,61 +150,47 @@ public void Activate()
...
-            _spellDescWindow.Dispose();
-            _spellDescWindow = null;
-        }
+    private void IconImage_HoverLeave(Base sender, EventArgs arguments)
+    {
+        _itemDescWindow?.Dispose();
+        _itemDescWindow = null;
+
+        _spellDescWindow?.Dispose();
+        _spellDescWindow = null;
```
lodicolo, line 175: `_itemDescriptionWindow`
`_spellDescriptionWindow`

Intersect.Client.Core/Interface/Game/Hotbar/HotbarItem.cs
```diff
@@ -229,29 +206,40 @@ private void _iconImage_HoverEnter(Base sender, EventArgs arguments)
...
+        while (targetNode != default)
         {
-            X = ToCanvas(new Point(0, 0)).X,
-            Y = ToCanvas(new Point(0, 0)).Y,
-            Width = Width,
-            Height = Height
-        };
+            if (targetNode is HotbarItem hotbarItem)
+            {
+                Globals.Me?.HotbarSwap(SlotIndex, hotbarItem.SlotIndex);
```
lodicolo, line 227: Cache `Globals.Me`, should return false before `targetNode` if it's null

Intersect.Client.Core/Interface/Game/Hotbar/HotbarItem.cs
```diff
@@ -405,12 +393,26 @@ public void Update()
...
+            _equipLabel.IsHidden = true;
+            _quantityLabel.IsHidden = true;
+            _cooldownLabel.IsHidden = true;
+        }
+        else
+        {
+            _equipLabel.IsHidden = !_isEquipped || _inventoryItemIndex < 0;
+            _quantityLabel.IsHidden = _currentItem?.Stackable == false || _inventoryItemIndex < 0;
+            _cooldownLabel.IsHidden = !_isFaded || _inventoryItemIndex < 0;
+        }
```
lodicolo, line 408: Cache `_inventoryIndex < 0`

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -269,72 +258,55 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
         }
     }
 
-    private void _iconImage_Clicked(Base sender, MouseButtonState arguments)
+    private void IconImage_Clicked(Base sender, MouseButtonState arguments)
     {
-        switch (arguments.MouseButton)
+        if (arguments.MouseButton is MouseButton.Right)
```
lodicolo, line 263: Since this is the only one in this method, flip it and return if it's not Right to reduce nesting

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -269,72 +258,55 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
         {
-            case MouseButton.Left:
-                _clickTime = Timing.Global.MillisecondsUtc + 500;
-                break;
-
-            case MouseButton.Right:
-                if (ClientConfiguration.Instance.EnableContextMenus)
+            if (ClientConfiguration.Instance.EnableContextMenus)
+            {
+                OpenContextMenu();
```
lodicolo, line 267: add a return after this line inside of the if statement so the `else` can be un-nested

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -269,72 +258,55 @@ private void _iconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+            }
+            else
+            {
+                if (Globals.GameShop != null)
                 {
-                    OpenContextMenu();
+                    Globals.Me?.TrySellItem(SlotIndex);
```
lodicolo, line 273: Cache `Globals.Me` and return if it's null

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -428,6 +400,80 @@ void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
 
     #endregion
 
+    #region Drag and Drop
+
+    public override bool DragAndDrop_HandleDrop(Package package, int x, int y)
+    {
+        if (Globals.Me?.Inventory is not { } inventory)
```
lodicolo, line 407: Cache `Globals.Me` before everything and return false if null

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -428,6 +400,80 @@ void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+            return false;
+        }
+
+        if (!Interface.DoesMouseHitInterface() && !Globals.Me.IsBusy)
+        {
+            PacketSender.SendDropItem(SlotIndex, inventorySlot.Quantity);
+            return true;
+        }
+
+        var targetNode = Interface.FindComponentUnderCursor(NodeFilter.None);
```
lodicolo, line 423: Remove `NodeFilter.None`

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -428,6 +400,80 @@ void _iconImage_HoverEnter(Base? sender, EventArgs? arguments)
...
+                default:
+                    targetNode = targetNode.Parent;
+                    break;
+            }
+
+            // If we've reached the top of the tree, we can't drop here, so return false
+            if (targetNode == null)
+            {
+                return false;
+            }
```
lodicolo, line 469: Remove, unnecessary

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -481,299 +514,68 @@ public override void Update()
...
-            {
-                _descWindow.Dispose();
-                _descWindow = null;
-                _iconImage_HoverEnter(null, null);
-            }
+            return;
         }
 
-        if (!IsDragging)
+        var itemTexture = Globals.ContentManager?.GetTexture(Framework.Content.TextureType.Item, descriptor.Icon);
```
lodicolo, line 544: `GameContentManager.Current`

Intersect.Client.Core/Interface/Game/Inventory/InventoryItem.cs
```diff
@@ -481,299 +514,68 @@ public override void Update()
...
+            if (IconImage.Texture != null)
             {
-                PacketSender.SendDropItem(SlotIndex, Globals.Me.Inventory[SlotIndex].Quantity);
+                IconImage.Texture = null;
+                IconImage.IsVisibleInParent = false;
             }
+        }
 
-            _dragIcon.Dispose();
+        if (_descWindow != null)
```
lodicolo, line 562: `_descriptionWindow`

lodicolo (changes requested):

Intersect.Client.Core/Core/Input.cs
WeylonSantana: I started doing the review, and line 142 in my opinion only blocks the open menu/close all windows, while line 205 only blocks the GUI's toogle, although the conditions are the same, the first one does not cancel the other, I tested with breakpoints, with the code as it is, maybe it is the bad arrangement of how the code is in the input, with controls being checked in separate ifs while others are checked in loops, but as it is, it is how it works, I need both if, inside the open menu and in the toggle gui

Intersect.Client.Core/Interface/Game/Inventory/InventoryWindow.cs
WeylonSantana: It is no longer repeated 4 times so I will not add the comment, the correct thing would be to block at the root, and I tried to do that now, the windows do not hide and others do not open while dragging, which is correct

Intersect.Client.Core/Interface/Game/Bank/BankItem.cs
```diff
@@ -183,37 +183,35 @@ private void IconImage_DoubleClicked(Base sender, MouseButtonState arguments)
...
+            ChatboxMsg.AddMessage(
+                new ChatboxMsg(
+                    Strings.Guilds.NotAllowedSwap.ToString(player.Guild),
+                    CustomColors.Alerts.Error,
+                    ChatMessageType.Bank
+                )
+            );
+
+            return false;
+        }
```
lodicolo, line 205: Isn't this breaking regular bank if you're not a guildmaster?

Intersect.Client.Framework/Input/ControlBinding.cs
```diff
@@ -34,7 +35,7 @@ public bool IsDown()
 
         if (IsMouseKey && Key.TryGetMouseButton(out var mouseButton))
         {
-            return gameInput.IsMouseButtonDown(mouseButton);
+            return gameInput.IsMouseButtonDown(mouseButton) && !DragAndDrop.IsDragging;
```
lodicolo, line 38: This also needs to be in `WasDown` to avoid unexpected mouse "release" behavior.

Intersect.Client.Core/Interface/Game/Draggable.cs
lodicolo: Where's the TODO? xD

lodicolo (changes requested):

lodicolo (approved):

PR #2678, opened 2025-03-18
fix: rewrite forgot password/password change windows, add change password button to select character window

### LINKED ISSUE #2674: bug: Recovery button missing
### Description

I've tryed to delete the "Dock" of the recovery Button on the UI because its "missing"
I got the "Recovery button" but only one time, after that you need to restart the client for see it again (only one time again)
https://cdn.discordapp.com/attachments/965617981159788614/1350918946311766139/ScreenRecorderProject8.mp4?ex=67d87ccf&is=67d72b4f&hm=8da44a5beb53bb768251ced022546e644ace730365382ab005517bdc06947415&

### Steps to Reproduce

1. Configure all your recovery password
2. Run the client and watch if the "recovery button" its there or not
3. if its there, logout and test the login button for watch it again
4. see the bug

### Version with bug

v0.8.0.525-beta+820d301c9027d8e2254896718498686a74ed62cf

Unknown

### Affected platforms

Windows 10

_No response_

### Relevant log output

```shell

```

### Duplicate Bug Check

pandinocoder:
Resolves #2674


TalikanPME:
@pandinocoder 


Following this update, the JSON on the client side was never updated.
The "change password" button is missing.
It appears correctly, but it is not possible to change the size or location.


Good day :)

PR #2689, opened 2025-03-22
fix: Add Close Button to Animation Editor & Conditionally Show Save Button Across Windows

### LINKED ISSUE #2688: bug: Cancel Button Missing in Animation Editor When No Item is Selected
### Description

The FrmAnimation window in the Intersect Editor is missing a visible Cancel button when no animation item is selected. 

### Steps to Reproduce

1. Open the Animation Editor window
2. Have no Animation selected.
3. Observe that there is no visible Cancel button

### Version with bug

v0.8.0.535-beta+a16320822a117256673a243e45da8b430502d8ba

Unknown

### Affected platforms

Windows 11

_No response_

### Relevant log output

```shell

```

### Duplicate Bug Check

TimtyG:
Enabled a visible Close button in Editor Windows.

The Save and Cancel button is now only shown when an item is selected

Showcase

Edit (Projectile Window fixed)

https://github.com/user-attachments/assets/55d2dd2f-1bdc-4470-bd14-6bd972a2b7d0

Resolves #2688


WeylonSantana:
ok, 3 main changes to happen here
a - some windows allow minimizing, but not maximizing
-- animation, common event, npcs, spell, variable,
b - some windows allow maximizing (without minimizing)
-- quest,
c - some windows do not allow minimizing or maximizing
-- crafts, crafting tables,
d - some windows allow both
-- class, item, projectile, resources, shop, time
all allow closing, ok.

change 1 - all windows have to allow only the close button, (look at crafts, crafting table)

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
```diff
@@ -0,0 +1,30 @@
+using System.Windows.Forms;
+
+namespace Intersect.Editor.Forms.Editors;
+
+public class BaseEditorForm : EditorForm
+{
+    protected Button btnSave;
+    protected Button btnCancel;
+
+    protected void SetEditorButtons(Button saveButton, Button cancelButton)
```
WeylonSantana, line 10: change number 2 - the class fields are protected, that is, the children have access to the fields, this method is not necessary if the children can set them directly

WeylonSantana (changes requested):

WeylonSantana:
change 3 - in all files - .Designer.cs, is where the code will assign these buttons for you, currently each designer file is overwriting them, remove of all .Designer.cs (is on end of every file)
image

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
```diff
@@ -0,0 +1,30 @@
+using System.Windows.Forms;
+
+namespace Intersect.Editor.Forms.Editors;
+
+public class BaseEditorForm : EditorForm
+{
+    protected Button btnSave;
+    protected Button btnCancel;
```
WeylonSantana, line 8: must be `DarkButton?` and not Button

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
```diff
@@ -0,0 +1,30 @@
+using System.Windows.Forms;
```
WeylonSantana, line 1: and nothing prevents these two properties and the UpdateEditorButtons method from being in the EditorForm file, so take them there to avoid more unnecessary inheritances for something so trivial

WeylonSantana (changes requested):

WeylonSantana:
finally, in each editor, look for references to btnSave and btnCancel, and check if there are any warnings about possible null references, if there are, check before `if (btn != default)`

TimtyG:
> in all files - .Designer.cs, is where the code will assign these buttons for you, currently each designer file is overwriting them, remove of all .Designer.cs (is on end of every file) image

This will remove buttons from visual designer preview.

TimtyG:
> in all files - .Designer.cs, is where the code will assign these buttons for you, currently each designer file is overwriting them, remove of all .Designer.cs (is on end of every file) image

This will remove the buttons from visual preview in designer.

TimtyG:
> ok, some main changes to happen here a - some windows allow minimizing, but not maximizing -- animation, common event, npcs, spell, variable, b - some windows allow maximizing (without minimizing) -- quest, c - some windows do not allow minimizing or maximizing -- crafts, crafting tables, d - some windows allow both -- class, item, projectile, resources, shop, time all allow closing, ok.
> 
> change 1 - all windows have to allow only the close button, (look at crafts, crafting table)

OK, done. All windows have only close button.

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
TimtyG: EditorForm is handling lifecycle and hook, not UI controls. So I made the class to sit between , so we can isolate save/cancel for editor windows that use them.

TimtyG: Every single form that inherits from it would get implicitly Save/Cancel logic.
Over time, EditorForm will become a junk drawer for unrelated logic and new forms that don’t need Save/Cancel buttons are still carrying dead weight.

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
TimtyG: True, but I added SetEditorButtons just to make setup explicit and safer across forms, it's a small layer to avoid silent breakage and help with future logic reuse.

Intersect.Editor/Forms/Editors/BaseEditorForm.cs
```diff
@@ -0,0 +1,31 @@
...
+public class BaseEditorForm : EditorForm
+{
+    protected DarkButton? btnSave;
+    protected DarkButton? btnCancel;
+
+    protected void SetEditorButtons(DarkButton saveButton, DarkButton cancelButton)
+    {
+        btnSave = saveButton;
+        btnCancel = cancelButton;
+    }
```
lodicolo, line 15: Pick one, either the fields are protected, or you have a `Set` method, but not both.

More details in `frmQuest`

Intersect.Editor/Forms/Editors/Quest/frmQuest.cs
```diff
@@ -31,8 +31,11 @@ public FrmQuest()
         Icon = Program.Icon;
 
         InitLocalization();
+        SetEditorButtons(btnSave, btnCancel);
```
lodicolo, line 34: They have the exact same name which means one of three things that each have exactly one remediation:
- They are defined in the child class which means they are shadowing the field in the base class, something we should not be doing.
  - The remediation here is to change the fields in the base class to be private.
- If they aren't private in the base class and are required to be protected because not all of the child classes have the same name, the SetEditorButtons() is are not a benefit here because the child classes can set all of the fields directly. 
  - The SetEditorButtons() method should be removed, and the base class field names should be changed to _buttonSave, _buttonCancel.
- They aren't defined in the child class at all, which means the sets wouldn't even be necessary since they would in effect be set to themselves, which means they already have the required value, which means there's no reason to re-set them.

Quite frankly, I am of the opinion that the method should not exist, or there should be one method per button so that they aren't all required for a form to exist.

Regardless of the implementation  however, they are not set in the constructor of the base class and so they must all be nullables.

lodicolo (changes requested):

lodicolo:
I didn't say anything was wrong, GitHub doesn't allow me to do something I tried to do, and the result is that it says I removed my request for review.

WeylonSantana (approved):

lodicolo (approved):

PR #2775, opened 2025-12-08
Feat: Three spell effects added

Kamusvalenzuela:
- Knockback (for some reason it was "invisible" and unused)
- Grievous Wounds
- Healing Boost

The healing and healing null percentages are editable in the server settings (I had trouble modifying the database, so I decided to leave it to someone more experienced).

I added comments to everything possible to make it as clear as possible.

Feedback is appreciated, as this is my first pull request for this project.


WeylonSantana:
Hey @Kamusvalenzuela
Thank you for contributing to the change. I haven't done a review yet, but I'll point out something I saw, even though it was only superficially. I believe it will give you enough work to change a lot of things.
1 - The MySQL migration is missing, and yes, you would need to have a MySQL server installed, an empty database, and easy connection between Intersect and MySQL. Generating the migration is just one command, just like you did for SQLite, but you need to have a functional database. You don't need data; just the existence of the database is enough.

2 - 
A. Its implementation method is not customizable; a global value for all spells is terrible when many users may want different values ​​for each spell (think of attempts to create a spell leveling system where the boost or reduction of healing would improve/worsen, allowing spell swapping with events).
B. It's not scalable, it doesn't allow for expansion later, it would only add more complication to adding other effects. Here's my suggestion:

In the spell combat descriptor, add "PercentageEffect" int, remove it from the server options.
This would be something more "global" to use and reuse, since only one effect is applied at a time. A PercentageEffect value can be used for this. Then, if I want to create an effect like haste, tenacity, etc., I would use PercentageEffect as well. Set PercentageEffect to appear only when it needs to appear; effects like stun, sleep, etc., wouldn't need it, right?

3 - The effect name seems too long "GrievousWoundsHealingReduction", and specific to some game, the effect name should just be HealingReduction and HealingBoost, in the enum and everywhere else.

4 - Okay then, add these 3 effects, but I need you to send a video showing everything working, the values ​​being what they should be, correct calculations and so on, otherwise I or another reviewer would have to clone and run it, create items/spells/NPCs, do all the setup and it would take a lot of time, I myself don't have time, I will analyze as much as possible by code and as a last resort before approving is when I would do all this to test, so, for that reason (lack of time) I am always looking at the files superficially, so a video is good, and a description showing how to do the setup is better so we just follow a script to test if it's ok.

I believe these 4 tasks will keep you busy for now.

pandinocoder:
Closing this for now, it can be reopened if you have time to finish it @Kamusvalenzuela

PR #2778, opened 2025-12-20
Fix: RichLabels improvements + Restores EventWindow

### LINKED ISSUE #2768: bug: no longer able to edit event window design options without modifying source
### Description

The ability to edit the skin, placement, design, etc of the event dialog window was removed in [v0.8.0-beta.382](https://github.com/AscensionGameDev/Intersect-Engine/releases/tag/v0.8.0-beta.382).

Loading UI properties from the event window jsons was disabled because it "broke stuff".

All of our player-facing UI windows should be able to be customized without programming. Whatever is broken we should fix and or find a reasonable workaround.

### Steps to Reproduce

CTRL+F EventDialogWindow json files
404 Not Found.

### Version with bug

v0.8.0-beta.382+

v0.8.0-beta.381

### Affected platforms

Windows 10

N/A

### Relevant log output

```shell
N/A
```

### Duplicate Bug Check

Arufonsu:
Fixes #2768

~~(Already merged)
[Previously required: merge assets into main_full branch](https://github.com/AscensionGameDev/Intersect-Assets/pull/71)~~

(Fix: wrap RichLabels when single word it's too long)
Image

Image

{924AE11D-F78A-4132-858B-69B30FF364C5}

{013C80BF-21D3-4E21-843B-04B0A1304914}

{2793577D-5E20-4B13-AFB7-81CD7E103A01}

### DIFF
--- a/Intersect.Client.Core/Interface/Game/EventWindow.cs
+++ b/Intersect.Client.Core/Interface/Game/EventWindow.cs
@@ -1,3 +1,4 @@
+using Intersect.Client.Core;
 using Intersect.Client.Core.Controls;
 using Intersect.Client.Entities.Events;
 using Intersect.Client.Framework.Content;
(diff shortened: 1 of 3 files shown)


Intersect.Client.Core/Interface/Game/EventWindow.cs
```diff
@@ -119,56 +83,62 @@ private EventWindow(Canvas gameCanvas, Dialog dialog) : base(gameCanvas, nameof(
...
 
-        _promptPanel.SizeToChildren(recursive: true);
+        CreateOptionButtons();
 
-        #region Configure and Display
+        // Load JSON to apply layout/templates
+        try
+        {
+            Name = nameof(EventWindow);
+            LoadJsonUi(GameContentManager.UI.InGame, Graphics.Renderer.GetResolutionString());
```
WeylonSantana: Event window lacks usable fallback layout when there is no JSON in the GUI.

Many of the windows that are already more up-to-date from the last times Panda and I were in this race are all styled before the `loadjsonui` call. This is so that if the user doesn't have the JSON (and with this pull, many won't), they don't need to download it from the official assets. Instead, it will be created from the default style styled in the code. Then, `loadjsonui` will load the user's styling. The assets shouldn't have been merged; the JSON sent to the assets should be a "pure" JSON created by the code styling and not customized manually.

Arufonsu: updated!
now the code will create it's own json file with a default styled layout as fallback if there's no json file

{72D7F9F9-A405-49EF-BAFC-1EB66AD62A51}

Arufonsu: ( also: deleted the merged EventWindow.json from the assets repo )

WeylonSantana (approved):

PR #2794, opened 2025-12-31
fix: stop drag-drop from reparenting item icons

### LINKED ISSUE #2793: bug: Spells and items being at same slot when dragged.
### Description

If you try to drag a item from inventory or bank, a spell from spellwindow or hot bat, theres a chance to put 2 of  they at same slot, and it can't be fixed until you relog.

### Steps to Reproduce

1 -  have 2 items at inventory, try drag it away from inventory sometimes, same for spells at spell window, hotbar and bank.
The bug occurs with no trigger.

I made a video showing how to reproduce
https://youtu.be/nQeBxordebI

### Version with bug

Latest version

Unknow

### Affected platforms

Windows 10, Windows 11

No work around, to fix you need to relog or the both items will be useless.

### Relevant log output

```shell

```

### Duplicate Bug Check

Arufonsu:
Fixes #2793

Prevent Gwen's default drop handler from reparenting the dragged icon into the hovered control, which could place the icon into the wrong slot/container. Draggable now forwards DragAndDrop_HandleDrop to its SlotItem parent, and the drag package stores the source SlotItem in Package.UserData to make slot handlers deterministic.

https://github.com/user-attachments/assets/304afac6-46d8-4d58-affa-1eabbed8ffbe

### DIFF
--- a/Intersect.Client.Core/Entities/Player.cs
+++ b/Intersect.Client.Core/Entities/Player.cs
@@ -415,7 +415,8 @@ public partial class Player : Entity, IPlayer
         }
 
         var quantity = inventorySlot.Quantity;
-        var canDropMultiple = quantity > 1;
+        var maxQuantity = GetQuantityOfItemInInventory(itemDescriptor.Id);
+        var canDropMultiple = maxQuantity > 1;
         var inputType = canDropMultiple ? InputType.NumericSliderInput : InputType.YesNo;
         var prompt = canDropMultiple ? Strings.Inventory.DropItemPrompt : Strings.Inventory.DropPrompt;
         _ = new InputBox(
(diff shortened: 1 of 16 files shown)


naziozeno:
Done a lot of testing for hours and the bug of stacking skills and items in the same slot has been fixed, thank you.
8 players tested for more than 4 hours and gave positive feedback for the fix

Intersect.Client.Core/Entities/Player.cs
```diff
@@ -415,7 +415,7 @@ public void TryDropItem(int inventorySlotIndex)
         }
 
         var quantity = inventorySlot.Quantity;
-        var canDropMultiple = quantity > 1;
+        var canDropMultiple = GetQuantityOfItemInInventory(itemDescriptor.Id) > 1;
```
WeylonSantana: Cache this value 1 line above, is being used on maxQuantity below, the cache will avoid 1 extra loop

WeylonSantana (changes requested):

PR #2796, opened 2026-01-07
fix: settings window crash after changing tile dimensions

### LINKED ISSUE #2795: bug: Tile change > Options not working
### Description

If I change the tilewidth and tileheight in Server>Resources>config.json, the options in the game no longer open and the client crashes.

    "MapHeight": 26,
    "MapItemHeight": 16,
    "MapItemWidth": 16,
    "MapWidth": 32,
    "TileHeight": 16,
    "TileWidth": 16,
    "TimeUntilMapCleanup": 30000,
    "ZDimensionVisible": false

### Steps to Reproduce

...

### Version with bug

0.8.0-beta.569

Unknown

### Affected platforms

Windows 11

_No response_

### Relevant log output

```shell

```

### Duplicate Bug Check

Arufonsu:
Fixes #2795
This fix ensures that when updating value ranges within Sliders and LabeledSliders:
- If the new minimum is greater than the current maximum, we set the maximum first to avoid the validation error
- If the new maximum is less than the current minimum, we set the minimum first
- Otherwise, we can safely set them in the original order

This prevents the intermediate state where minimum temporarily exceeds maximum, which was causing the crash when opening the settings window after changing tile dimensions.

Note: Tested by decreasing dimensions from 32x32 to 16x16, leading to no more crashes at settings windows
{860FFCF6-ECC1-4B10-8EBF-6DAADC636395}

### DIFF
--- a/Intersect.Client.Framework/Gwen/Control/Slider.cs
+++ b/Intersect.Client.Framework/Gwen/Control/Slider.cs
@@ -461,10 +461,15 @@ public partial class Slider : Base
         var actualMin = Math.Min(min, max);
         var actualMax = Math.Max(min, max);
 
-        if (actualMin > Maximum)
+        if (actualMin > _maximumValue)
         {
             _maximumValue = actualMax;
             _minimumValue = actualMin;
+        }
+        else if (actualMax < _minimumValue)
+        {
+            _minimumValue = actualMin;
+            _maximumValue = actualMax;
         }
         else
         {
--- a/Intersect.Client.Framework/Gwen/Control/LabeledSlider.cs
+++ b/Intersect.Client.Framework/Gwen/Control/LabeledSlider.cs
@@ -388,6 +388,11 @@ public partial class LabeledSlider : Base, ISmartAutoSizeToContents, INumericInp
             Maximum = actualMax;
             Minimum = actualMin;
         }
+        else if (actualMax < Minimum)
+        {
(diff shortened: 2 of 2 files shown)


Intersect.Client.Framework/Gwen/Control/Slider.cs
```diff
@@ -458,8 +458,22 @@ protected virtual void SetValueInternal(double newInternalValue, bool forceUpdat
     /// <param name="max">Maximum value.</param>
     public void SetRange(double min, double max)
     {
-        _minimumValue = min;
-        _maximumValue = max;
+        if (min > Maximum)
```
pandinocoder: Both comments from LabeledSlider apply here

Intersect.Client.Framework/Gwen/Control/Slider.cs
```diff
@@ -458,8 +458,19 @@ protected virtual void SetValueInternal(double newInternalValue, bool forceUpdat
...
+        if (actualMin > Maximum)
+        {
+            _maximumValue = actualMax;
+            _minimumValue = actualMin;
+        }
+        else
+        {
+            _minimumValue = actualMin;
+            _maximumValue = actualMax;
+        }
```
pandinocoder: This doesn't take into account if `actualMax < Minimum` before setting `_maximumValue = actualMax`

PR #2801, opened 2026-01-17
fix: character creation and selection flow

Arufonsu:
- Bug fix: Back button should always be visible on character creation window
- Bug fix: Play button was shown despite of having no characters created, this commit fixes that by properly updating character selection window buttons and display during initializing

Before:
image

After (with max chars 1, back logs account out):
{0C230371-336E-4EAD-9700-86C8959A4E2B}

After (with max chars >1, back sends to character select window, with proper display update):
{8BD4314D-66EE-4763-A6E9-32B92C2D334E}

{FE0BF14A-AA43-4D95-8C7E-E0A5F74C9F89}


Arufonsu:
(Commit got unsigned after rebase - unable to merge)

PR #2802, opened 2026-01-17
fix: duplicated exit window

Arufonsu:
- fixes duplication of exit confirmation window behavior across the game client

### DIFF
--- a/Intersect.Client.Core/Interface/Game/SimplifiedEscapeMenu.cs
+++ b/Intersect.Client.Core/Interface/Game/SimplifiedEscapeMenu.cs
@@ -6,8 +6,8 @@ using Intersect.Client.Framework.Gwen.Control.EventArguments;
 using Intersect.Client.General;
 using Intersect.Client.Interface.Shared;
 using Intersect.Client.Localization;
+using Intersect.Client.MonoGame;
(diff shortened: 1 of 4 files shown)


Intersect.Client.Core/Interface/Shared/AlertWindow.cs
```diff
@@ -49,6 +52,15 @@ private AlertWindow(
...
         TitleLabel.Margin = titleLabelMargin with { Left = titleLabelMargin.Left + 4 };
+
+        // Subscribe to the Closed event to handle X button clicks
+        Closed += OnWindowClosed;
+    }
+
+    private void OnWindowClosed(Base sender, EventArgs args)
+    {
+        // Trigger the cancel handler when window is closed via X button
+        _handleCancel?.Invoke(this, args);
```
pandinocoder: This should be in `InputBox`

pandinocoder: This would be invoking `Cancelled` from `InputBox`

pandinocoder (changes requested):

pandinocoder (approved):

PR #2803, opened 2026-02-01
performance: optimize entity packet broadcasting

Arufonsu:


Arufonsu:
Reminder: what about entity status updates? ,,,

